PR #5, opened 2023-02-09
Feature/accordion story

dgibson666:
* Integrated Storybook 6.5.x for HTML and Vite (builder)
* Added the HTML addon
* Added a barebones story for Accordion

Run from packages/styles:
* npm install
* npm run build
* npm run storybook

(Requires node 16; Storybook does not run on node 18 yet)


packages/styles/package.json
```diff
@@ -11,11 +11,29 @@
...
+    "@whitespace/storybook-addon-html": "^5.1.1",
+    "babel-loader": "^8.3.0",
+    "prettier": "^2.8.3",
+    "react-syntax-highlighter": "^15.5.0",
     "sass": "^1.58.0",
     "style-dictionary": "^3.7.2",
     "vite": "^4.0.0"
+  },
+  "dependencies": {
+    "nvm": "^0.0.4"
```
dannyk3941, line 37: We need to remove this package, `nvm` needs be installed on your machine and not as a project dependency. 

- [https://www.npmjs.com/package/nvm](https://www.npmjs.com/package/nvm)

packages/styles/.storybook/preview-head.html
```diff
@@ -2,5 +2,7 @@
     window.global = window;
 </script>
 
-<link rel="stylesheet" href="./dist/style.css" />
+<link rel="stylesheet" href="../dist/style.css" />
```
dannyk3941, line 5: Remove `../dist/` and `../assets/` from below, the `staticDirs: ['../dist', '../assets`] config serves these files up when running the dev server and build.

dannyk3941 (changes requested):
A few small changes and should be good to go after.

dannyk3941 (approved):
LGTM

PR #8, opened 2023-02-13
Fix/root font size

dannyk3941:
#### What's this PR do?
- Removes `:root` pseudo-class and replace with `html` selector
- Ignore `scss` where only generated date is modified
- Run `npm install`
- Run `npm run build-tokens`
#### Questions:
- Is there a blog post?
- Does the knowledge base need an update?


dgibson666 (approved):
looks good

PR #12, opened 2023-03-07
Feature/univ header story

dannyk3941:
#### What's this PR do?
1. Fix markup for Universal Header pattern
2. Change SCSS class names for Universal Header
3. Remove items from Storybook Toolbar (backgrounds, docs, measure, outline, grid, zoom)
4. Add storybook addon for dark-mode toggling with data-attributes
- `components/header.stories.js`
- `.storybook/preview.cjs`
- `.storybook/main.js`
- `sass/components/header/_universal.scss`
1. `npm install`
2. `npm run build-styles`
3. `npm run storybook -w packages/styles`
#### Questions:
- Is there a blog post?
- Does the knowledge base need an update?

### DIFF
--- a/packages/styles/index.html
+++ b/packages/styles/index.html
@@ -1,6 +1,5 @@
 <!DOCTYPE html>
 <html lang="en">
-
 <head>
   <meta charset="UTF-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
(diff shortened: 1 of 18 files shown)


packages/styles/index.html
```diff
@@ -35,24 +34,39 @@ <h6>Secure .gov websites use HTTPS</h6>
...
-      <div class="cbp-header__info dh-lg-none">
-        <a href="#" class="cbp-header__item" target="_blank" rel="noopener noreferrer"><i class="fas fa-book"></i></a>
-        <button class="cbp-header__item"><i class="fas fa-comment"></i></button>
-        <button class="cbp-header__item"><i class="fas fa-user"></i></button>
-      </div>
+  <header class="cbp-universal-header">
+    <div class="cbp-universal-header__brand">
+      <picture>
+        <source srcset="assets/images/CBP_SEAL.svg" media="(max-width: 599px)" type="image/svg+xml" height="44" width="44" />
+        <img src="assets/images/CBP_LOGO_HORIZONTAL_WHITE_TEXT.svg" alt="cbp brand logo" height="55" width="186" />
```
dgibson666, line 41: The alt text should be the full text of "U.S. Customs and Border Protection". Ideally, I'd like to revisit this implementation. We're making the user download separate images just to add the text, when the seal could/should be the same SVG used numerous places and cached on the first download. Do you know the font face of the text? Is it possible to reproduce that in real text instead of an image. Text as an image is never ideal.

packages/styles/src/sass/components/header/_universal.scss
```diff
@@ -1,61 +1,69 @@
...
+  font-size: $cbp-text-button;
   font-weight: 500;
   height: 36px;
   justify-content: center;
-  margin-right: 16px;
   text-decoration: none;
   text-transform: uppercase;
   width: 36px;
 
+  & span {
```
dgibson666, line 63: This is screaming for utility classes for responsive show/hide/visibly-hidden. We shouldn't have to write this into the component as a one-off and it should be visibly hidden at the small size so that the text can still label the link. Without it, it's not accessible. This is a TODO since we don't have those utility classes yet.

dgibson666 (commented):
The image alt text is the only blocker. The rest can be discussed and iterated at a later date.

dgibson666:
Seeing a console error before toggling the dark mode control and each time I click it:
contentScript.js:2 Uncaught (in promise) Error: Error when handling response of "IsInPageViewTrackingList": The message port closed before a response was received.
    at i (contentScript.js:2:1305573)

dgibson666 (approved):
With the alt-text fixed, we can address the other issues later and iterate since we don't have what we need to tackle them at the moment.

PR #24, opened 2023-04-13
Feature/interactive chips

dgibson666:
* Broke up combined (chip/badge/tag) story file into 3 stories and added controls
* Simplified markup (removed nesting where possible)
* Update/simplify CSS for simplified markup
* Added chip interactivity, refactored into a button for better accessibility and statefulness
* Created requirements for Chip since it's interactive


packages/styles/src/components/badge/badge.stories.js
```diff
@@ -0,0 +1,23 @@
+export default {
+  title: 'Patterns/Badges, Chips, and Tags',
+  argTypes: {
+    variant: {
+      control: { type: 'select' },
+      options: ['', 'danger'],
+      description: 'Adds a modifier class to the component to display the danger variation (see generated HTML)'
```
dannyk3941, line 7: Since this only has 2 options, switching this to a `boolean` control will make the JS more concise in the story. Also let's add a capitalized label to the controls panel by using the 'name' property in the argTypes object. (See Decision Card story argTypes object for "danger")

packages/styles/src/components/badge/badge.stories.js
```diff
@@ -0,0 +1,23 @@
...
+
+const BadgeTemplate = (args) => {
+  const badgeType = args.variant ? `cbp-badge--${args.variant}` : null;
+  
+  return `
+    <div class="${badgeType ? 'cbp-badge ' + badgeType : 'cbp-badge'}">
+      5
+    </div>
+  `;
+};
```
dannyk3941, line 20: - Line 12, lets destructure the `args` parameter to show only `variant` (or `danger`)
- Because of the change from the previous comment, we can delete the variable and put a single ternary operator in the class attribute. `cbp-badge` will always be there so only the variant that needs to show/hide is the modifier class (`variant ? 'cbp-badge--danger' : ''`)
- Since we delete the variable, we can replace the brackets with parentheses and remove the return statement since its only returning the markup in the template literal
```
const BadgeTemplate = ({ variant }) => ( `[markup here]`);
```

packages/styles/src/components/badge/badge.stories.js
```diff
@@ -0,0 +1,23 @@
...
+
+export const Badge = BadgeTemplate.bind({});
+Badge.args = {};
```
dannyk3941, line 23: Add default `false` arg for `variant` (`danger`)

packages/styles/src/components/chip/chip.stories.js
```diff
@@ -0,0 +1,25 @@
...
+
+  return `
+    <button type="button" class="cbp-chip" ${args.active != undefined ? 'aria-pressed="'+args.active+'"' : ''}>
+      <span class="cbp-chip__label">Hotlist</span>
+      <div class="cbp-chip__icon">
+        <i class="fas fa-plus"></i>
+      </div>
+    </button>
+  `;
+};
```
dannyk3941, line 22: Similar to the badge story
- Destructure the `args` parameter
- Let's discuss this, does the `aria-pressed` attribute need to be hidden when the markup is first loaded? then after its clicked, it will set the attribute and set it to `true/false` on subsequent clicks. Can we just set the attribute when its first loaded? 
- Same as previous comment, if its only return the mark up and no variables inside the entire function we can replace the brackets with parentheses and delete the `return` statement.

packages/styles/src/components/tag/tag.stories.js
```diff
@@ -0,0 +1,24 @@
...
+
+const TagTemplate = (args) => {
+  const tagType = args.variant ? `cbp-tag--${args.variant}` : null;
+  
+  return `
+      <div class="${tagType ? 'cbp-tag ' + tagType : 'cbp-tag'}">
+        Hotlist
+      </div>
+  `;
+};
```
dannyk3941, line 21: - Destructure the `args` parameter
- Similar to the badges, we can leave the `cbp-tag` name in the `class` attribute and only hide/show the modifier class. We can do something like this `<div class="cbp-tag ${variant ? 'cbp-tag--' + variant : ''}">` since empty strings in JS are falsey and one of the options in the list.

packages/styles/src/sass/components/_badges-chips-tags.scss
```diff
@@ -56,58 +46,63 @@
...
-    outline: none;
+    outline: 0;
 
-    > span + .plus-border {
+    > .cbp-chip__label + .cbp-chip__icon {
       background-color: transparent;
-      border-color: var(--cbp-color-white);
+      border-color: var(--cbp-color-white); 
+      outline: 2px solid white;
+      outline-offset: -4px;
```
dannyk3941, line 104: Use CSS variables

packages/styles/src/sass/components/_badges-chips-tags.scss
```diff
@@ -56,58 +46,63 @@
...
+  
+  & > .cbp-chip__label + .cbp-chip__icon {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    min-width: 28px;
+    min-height: 28px;
+    border-width: 0;
+    border-radius: 14px;
+    margin-left: 4px;
```
dannyk3941, line 71: Use CSS variables

dannyk3941 (changes requested):
Looks good, use of CustomEvent on the Chip is great! see code comments.

packages/styles/src/components/badge/badge.stories.js
dgibson666: this shouldn't matter either way. undefined or false will not output the extra class. The default state of the control can be unset for consistency.

packages/styles/src/components/badge/badge.stories.js
dgibson666: Done

packages/styles/src/components/badge/badge.stories.js
dgibson666: done.

I don't give a lot of thought or priority to optimizing story code. My thinking was to keep the stories more uniform by passing `args` in. It's less re-work if we add more controls later.

packages/styles/src/components/chip/chip.stories.js
dgibson666: Actually there should be aria-pressed by default, so I went ahead and updated. I think I copied that from the segmented button group and that component sets aria-pressed automatically for the whole group.

Fixed the rest.

dannyk3941 (approved):

PR #26, opened 2023-04-20
Fix/consolidate buttons links

dannyk3941:
#### What's this PR do?
- Fix Link pattern CSS
- Re-organize CSS for button patterns
- bump `vite-builder` npm package to use MDX files/docs in Storybook

- `styles/src/sass/button/*.scss`
- `styles/src/components/button/`

- **Run `npm install` to bump `vite-builder` npm package**
- Compare buttons to mock ups


packages/styles/src/sass/components/button/_base.scss
```diff
@@ -0,0 +1,108 @@
+// Base Button CSS
+.cbp-btn {
+  border: 0;
```
dgibson666, line 3: I think all buttons should have a border-width set so that they are sized the same and there's no jumpiness on hover/focus. Only the color changes.

packages/styles/src/sass/components/link/_index.scss
```diff
@@ -0,0 +1,80 @@
+.cbp-link {
+  font-weight: var(--cbp-font-weight-regular);
```
dgibson666, line 2: This should not have to be set explicitly - it should inherit from its context.

dgibson666 (changes requested):
Lots of good updates. In addition to the code-level comments, I would recommend:

* Removing separate stories for primary, secondary, and danger buttons and rolling those up into a `color` control.
* Add a "type" (per comment below) control that controls the button's type attribute and 
* Rename the "type" control to "fill".
* There should still be a "solid" option - let the code handle the output.

dgibson666:
Add-on notes about disabled links:

The disabled link reads fine in NVDA. It reads as "unavailable link, disabled text link", the second half being the link text itself. As mentioned, without an href, this is not an interactive element. It is not tab-able, nor does the "next link" keyboard shortcut find it ("no more links" or something).

Adding an href makes it interactive - tab-able and the shortcut key finds it and it reads exactly the same, indicating that it's unavailable. Perhaps we can wire this up with JS to event.preventDefault() if it's disabled so that it works more like a link but doesn't navigate.

Neither case (no href or preventDefault) would stop a routing link from firing in Angular.

packages/styles/src/components/link/link.md
```diff
@@ -0,0 +1,22 @@
+# Hyperlink (Anchor)
+
+### Purpose
+
+`<a href="[url]" class="cbp-link">` creates a hyperlink to web pages within or outside of the current URL. Text links can be embedded within a block of text and should not be used in lieu of a button. 
+
+### Functional Requirements
+
+- **Inline Text Link:** used when a link needs to be embedded in a chunk of text. These links These links should lead to toher pages or applications.
```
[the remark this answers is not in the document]
dannyk3941: fixed

packages/styles/src/components/link/link.md
```diff
@@ -0,0 +1,22 @@
...
+- **Definition Text Link:** used when the developer wants to give more informatoin about a term or acronym and can be used in conjuction with an `<abbr>` tag. These links should lead to other pages or applications.
```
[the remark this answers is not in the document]
dannyk3941: good idea. fixed.

packages/styles/src/sass/components/link/_index.scss
dannyk3941: fixed.

packages/styles/src/components/button/button.stories.js
```diff
@@ -1,65 +1,164 @@
...
+  }
+
+  return btnColor;
+}
+
+const Template = ({ square, color, type, label, disabled }) => (
+  `
+    <button
+      class="${square ? 'cbp-btn-square' : 'cbp-btn'} ${setBtnType(color, type)}"
+      type="button"
```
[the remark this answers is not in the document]
dannyk3941: added the controls for `type` attribute.

packages/styles/src/sass/components/button/_base.scss
dannyk3941: different padding is set for outline buttons vs solid and ghost to match the height in the mockups.

dgibson666:
I have a few notes, but am fine with merging and iterating from here:

Shouldn't we use the disabled cursor for all disabled buttons and links? Maybe something in our base.css, not component-specific:

``:disabled,
[disabled],
[aria-disabled],
a[aria-disabled=true]:not[href] {
  cursor: not-allowed;
}``

Does the floating action button's dropshadow match an elevation design token? I'm still unsure how we would make positioning of this easily customizable without writing app-specific CSS. We can revisit this as an implementation detail later.

PR #28, opened 2023-05-03
Fix/input groups

dannyk3941:
#### What's this PR do?
- Remove and consolidate text inputs css and stories, inputs include: 

1. Simple Text Field
2. Textarea
3. Password Field
4. Simple Numeric
5. Numeric Counter
6. Numeric Switch (excludes `numberSwitch.js`, I would like to tackle this another time)
7. Button Text Field + Square Button
8. Search field (not including predictive search functionality)

Does not include: 
1. Phone Number field (`<select>` and Dropdown pattern needs fix)
2. Content Search with predictive search

- `sass/text-input`
- `components/text-input`
- `components/textarea`


packages/styles/src/sass/components/text-input/_index.scss
```diff
@@ -0,0 +1,233 @@
...
+    -webkit-appearance: none;
+    margin: 0; /* Margins still appear even if hidden */
+  }
+}
+
+textarea.cbp-input {
+  min-height: 6.75rem;
+  padding-top: var(--cbp-space-2x);
+  padding-bottom: var(--cbp-space-2x);
+  resize: none;
```
dgibson666, line 103: This should be "auto" - we should never force turning off a user feature like this. If the width is 100% (not sized otherwise), maybe use "vertical".

packages/styles/src/sass/components/text-input/_index.scss
```diff
@@ -0,0 +1,233 @@
...
+  display: flex;
+
+  & > .cbp-input {
+    border-right: 0;
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+    flex: 1 1 auto;
+
+    &:disabled {
+      cursor: pointer;
```
dgibson666, line 117: can we delete and handle at a higher level? Doesn't need the added specificity.

packages/styles/src/sass/components/text-input/_index.scss
```diff
@@ -0,0 +1,233 @@
...
+  & > div:first-child > .cbp-input {
+    -moz-appearance: textfield; // Removes the default increment arrows in Firefox
+    border: none;
+    min-height: 100%;
+    padding-top: 0;
+    padding-bottom: 0;
+  }
+
+  & > div:first-child::after {
+    content: 'lbs';
```
dgibson666, line 193: This doesn't lend itself to updating very well. Also noted in my main comments.

packages/styles/src/sass/components/text-input/_index.scss
```diff
@@ -0,0 +1,233 @@
...
+  border-style: solid;
+  border-width: var(--cbp-border-size-md);
+  border-radius: var(--cbp-border-radius-soft);
+  color: var(--cbp-color-text-darkest);
+  font-size: var(--cbp-font-size-body);
+  font-weight: var(--cbp-font-weight-regular);
+  letter-spacing: var(--cbp-letter-spacing-regular);
+  min-height: 2.25rem;
+  padding: var(--cbp-space-2x);
+  width: 100%;
```
dgibson666, line 34: Can we set this only if :not([size]) and for textareas :not([cols]) ?

dgibson666 (commented):
Overall looks good. I'm very happy you renamed a bunch of classes to use input instead of form. I felt like form was not specific enough for these cases.

Here are some comments outside of the code-level comments:

We should pass a default arg for input `name`. The default story without setting it shows `named="undefined"` in the HTML, which is not ideal. The name should never be undefined.

We should really not be showing placeholder text in every single story. Generally speaking, placeholder text is discouraged. USWDS even explicitly advises against using it. In most of our stories, it is duplicative of the label, adding no value and causing verbosity for screen readers (not to mention the other usability issues).

Inputs with buttons: 
They should use `aria-controls`, referencing the input by ID. This doesn't cause any additional overhead because the inputs have to have an ID for the label to reference them.

For Numeric switch field, I think you need `data-segmented-button-type="single"` to force the single-select functionality on the group. One should be selected by default. The overlay of "lbs/kg" should update when toggled.

For password:
The button is basically acting as a toggle buttons, so let's give it some statefulness using `aria-pressed`. We just need to establish which is the default state, which corresponds to `aria-pressed="false"`.


Several issues that are larger discussion including required fields, focus states, and error states (and accessibility) can be discussed and addressed later. Some of these also seem like they need additional functionality (like Textfield with button), which can be added in future iterations.

packages/styles/src/sass/components/text-input/_index.scss
dannyk3941: done

packages/styles/src/sass/components/text-input/_index.scss
dannyk3941: we could add a span element and change with JS, let's tackle this once we nail down the functionality of the numeric switch pattern.

packages/styles/src/sass/components/text-input/_index.scss
dannyk3941: removed

packages/styles/src/sass/components/text-input/_index.scss
dannyk3941: removed

packages/styles/src/sass/components/text-input/_index.scss
```diff
@@ -0,0 +1,233 @@
...
+    &:hover,
+    &:focus {
+      border-color: var(--cbp-color-interactive-neutral-light);
+    }
+  }
+
+  &:disabled {
+    background-color: var(--cbp-color-interactive-disabled-light);
+    border-color: var(--cbp-color-interactive-disabled-dark);
+    cursor: pointer;
```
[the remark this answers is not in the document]
dannyk3941: removed

dannyk3941:
> We should really not be showing placeholder text in every single story. Generally speaking, placeholder text is discouraged. USWDS even explicitly advises against using it. In most of our stories, it is duplicative of the label, adding no value and causing verbosity for screen readers (not to mention the other usability issues).
> 
> Several issues that are larger discussion including required fields, focus states, and error states (and accessibility) can be discussed and addressed later. Some of these also seem like they need additional functionality (like Textfield with button), which can be added in future iterations.

I'll take note of these to discuss as a team, maybe we can use the `placeholder` to use to show examples of what the input should be and the error states/colors need to be discussed.

packages/styles/src/components/text-input/obfuscatedField.js
```diff
@@ -10,9 +14,11 @@ class ObfuscatedField {
     btn.addEventListener("click", (e) => {
```
dgibson666, line 14: The `aria-label` needs to be toggled as well from "show password" to "hide password". Can this be used for non-password fields like SSN? Maybe we should use more generic terminology like "show entry" and "hide entry"?

packages/styles/src/components/text-input/obfuscatedField.js
```diff
@@ -2,6 +2,10 @@ class ObfuscatedField {
   constructor(component) {
     this.input = component;
     this.btn = component.nextElementSibling;
+     
+    if (this.btn) {
+      this.btn.setAttribute('aria-pressed', false)
```
dgibson666, line 7: aria-attributes should always be a string. So 'false' should be quoted.

packages/styles/src/components/text-input/obfuscatedField.js
```diff
@@ -10,9 +14,11 @@ class ObfuscatedField {
     btn.addEventListener("click", (e) => {
       if (input.type === "password") {
         input.type = "text";
+        btn.setAttribute('aria-pressed', true)
```
dgibson666, line 17: needs a string value

packages/styles/src/components/text-input/obfuscatedField.js
```diff
@@ -10,9 +14,11 @@ class ObfuscatedField {
...
         btn.firstElementChild.className = "fas fa-eye";
       } else {
         input.type = "password";
+        btn.setAttribute('aria-pressed', false)
```
dgibson666, line 21: needs a string value

packages/styles/src/components/text-input/obfuscatedField.js
dannyk3941: done.

packages/styles/src/components/text-input/obfuscatedField.js
dannyk3941: `setAttribute` will convert non-string values to strings.

dgibson666: Yep, I totally forgot that. Thinking about the issues I've run into in JSX.

dgibson666:
This looks good. I think the numeric switch field needs functionality hooked up, but that can be done in a later iteration.

dgibson666 (approved):

PR #31, opened 2023-05-15
Fix/form control updates

dannyk3941:
#### What's this PR do?
- Clean up CSS for checkboxes and radio buttons
- Add story controls for checkboxes and radio buttons
- Bootstrap indeterminate checkbox behavior


packages/vanilla/src/components/checkbox/checkbox.md
```diff
@@ -0,0 +1,25 @@
+# Checkbox
+
+## Purpose
+
+Common form control component that allows the user to choose one or more options.
```
dgibson666, line 5: I feel like this is a source of confusion from the start - this pattern is called checkbox, but it's describing a checklist. Maybe we should have a checklist pattern, because having a single standalone checkbox is also a legitimate use case.

packages/vanilla/src/components/checkbox/checkbox.stories.js
```diff
@@ -0,0 +1,165 @@
...
+)
+
+const IndeterminateCheckboxTemplate = ({ indeterminateCheckboxObj, nestedCheckboxObj: { name: nestCheckboxName, checkbox1, checkbox2, checkbox3 }, disabled }) => {
+  return `
+    <div class="cbp-checkbox-item cbp-margin-bottom-5x">
+      <input type="checkbox" name=${indeterminateCheckboxObj.name} class="cbp-input__checkbox" id=${indeterminateCheckboxObj.id} value=${indeterminateCheckboxObj.value} data-checkbox="indeterminate" ${disabled ? 'disabled' : ''}>
+      <label for=${indeterminateCheckboxObj.id}>${indeterminateCheckboxObj.label}</label>
+    </div>
+    <div class="cbp-checkbox--nested">
+      <div class="cbp-checkbox-item cbp-margin-bottom-5x">
```
dgibson666, line 61: I would probably use flexbox and gap for the list spacing. It's more obvious for radios where you allow switching between vertical and horizontal.

packages/vanilla/src/components/radio/radioButton.stories.js
```diff
@@ -0,0 +1,95 @@
...
+    },
+    checked: {
+      name: 'Checked',
+      description: 'Add the `checked` attribute to the radio button (first choice only)',
+      control: { type: 'boolean' }
+    }
+  },
+  decorators: [
+    (Story, context) => `
+      <fieldset class="cbp-fieldset ${context.args.displayInline ? 'cbp-fieldset--inline' : ''}">
```
dgibson666, line 45: Same as for checklist, but instead of `role="group"` I believe you can use `role="radiogroup"` here (in place of fieldset/legend).

packages/vanilla/src/components/radio/radioButton.stories.js
```diff
@@ -0,0 +1,95 @@
...
+        <p class="cbp-input__description cbp-input__description--error" hidden><i class="fas fa-exclamation-triangle"></i>${context.args.errorMessage}</p>
+        ${Story().outerHTML || Story()}
+      </fieldset>
+    `
+  ]
+};
+
+const Template = ({ formControlName, radioBtnObj: { radio1, radio2, radio3 }, disabled, checked }) => (
+  `
+    <div class="cbp-radio-item cbp-margin-bottom-5x">
```
dgibson666, line 57: Using gap instead of the margins allows this to transition seamlessly between flow row and column.

packages/vanilla/src/sass/base/_core.scss
```diff
@@ -23,7 +20,7 @@ button {
 
 [aria-disabled=true],
 :disabled,
-input:read-only,
+input:read-only:not([type="checkbox"], [type="radio"], [type="file"], [type="range"], [type="color"]),
```
dgibson666, line 23: is this necessary? Why would you not use the not-allowed cursor on these input types if they are disabled? Or are they showing up because of the "readonly" aspect?

dgibson666 (commented):
It looks like we're replacing the native checkboxes and radios entirely and it takes a lot of CSS to make them appear right? Have we tried just styling the native controls? I've had much more success with this in modern browsers since dropping IE.

packages/vanilla/src/components/checkbox/checkbox.md
dannyk3941: We can add a story for a single checkbox and change the default to "Checklist" to make it more clear.

packages/vanilla/src/sass/base/_core.scss
dannyk3941: These are necessary because the not-allowed cursor is showing up on these form controls when they are non-disabled, these form controls do not have a `readonly` attribute and no `:readonly` pseudo-class in their css.

packages/vanilla/src/components/checkbox/indeterminate.js
```diff
@@ -0,0 +1,46 @@
+class IndeterminateCheckbox {
+  constructor(checkbox) {
+    this.indeterminateCheckbox = checkbox;
+    this.nestedCheckboxWrapper = this.indeterminateCheckbox.closest('.cbp-checkbox-item').nextElementSibling;
+    this.nestedCheckboxInputs = this.nestedCheckboxWrapper.querySelectorAll('input[type="checkbox"');
```
dgibson666, line 5: I think you're missing the closing `]` in your attribute selector.

packages/vanilla/src/components/checkbox/checkbox.stories.js
```diff
@@ -0,0 +1,165 @@
...
+    },
+    disabled: {
+      name: 'Disabled',
+      description: 'Disable checkbox. **In order to see a "checked" disabled checkbox, toggle the Checked story control to `true`**',
+      control: { type: 'boolean' }
+    }
+  },
+  decorators: [
+    (Story, context) => `
+      <fieldset class="cbp-fieldset ${context.args.displayInline ? 'cbp-fieldset--inline' : ''}">
```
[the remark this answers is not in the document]
dannyk3941: We can leave the `<fieldset>` and `<legend>` elements per our discussion and test for accessibility in the near future.

packages/vanilla/src/components/checkbox/indeterminate.js
dannyk3941: thanks for catching that!

dgibson666 (approved):
Looks good

PR #32, opened 2023-05-18
Feature/tabs

dgibson666:
* First cut of tabs (no responsive, no roving tabindex)
* Supports Danger variation.
* Minor tweaks to base and reset CSS around bottom margins for paragraphs and lists.

The design specs have not been updated yet, so some of this look off visually. E.g., the active tab color is not distinguishable enough from the other tabs in black. But my token values should match what are in the specs. These can be updated when the design specs are revised.


packages/vanilla/src/sass/components/tabs/_index.scss
```diff
@@ -0,0 +1,58 @@
...
+  //gap: var(--cbp-space-2x);
+  border-bottom: var(--cbp-border-size-md) solid var(--cbp-color-gray-cool-30);
+  margin-bottom: var(--cbp-space-4x);
+
+  .cbp-tab {
+    all: unset;
+    min-height: var(--cbp-space-14x);
+    padding: 0 var(--cbp-space-3x);
+    margin-bottom: calc(var(--cbp-space-half-x) * -1);
+    border-bottom: var(--cbp-border-size-lg) solid transparent;
```
dannyk3941, line 13: Mockups say the border-bottom of a hovered/active tab should be 4px width

packages/vanilla/src/sass/components/tabs/_index.scss
```diff
@@ -0,0 +1,58 @@
+.cbp-tabset {
+  display: flex;
+  align-items: flex-end;
+  //gap: var(--cbp-space-2x);
+  border-bottom: var(--cbp-border-size-md) solid var(--cbp-color-gray-cool-30);
+  margin-bottom: var(--cbp-space-4x);
+
+  .cbp-tab {
+    all: unset;
+    min-height: var(--cbp-space-14x);
```
dannyk3941, line 10: The height I'm getting when inspecting the tab is 59px, should be 56px. I think this has to do with the `all: unset` property because I'm seeing that the tabs have the `box-sizing: content-box` property being set in the Computed tab of the inspector.

packages/vanilla/src/sass/components/tabs/_index.scss
```diff
@@ -0,0 +1,58 @@
+.cbp-tabset {
+  display: flex;
+  align-items: flex-end;
+  //gap: var(--cbp-space-2x);
+  border-bottom: var(--cbp-border-size-md) solid var(--cbp-color-gray-cool-30);
+  margin-bottom: var(--cbp-space-4x);
```
dannyk3941, line 6: I don't think we need this, we can let the content of whatever it's going to have pad/margin itself.

dannyk3941 (changes requested):

packages/vanilla/src/sass/base/_core.scss
```diff
@@ -24,4 +31,8 @@ input:read-only:not([type="checkbox"], [type="radio"], [type="file"], [type="ran
 textarea:read-only,
 select:read-only {
   cursor: not-allowed;
+}
+
+[hidden] {
+  display: none !important;
```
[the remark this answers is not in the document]
dgibson666: Looks like I didn't add my comment on it. This is to address cases where you may be styling something with flex/grid/etc. in a class with more specificity, which then overrides the `hidden` attribute's behavior. This is typically not desired - adding or removing the hidden attribute should just work. I've seen this override in other systems like Bootstrap as well, but not all of them.

packages/vanilla/src/sass/base/_core.scss
```diff
@@ -13,6 +13,13 @@ body {
   background-color: var(--cbp-color-branding-cbp-dark);
 }
 
+p,
+blockquote,
+ul:not(ul ul):not(ul ol),
+ol:not(ol ol):not(ol ul) {
+  margin-bottom: var(--cbp-space-4x);
```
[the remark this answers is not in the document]
dgibson666: We need to make a decision on this because we're going back and forth. Is a class really required on a `p` tag to prevent it from running together into a wall of text? This is a super basic style on default HTML tags that I believe is needed as the base level. The rules on the list are just to make sure they have space below them as well, except when nested,

packages/vanilla/src/components/tabs/tabs.js
```diff
@@ -0,0 +1,66 @@
...
+        this.handleTabClick(e);
+      });
+    });
+
+    // Set the initial active state
+    this.initTabset();
+  }
+
+  initTabset() {
+    let err;
```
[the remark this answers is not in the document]
dgibson666: Will remove. I had plans and reworked my code.

packages/vanilla/src/sass/components/tabs/_index.scss
dgibson666: Ah, good catch. I missed that. Also a reason I rarely use `all: unset`. I'll revisit and see what's more efficient.

packages/vanilla/src/sass/components/tabs/_index.scss
dgibson666: Almost nothing has a default top margin/padding, so it won't - the content will butt right up to the bottom of the tabs, which looks bad. I find this preferable to putting the styling on the tab panels. There were no specs on this, so I'd rather revisit this when the specs get updated. This wouldn't apply for vertical tabs either, so we may need to tweak things when we allow for both as well. But that change would affect the tabset, not the panels, which goes with my reasoning for putting it there.

packages/vanilla/src/sass/components/tabs/_index.scss
dgibson666: Fixed. I thought I had the right token, but must have miscalculated. Nice catch.

dannyk3941 (approved):

PR #33, opened 2023-05-30
fix: update accordion content and css

dannyk3941:
#### What's this PR do?

- Added CSS for Accordion content border and change class names
- Added `aria-expanded` to accordion js 

**"base" colors are still used and needs to be updated once colors are finalized**


packages/vanilla/src/components/accordion/accordion.js
```diff
@@ -1,13 +1,25 @@
 class Accordion {
-  constructor(domNode) {
-    this.accordionNode = domNode;
+  constructor(accordionItem) {
+    this.accordionNode = accordionItem;
+    this.accordionTrigger = this.accordionNode.querySelector('.cbp-accordion__trigger');
```
dgibson666, line 4: General comments on naming for consistency:

* Why not just "accordion" for the self-referencing DOM node? I think we've done that in others.
* Can we use the word "control" (instead of "trigger") for consistency, e.g., `accordionControl`?

packages/vanilla/src/sass/components/accordion/_index.scss
```diff
@@ -1,148 +1,127 @@
-//@use '../../base/colors' as *;
-@use '../../typography' as *;
-@use './states' as *;
-//@use '../../tokens/font' as *;
-
-.cbp-accordion {
-  margin: 24px 0;
+.cbp-accordion-group {
+  margin-top: var(--cbp-space-6x);
```
dgibson666, line 2: Why not still use the shorthand than 2 separate properties? It's shorter.

packages/vanilla/src/sass/components/accordion/_index.scss
```diff
@@ -1,148 +1,127 @@
...
-    padding: 16px;
-    transition: all 300ms;
-
-    div:first-child i {
-      color: var(--cbp-color-text-darker);
-      font-size: var(--cbp-font-size-heading-md);
-      transition: transform .2s ease-in-out;
-    }
+.cbp-accordion__item:not(:last-child) {
+  margin-bottom: var(--cbp-space-3x);
```
dgibson666, line 8: Did you test if this is needed? Margins should collapse, so this should basically disappear for the last item since the wrapper has a larger bottom margin.

packages/vanilla/src/components/accordion/accordion.js
dannyk3941: Changed names in accordion js.

packages/vanilla/src/components/accordion/accordion.stories.js
```diff
@@ -1,57 +1,77 @@
...
+const Template = ({ defaultAccordion, badgeAccordion, iconAccordion }) => {
   return `
-    <div class="cbp-accordion">
-      <div class="cbp-accordion__item">
-        <button class="cbp-accordion__title" aria-expanded="false" aria-controls="accordion-demo-1">
-            <div>
-                <i class="fas fa-angle-down"></i>
-                <span>Accordion Control</span>
-            </div>
+    <div class="cbp-accordion-group">
```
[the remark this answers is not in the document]
dannyk3941: done

packages/vanilla/src/sass/components/accordion/_index.scss
dannyk3941: By default it doesn't add left/right margins so it just needs to be set on the top/bottom. There is a top margin in the mockups so we can leave it in there and see if a future mock up with other patterns makes it look bad.

packages/vanilla/src/sass/components/accordion/_index.scss
dannyk3941: Yes, this should add bottom margins to every accordion item except the last accordion item otherwise there would need to be another line with `cbp-accordion__item:last-child { margin-bottom: 0}` so its should make the css shorter.

dgibson666: This isn't true. I'm saying you don't need to specify no margin for the last item because the wrapper has a margin and margins collapse - only the larger (wrapper's) margin will be applied. If you don't omit the last item, there's no change. I verified in-browser.

dgibson666 (approved):
looks good

PR #34, opened 2023-06-01
Feat/select element

dannyk3941:
#### What's this PR do?
- Clean up `<select>` element CSS
- Create separate story for `<select>` element


packages/vanilla/src/components/select/select.stories.js
```diff
@@ -0,0 +1,106 @@
...
+      name: '`disabled` Attribute',
+      description: 'Attribute that indicates that the user cannot interact with the control.',
+      control: { type: 'boolean' }
+    }
+  },
+  decorators: [
+    (Story, context) => `
+      <form class="cbp-form">
+        <label for=${context.args.labelFor} class="cbp-input__label">${context.args.label}</label>
+        <p class="cbp-input__description" ${!context.args.required ? '' : 'hidden'}>${context.args.inputDescription}</p>
```
dgibson666, line 54: I don't see a `required` control that would set this. See my higher-level comment about this though.

packages/vanilla/src/components/select/select.stories.js
```diff
@@ -0,0 +1,106 @@
...
+        ${Story().outerHTML || Story()}
+      </form>
+    `
+  ]
+}
+
+const Template = ({ labelFor, formControlName, placeholderOption, disabled, optionsObj: { option1, option2, option3, option4, option5} }) => (
+  `
+    <select class="cbp-input__select" name=${formControlName} id=${labelFor} ${disabled ? 'disabled' : ''}>
+      <option value="" selected>${placeholderOption}</option>
```
dgibson666, line 65: We don't need `selected` on the first option. It will be the default anyway.

The presentation of this option looks like something is selected. I think we need to establish a best practice here rather than make it an editable control/option. At SSA we used "--" for the default/no selection because it cannot be mistaken for a selection. USWDS uses "- Select -" which is also common. Because the label says what the thing is being selected, we don't need to repeat it here though - either of the above options would be a better choice and we should not let the user customize it in Storybook to make bad decisions.

packages/vanilla/src/components/select/select.stories.js
dannyk3941: Forgot this control, should be added now along with the `aria-required` attribute. Toggling this control will show the css of the `:invalid` pseudo-class but regarding your comment below, we should revisit/discuss and finalize how to implement this along with the error message class.

packages/vanilla/src/components/select/select.stories.js
dannyk3941: Removed the control and replaced with `-- Select --`

dannyk3941:
We have tickets for removing placeholders for inputs and along with that we should discuss/clarify story controls. This could be a good time to discuss the other attributes like`required`, `aria-required`, `:invalid` pseudo-class and the error message class.

packages/vanilla/src/components/select/select.stories.js
dgibson666: Yeah, I'm not sure why it would show the `:invalid` styling. Is this because you're still using the `required` attribute as well? This is certainly not desired behavior on an unfilled form and would be another reason not to use `required`.

packages/vanilla/src/components/select/select.stories.js
```diff
@@ -102,5 +101,6 @@ Select.args = {
       label: 'Port of Zoolily'
     },
   },
-  disabled: false
+  disabled: false,
```
dgibson666, line 104: Why not leave these unset? That should be the default state, IMO. Just don't pass them as args unless you need to set it. Not going to hold this up over this one, so your call if you update or not.

dgibson666 (approved):
Looks good overall - see my one comment about leaving boolean controls unset.

PR #38, opened 2023-06-20
Refactor/banner cards

dannyk3941:
#### What's this PR do?
- Refactor banner card pattern and add story controls
- Add story controls for General Card patterns

### DIFF
--- a/packages/vanilla/src/components/card/bannerCard.stories.js
+++ b/packages/vanilla/src/components/card/bannerCard.stories.js
@@ -19,21 +19,28 @@ export default {
         'success', 
         'danger'
       ]
-    }
+    },
+    showIcon: {
+      name: 'Show Icon',
+      description: 'Show an icon in the title',
+      control: 'boolean'
+    },
   }
 };
 
+const icons = {
+  default: '<i class="fas fa-info"></i>',
+  info: '<i class="fas fa-info-circle"></i>',  
+  success: '<i class="fas fa-check"></i>',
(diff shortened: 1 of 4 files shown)


packages/vanilla/src/components/card/bannerCard.stories.js
```diff
@@ -0,0 +1,47 @@
...
+  parameters: {
+    layout: 'centered'
+  },
+  argTypes: {
+    title: {
+      name: 'Title',
+      description: 'Set the title in the banner area of the card',
+      control: 'text'
+    },
+    modifier: {
```
dgibson666, line 12: Can we call this "Variant" instead of "Modifier"?

dgibson666 (commented):
* Can we call the control "Variant" instead of "Modfier"? (commented in code as well)
* I think the story should use the info-circle icon. Using a letter-form icon next to text doesn't work well.
* Should the success/danger variants use different icons as well?
* The padding of the content region is doubling up with the margin from the paragraph. Perhaps a `:last-child` selector is needed to zero out the `padding-bottom` and `margin-bottom`.

Regarding the last item, another option would be to apply the light gray background to the parent wrapper and use margins on the content area. That way the margins will collapse without a `:last-child` selector.

packages/vanilla/src/components/card/bannerCard.stories.js
dannyk3941: Changed to 'variant'

dgibson666 (approved):
Looks good!

PR #39, opened 2023-06-20
Refactor/modal

dannyk3941:
#### What's this PR do?
- Refactor modal to use `<dialog>` 
- Focus trapping is not in this PR
- Add README to /vanilla package
- Update /vanilla `package.json`


packages/vanilla/src/components/modal/modal.md
```diff
@@ -0,0 +1,39 @@
...
+### Purpose
+
+- A Modal disables page content and focuses the user’s attention on a single task or message. (*Only the content in the modal can be interacted with.*)
+
+- Modal vs. Dialog - The key difference between a modal and a dialog is that modals will render the background `inert` meaning no elements "behind" the modal are interactive. 
+
+- *Non-modal dialogs allow interaction with content outside of the dialog*
+
+### Functional Requirements
+
```
dgibson666, line 12: These are very granular for functional requirements. I'd move most of this down to the technical specifications. Functional requirements should be high-level in plain language.

packages/vanilla/src/components/modal/modal.stories.js
```diff
@@ -1,182 +1,75 @@
...
+const ModalTemplate = ({ heading, modalID, headingID, describedByID, controlsLayout }) => {
   return `
-    <button class="cbp-btn cbp-btn__primary cbp-hidden" data-modal-target="form-modal" onclick="openModal('form-modal', this)">Form Modal</button><br>
-    <div id="form-modal" role="dialog" role="dialog" aria-labelledby="form-modal-label" aria-modal="true">
-      <div class="cbp-modal__header">
-        <i class="fas fa-comment"></i>
-        <h5 class="cbp-modal__title" id="form-modal-label">Feedback</h5>
-      </div>
-      <hr>
+    <button type="button" class="cbp-btn cbp-btn__primary" data-modal="${modalID}">Open Model</button>
```
dgibson666, line 56: "Modal", not "Model" for the button label

packages/vanilla/src/sass/components/_modal.scss
```diff
@@ -1,146 +1,69 @@
...
-  display: block;
-}
-
-.cbp-modal, [role="dialog"] {
-  position: fixed;
-  left: 50%;
-  bottom: 2rem;
-  transform: translateX(-50%);
+.cbp-modal,
+dialog,
```
dgibson666, line 2: I would not include the `dialog` tag or `role="dialog"` here because there are other uses than for a full-screen dialog/modal that may be affected.

packages/vanilla/src/sass/components/_modal.scss
```diff
@@ -1,146 +1,69 @@
...
-  align-items: center;
-  padding-top: 20px;
-  padding-left: 16px;
-  padding-right: 16px;
-  padding-bottom: 16px;
-  color: var(--cbp-color-text-darker);
+  border-radius: var(--cbp-border-radius-softest);
+  border-width: 0;
+  box-shadow: var(--cbp-shadow-level-1-down);
+  width: 22rem;
```
dgibson666, line 8: A fixed width of 22rem seems very heavy-handed and may like a bad idea for forms in a modal.

dgibson666 (commented):
There's a lot to unpack here without discussing or working from set requirements. I'm fine with discussing and leaving however much of this you want for future iterations:

* Using the native `dialog` tag limits Safari compatibility to version 15.4 (Released 2022-03-15) and up, which is pretty recent. It should probably be noted specifically. It might be worth investigating the polyfill (https://github.com/GoogleChrome/dialog-polyfill).
* I'm also curious as to how the screen reader support holds up, as I've never tested a native dialog on all platforms/devices.
* I don't see any code managing focus - inside of the dialog or returning to the control when it's closed. Is this all automatic with the `dialog` tag?
* It's been common functionality since before the `dialog` tag that clicking the backdrop also closes the modal. This can be implemented with a little bit of JS, such as https://codepen.io/aslhnates/pen/KKKpmpd
* I'm really confused as to why none of the WAI examples use `aria-haspopup="dialog"` - this seems to be exactly what it's for. Any time a button does something that has UI implications, it typically has ARIA attributes that provide that context to screen readers (haspopup, expanded, controls, etc.). I would recommend adding that to the control.
* I don't think the Form Modal meets the description of `role="alertdialog"` - a similar example on the WAI site does not use that role. We should probably differentiate that somehow, either by separating the story from the alerts and/or making a control for the role for the general modal story.
* At SSA, there was more granularity between a modal and full-screen overlay dialog. A true modal forces you to take an action and pressing ESC and clicking the backdrop don't close it. There could very well be applications that need this sort of functionality, so it may be worth discussing.

I'm amazed at how little code is required as a result of using the native `dialog` tag - that's a big win!

dannyk3941:
> There's a lot to unpack here without discussing or working from set requirements. I'm fine with discussing and leaving however much of this you want for future iterations:
> 
> * Using the native `dialog` tag limits Safari compatibility to version 15.4 (Released 2022-03-15) and up, which is pretty recent. It should probably be noted specifically. It might be worth investigating the polyfill (https://github.com/GoogleChrome/dialog-polyfill).
> * I'm also curious as to how the screen reader support holds up, as I've never tested a native dialog on all platforms/devices.
> * ~~I don't see any code managing focus - inside of the dialog or returning to the control when it's closed. Is this all automatic with the `dialog` tag?~~
> * ~~It's been common functionality since before the `dialog` tag that clicking the backdrop also closes the modal. This can be implemented with a little bit of JS, such as https://codepen.io/aslhnates/pen/KKKpmpd~~
> * ~~I'm really confused as to why none of the WAI examples use `aria-haspopup="dialog"` - this seems to be exactly what it's for. Any time a button does something that has UI implications, it typically has ARIA attributes that provide that context to screen readers (haspopup, expanded, controls, etc.). I would recommend adding that to the control.~~
> * ~~I don't think the Form Modal meets the description of `role="alertdialog"` - a similar example on the WAI site does not use that role. We should probably differentiate that somehow, either by separating the story from the alerts and/or making a control for the role for the general modal story.~~
> * At SSA, there was more granularity between a modal and full-screen overlay dialog. A true modal forces you to take an action and pressing ESC and clicking the backdrop don't close it. There could very well be applications that need this sort of functionality, so it may be worth discussing.
> 
> I'm amazed at how little code is required as a result of using the native `dialog` tag - that's a big win!

Add this information to ticket to discuss further and strikethough comments are what's addressed in this PR. `<dialog>` has the focus handled within the element to circle back to modal choices within modal and back to the control that opened the modal.

packages/vanilla/src/sass/components/_modal.scss
dannyk3941: removed

packages/vanilla/src/components/modal/modal.stories.js
dannyk3941: fixed

packages/vanilla/src/sass/components/_modal.scss
dannyk3941: leaving this fixed width for now.

dgibson666 (approved):
Updates look good - can discuss and iterate on the rest.

PR #47, opened 2023-07-27
Feat/structured list

dannyk3941:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26508,7 +26508,7 @@
     "ux-guidelines": {
       "version": "file:packages/ux-guidelines",
       "requires": {
-        "@11ty/eleventy": "*"
+        "@11ty/eleventy": "^2.0.1"
       }
     },
     "v8-compile-cache": {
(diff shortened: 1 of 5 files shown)


packages/vanilla/src/components/structured-list/structured-list.md
```diff
@@ -0,0 +1,34 @@
...
+
+Displays a list with a considerable amount of similar or related content.
+
+## Functional Requirements
+
+* Has an overall wrapper `<div>` which contains the structured list header, structured list.
+* Structured list header is a required part of the structured list pattern and the text inside describes the number of results, any filters applied and time stamp of when the list was updated. 
+* The structured list is a `<ul>` element for the list and `<li>` for the list items.
+* Developers can add a `.cbp-structured-list--striped` to the `<ul>` element to stripe the list. 
+* More complex structured lists have more features:
```
dgibson666, line 13: I would move this bullet and sub-bullets to "Usage notes and Additional Considerations" because it's not really a functional requirement and is likely referring to external components (e.g., pagination) that get integrated.

dgibson666 (approved):
Other than possibly moving the story argTypes to the top of the file and fleshing out the docs a bit, I think this is fine. Those can be done later.

PR #48, opened 2023-07-28
Bugfix/button design

dgibson666:
* Updated button CSS per updated design specs
* Removed hover states on disable buttons
* Updated cursor to "not-allowed" on disabled buttons
* Reordered some CSS to cascade properly

### DIFF
--- a/packages/vanilla/src/sass/components/button/_base.scss
+++ b/packages/vanilla/src/sass/components/button/_base.scss
@@ -16,41 +16,33 @@
     margin-right: var(--cbp-space-2x);
   }
 
-  &:focus {
-    background: var(--cbp-color-interactive-focus-dark);
-    border-color: var(--cbp-color-interactive-focus-dark);
-    border-style: solid;
-    color: var(--cbp-color-text-lightest);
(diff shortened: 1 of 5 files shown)


packages/vanilla/src/sass/components/button/_base.scss
```diff
@@ -12,44 +12,37 @@
   min-height: 2.25rem;
 
   & > i {
+    font-size: 1rem;
```
dannyk3941, line 15: Looks like this is increasing the height of the button to 37px, I think this can be removed since buttons get their font size from the overall default font size.

dannyk3941 (changes requested):
This should be good to go after fixing the one minor comment and fixing the merge conflict in _css-variables.scss

packages/vanilla/src/sass/components/button/_base.scss
dgibson666: I'm going to just comment this out as a reminder then. Jeremy is specing these out as individual SVGs rather than fonts now, so the plan was to size them according to spec and update the selector when we switch to SVGs.

dannyk3941 (approved):

PR #56, opened 2023-08-17
Bugfix/accordion css

dgibson666:
* Restrict click handler to the accordion controls - clicking accordion contents should not collapse the item
* Replace "base" colors with proper tokens
* Add `.cbp-tag` selector to the CSS to behave the same as badges (for now)
* Restrict CSS and transition/animation to the chevron icon (was impacting the icon inside of a tag for the app directory)


dannyk3941 (approved):
Good catch on the bugs! This is good to go.

PR #59, opened 2023-08-24
Feat/pagination css

dannyk3941:
#### What's this PR do?
- Pagination CSS 
- Functionality will be created in separate PR


packages/vanilla/src/components/pagination/pagination.md
```diff
@@ -0,0 +1,36 @@
+# Panel
+
+## Purpose
+
+Pagination is used for splitting up content or data into several pages with a control for navigating to the next or previous page. 
+
```
dgibson666, line 6: A Purpose should be one line. The following 2 lines are part of the functional requirements.

packages/vanilla/src/components/pagination/pagination.stories.js
```diff
@@ -0,0 +1,40 @@
+export default {
+  title: 'Patterns/Pagination'
+};
+
+const Template = () => {
+  return `
+    <div class="cbp-pagination">
+      <div class="cbp-pagination__items">
+        <select aria-label="Select Item Amount">
```
dgibson666, line 9: How about a label of "Items per page"? I think that's more clear.

packages/vanilla/src/components/pagination/pagination.stories.js
```diff
@@ -0,0 +1,40 @@
...
+        <span class="cbp-pagination__item-count">
+          1-10 of 103 Items
+        </span>
+      </div>
+      <div class="cbp-pagination__pages">
+        <button class="cbp-pagination__prev" type="button" aria-label="Previous Page">
+          <i class="fas fa-chevron-left"></i>
+        </button>
+        <select aria-label="Select Page">
+          <option value="1">Pg. 1 of 1000</option>
```
dgibson666, line 25: I don't think abbreviating Page as "Pg." is buying us anything for the cognitive burden it creates (especially in screen readers). It's saving so little space, I'd just spell it out, if it's needed at all.

packages/vanilla/src/sass/components/_pagination.scss
```diff
@@ -1,153 +1,67 @@
...
-  margin-left: auto;
-}
+@use './button/base';
+@use './button/secondary';
+@use './select';
 
 .cbp-pagination {
   align-items: center;
   display: flex;
   justify-content: space-between;
```
dgibson666, line 8: I would add `flex-wrap: wrap` to this. They are overlapping at mobile size even though you've hidden the page size dropdown.

dgibson666 (changes requested):
Mostly minor comments, below.

dgibson666 (approved):
Per our discussion, I think the chevrons for the two selects should be the same and they should always point down when closed and up when open (consistent with accordion, expand, menu, etc.), regardless of directionality. Since there's more to do functionally, we can discuss and circle back on those, so approving this as the initial cut.

PR #63, opened 2023-09-05
Bugfix/banner refactor2

dgibson666:
* Created banner requirement docs
* Add controls to banner stories
* Refactored banner CSS to use design tokens/variables
* Merge remote-tracking branch 'origin/develop' in
* Updated banners to use responsive spacing(padding)

### DIFF
--- a/packages/vanilla/src/components/banner/banner.stories.js
+++ b/packages/vanilla/src/components/banner/banner.stories.js
@@ -29,14 +29,14 @@ const SecurityBannerTemplate = ({headingLevel, headingText, bannerText}) => {
       <div>
         <div>
           <img src="../assets/images/cbp-icon/SECURITY ICON - TS SCI.svg" alt="security logo">
-          <h6 class="cbp-banner__info-title d-none-lg">
-            <i class="fas fa-exclamation-triangle"></i> Infosec Classification: TS/SCI
-          </h6>
+          <${headingLevel} class="cbp-banner__info-title d-none-lg">
+            <i class="fas fa-exclamation-triangle"></i> ${headingText}
+          </${headingLevel}>
         </div>
         <div>
-          <h6 class="cbp-banner__info-title d-none-sm">
-            <i class="fas fa-exclamation-triangle"></i> Infosec Classification: TS/SCI
-          </h6>
+          <${headingLevel} class="cbp-banner__info-title d-none-sm">
+            <i class="fas fa-exclamation-triangle"></i> ${headingText}
+          </${headingLevel}>
           <p class="cbp-banner__text">
             ${bannerText ? bannerText : `This page contains information, in whole or part, that is marked <span class="cbp-text-bold cbp-text-italic">CLASSIFIED: TOP SECRET (TS)</span>. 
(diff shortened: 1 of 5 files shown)


packages/vanilla/src/sass/components/banner/_mixin.scss
```diff
@@ -1,53 +1,42 @@
...
-
-    @media (min-width: 1440px) {
-      margin: 0 28px;
-    }
   }
 }
 
 @mixin banner-text($text-color) {
   color: $text-color;
   font-size: 14px;
```
dannyk3941, line 21: Replace with css variable

packages/vanilla/src/sass/components/banner/_mixin.scss
```diff
@@ -1,53 +1,42 @@
...
+  margin-bottom: var(--cbp-space-4x);
+  max-width: 100%; 
 }
 
 @mixin banner-title($text-color) {
   color: $text-color;
-  font-size: 20px;
-  font-weight: 500;
-  line-height: 1.5;
+  font-size: var(--cbp-font-size-masthead-2);
```
dannyk3941, line 31: Same as the previous `font-size` comment for the Banner titles.

dannyk3941 (changes requested):

packages/vanilla/src/sass/components/banner/_mixin.scss
dgibson666: The other was actually the icon size. Updated this one to heading-md, which is 20px per design spec.

packages/vanilla/src/sass/components/banner/_mixin.scss
dgibson666: I just removed this and the weight, which are identical to body text, which this should inherit since it's just a `p` tag.

packages/vanilla/src/sass/components/banner/_info.scss
```diff
@@ -2,18 +2,21 @@
...
       align-items: center;
+      flex-shrink: 0;
+      color: var(--cbp-color-info-darker);
       background-color: var(--cbp-color-info-light);
       border-radius: 50%;
-      color: var(--cbp-color-info-darker);
-      display: flex;
-      font-size: 24px;
-      height: 48px;
+      font-size: var(--cbp-font-size-masthead-2);
```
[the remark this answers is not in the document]
dgibson666: This is actually the icon size. I replaced this with the space token and added a comment about it. It will be refactored to height/width when we move to SVGs.

packages/vanilla/src/components/banner/banner.stories.js
```diff
@@ -38,10 +38,10 @@ const SecurityBannerTemplate = () => {
             <i class="fas fa-exclamation-triangle"></i> Infosec Classification: TS/SCI
           </h6>
```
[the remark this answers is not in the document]
dgibson666: Oops, looks like an oversight on my part. Fixed.

dannyk3941 (approved):

PR #65, opened 2023-09-07
Fix/usa banner padding

dannyk3941:
#### What's this PR do?
- Add `.cbp-fluid-container` to usa-banner css and remove `cbp-grid-container` classes
- Fix usa banner heights in mobile and larger viewports

### DIFF
--- a/packages/vanilla/src/sass/components/banner/_usa.scss
+++ b/packages/vanilla/src/sass/components/banner/_usa.scss
@@ -1,4 +1,4 @@
-@use '../../grid/breakpoints' as *;
+@use '../../layout/container' as *;
 
 :root {
   --cbp-usa-banner-color: var(--cbp-color-text-lightest);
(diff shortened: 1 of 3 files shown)


packages/vanilla/src/sass/components/banner/_usa.scss
```diff
@@ -92,4 +94,12 @@
 .banner-active {
   display: flex;
   flex-wrap: wrap;
+}
+
+// 433px is the inline size where the text wraps
+@container usa-banner (min-width: 433px) {
+  .cbp-banner__title {
+    max-height: var(--cbp-space-8x);
```
dgibson666, line 102: this is actually problematic as there are some viewport sizes where there is no space before the expanded content due to this max height being less than the content. Seems to work perfectly fine without it.

dgibson666 (changes requested):
Just the one comment on the max-height and it looks good

dgibson666 (approved):
good to go

PR #66, opened 2023-09-11
Feat/util class doc

dannyk3941:
#### What's this PR do?
- Initial margin utilities class doc (mdx) to storybook
- Delete unused `_z-index.scss` file


packages/vanilla/src/sass/utilities/_spacing.scss
```diff
@@ -4,20 +4,29 @@
...
   1x: var(--cbp-space-1x),
+  half-x: var(--cbp-space-half-x),
+  1px: 1px,
   0: 0
 );
 
 @each $name, $space in $spacing {
   .cbp-margin-#{$name} {
-    margin-left: #{$space} !important;
+    margin: #{$space} !important;
```
dgibson666, line 29: Why do these use !important? This is an anti-pattern and breaks CSS specificity. These are utility classes and should act as such (classes).

dgibson666 (changes requested):
Other than removing !important, I think the tables could benefit from a "value" or "translated value" column at the end that gives relative and/or pixel values for ease of comprehension. Saying .cbp-margin-top-7x = margin-top: var(--cbp-space-7x) isn't helping much as one layer of abstraction pointing to another.

dgibson666 (approved):
looks good now

PR #69, opened 2023-09-18
fix: re-order modules for utlity classes in main.scss and breakdown padding css for fluid-container

dannyk3941:
#### What's this PR do?
- Replace `padding` shorthand in `.cbp-fluid-container`
- Re-order `main.scss` modules so utility classes are last


dgibson666 (approved):
Looks good to me.

PR #71, opened 2023-09-22
Fix/container spacing

dannyk3941:
#### What's this PR do?

- Container padding on mobile breakpoint is supposed to be 20px
- Replace css values with variables in `_application.scss`

### DIFF
--- a/packages/vanilla/src/sass/layout/_container.scss
+++ b/packages/vanilla/src/sass/layout/_container.scss
@@ -1,8 +1,8 @@
 @use './../tokens/breakpoints' as *;
 
 .cbp-fluid-container {
-  padding-left: var(--cbp-space-4x);
-  padding-right: var(--cbp-space-4x);
+  padding-left: var(--cbp-space-5x);
+  padding-right: var(--cbp-space-5x);
 
   @media (min-width: $cbp-breakpoint-md) and (max-width: $cbp-breakpoint-xl) {
     padding-left: var(--cbp-space-8x);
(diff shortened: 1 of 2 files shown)


packages/vanilla/src/sass/layout/_container.scss
```diff
@@ -1,8 +1,8 @@
 @use './../tokens/breakpoints' as *;
 
 .cbp-fluid-container {
-  padding-left: var(--cbp-space-4x);
-  padding-right: var(--cbp-space-4x);
+  padding-left: var(--cbp-space-5x);
+  padding-right: var(--cbp-space-5x);
 
   @media (min-width: $cbp-breakpoint-md) and (max-width: $cbp-breakpoint-xl) {
```
dgibson666, line 7: I know this line hasn't actually changed, but you don't need the upper limiter because you have another media query after it that will take precedence, directly overriding these rules.

packages/vanilla/src/sass/layout/_container.scss
```diff
@@ -1,8 +1,8 @@
 @use './../tokens/breakpoints' as *;
 
 .cbp-fluid-container {
-  padding-left: var(--cbp-space-4x);
-  padding-right: var(--cbp-space-4x);
+  padding-left: var(--cbp-space-5x);
+  padding-right: var(--cbp-space-5x);
```
dgibson666, line 5: 20px seems like a very large amount of padding for a potentially small device. If this is the spec, I think we need to discuss it further. I typically see very little padding at the mobile sizes and use .5rem tops myself. 40px is 10% (or more) of the screen width on a phone.

dgibson666 (changes requested):
One minor thing and one discussion point.

dgibson666 (approved):
We can discuss the outer padding later and update if needed.

PR #73, opened 2023-09-27
Fix/drawer nav content

dannyk3941:
#### What's this PR do?
- Removed old css set for nav items within `.cbp-drawer__content` that is unnecessary 
- Removed drawer "nav menu" and replace when a general nav/link menu is created in future.

### DIFF
--- a/packages/vanilla/src/components/drawer/drawer.stories.js
+++ b/packages/vanilla/src/components/drawer/drawer.stories.js
@@ -36,56 +36,9 @@ const Template = ({drawerPosition}) => {
       </div>
 
       <nav class="cbp-drawer__content">
-        <ul class="cbp-drawer__nav">
-          <li class="cbp-drawer__nav-item--a">
-            <button type="button">
-              <span>
-                <i class="fas fa-home"></i>
-                Default/Home Page (A)
-              </span>
-              <i class="fas fa-angle-up"></i>
-            </button>
-            <ul>
-              <li>
-                <button type="button">Child Page 1 (B)</button>
-              </li>
-              <li>
-                <button type="button">Child Page 2 (B)<i class="fas fa-angle-up"></i></button>
-                <ul>
-                  <li>
-                    <button type="button">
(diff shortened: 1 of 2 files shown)


packages/vanilla/src/components/drawer/drawer.stories.js
```diff
@@ -36,56 +36,9 @@ const Template = ({drawerPosition}) => {
       </div>
 
       <nav class="cbp-drawer__content">
```
dgibson666, line 38: technically you don't need the `nav` tag. If there are button instead of links inside (e.g., filtering), then it's not really nav at all. So if we're going to gut the example, let's make it as dumb as possible.

packages/vanilla/src/components/drawer/drawer.stories.js
```diff
@@ -36,56 +36,9 @@ const Template = ({drawerPosition}) => {
...
-            </button>
-          </li>
-          <li>
-            <button type="button">Sibling Page 2 (A)</button>
-          </li>
-          <li>
-            <button type="button">Sibling Page 3 (A)</button>
-          </li>
-        </ul>
+        <section style="height: 100%; display: flex; justify-content: center; align-items: center;">
```
dgibson666, line 39: Like above, I don't think this is needed. We just need the `.cbp-drawer__content` container and placeholder content as simply as possible.

dgibson666 (changes requested):
see code comments. If we're gutting the contents, make it as basic as possible.

packages/vanilla/src/components/drawer/drawer.stories.js
dannyk3941: switched with a `<div>` tag

packages/vanilla/src/components/drawer/drawer.stories.js
dannyk3941: removed

dgibson666 (approved):
looks good

PR #77, opened 2023-10-03
Refactor/decision card updates

dannyk3941:
#### What's this PR do?
- Updates to Decision Card from QA
1. Fix button hover state css
2. Add Tertiary button/action for decision cards (Tertiary button styling is only for decision cards)
3. Add storybook object control to customize decision card buttons (label, type, and disabled)
4. Fix danger card heading text color to $text-darker
5. Clean up CSS for cards

### DIFF
--- a/packages/vanilla/src/components/card/decisionCard.stories.js
+++ b/packages/vanilla/src/components/card/decisionCard.stories.js
@@ -20,6 +20,11 @@ export default {
       control: 'radio',
       options: ['single', 'double', 'triple', 'quadruple']
     },
+    actionsConfig: {
+      name: 'Configure Actions Layout',
+      description: 'Configure card button labels and types. <br /> Available button types: `primary`, `secondary`, `tertiary` and `danger`',
+      control: 'object'
+    },
     danger: {
       name: 'Danger',
       description: 'Display the "danger" state modifier of the card',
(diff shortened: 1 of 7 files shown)


packages/vanilla/src/components/card/decisionCard.stories.js
```diff
@@ -118,21 +119,47 @@ const ProfileCardTemplate = ({ title, danger, actionsLayout }) => {
...
-      ${renderActions(actionsLayout)}
+      ${renderActions(actionsLayout, actionsConfig)}
     </div>
   `;
 };
 
+const actionsConfig = {
+  btn1: {
+    label: 'Label',
+    type: 'primary',
```
dgibson666, line 130: This control is "color" in the buttons stories, so I would use "color" instead of "type", which is a reserved attribute on buttons anyway.

packages/vanilla/src/components/card/decisionCard.stories.js
```diff
@@ -2,111 +2,112 @@ export default {
...
   `;
 };
 
-const ProfileCardTemplate = ({ title, danger, actionsLayout }) => {
+const ProfileCardTemplate = ({ title, danger, actionsLayout, actionsConfig }) => {
   return `
     <div class="cbp-card ${danger ? 'cbp-card--danger' : ''} cbp-card--decision">
       <div class="cbp-card__content cbp-padding-0">
-        <img src="assets/images/profile-page/passenge-photo-v2.jpg" class="cbp-card__img-top" alt="portrait image of person" width="312px" />
+        <img src="assets/images/profile-page/passenge-photo-v2.jpg" class="cbp-card__img-top" alt="portrait image of person" style="width: 100%" />
```
dgibson666, line 99: should the `style` be part of the CSS applied to the class specified?

dgibson666 (changes requested):
Overall looks good - just a couple of minor comments

dgibson666 (approved):
looks good

PR #79, opened 2023-10-04
Add controls to segmented button stories

dgibson666:
* Added controls to segmented buttons stories.

### DIFF
--- a/packages/vanilla/src/components/segmented-button-group/segmented-button-group.stories.js
+++ b/packages/vanilla/src/components/segmented-button-group/segmented-button-group.stories.js
@@ -12,13 +12,11 @@ export default {
 
 
 function generateButtons(buttons) {
-  let html=[];
-  buttons.forEach( ({ label, value, disabled }, i) => {
-    html=[...html, `<button type="button" class="cbp-btn" value="${value}" ${disabled==true ? 'disabled' : ''}>${label}</button>`];
-  });
-  return [...html].join("");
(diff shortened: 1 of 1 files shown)


packages/vanilla/src/components/segmented-button-group/segmented-button-group.stories.js
```diff
@@ -1,32 +1,45 @@
...
 
-const SingleSelectTemplate = () => {
+
+function generateButtons(buttons) {
+  let html=[];
+  buttons.forEach( ({ label, value, disabled }, i) => {
+    html=[...html, `<button type="button" class="cbp-btn" value="${value}" ${disabled==true ? 'disabled' : ''}>${label}</button>`];
+  });
+  return [...html].join("");
+};
```
dannyk3941, line 20: I think this function can be simplified with `.map()` and you won't have to use the spread syntax to push each template literal of the button into the `html` array. Instead of initializing a `let` with an empty array, you can do something like:
```javascript
function generateButtons(buttons) {
  const html = buttons.map(({ label, value, disabled}) => {
    return `<button type="button" class="cbp-btn" value="${value}" ${disabled==true ? 'disabled' : ''}>${label}  </button>`
  })

  return html.join("")
};
```

dgibson666: Good call. I was struggling with making this work inline in the template literal and didn't really revisit it.

dannyk3941 (approved):

PR #80, opened 2023-10-04
Bugfix/footer nav

dgibson666:
* Added button class to the internal footer nav links to provide button letter spacing
* Removed duplicate CSS
* Fixed the vertical padding on the `.cbp-footer__info` element (it was being reduced to 4x at a breakpoint, but should remain 5x)

### DIFF
--- a/packages/vanilla/src/components/footer/footer.stories.js
+++ b/packages/vanilla/src/components/footer/footer.stories.js
@@ -10,10 +10,10 @@ const InternalTemplate = () => {
     <footer class="cbp-footer">
       <nav>
         <ul>
-          <li><a href="#" class="cbp-btn">App Overview</a></li>
-          <li><a href="#" class="cbp-btn">Trainings</a></li>
-          <li><a href="#" class="cbp-btn">FAQ's</a></li>
-          <li><a href="#" class="cbp-btn">Release Notes</a></li>
+          <li><a href="#">App Overview</a></li>
+          <li><a href="#">Trainings</a></li>
+          <li><a href="#">FAQ's</a></li>
+          <li><a href="#">Release Notes</a></li>
         </ul>
       </nav>
       <div class="cbp-footer__info">
--- a/packages/vanilla/src/components/footer/footer.stories.js
+++ b/packages/vanilla/src/components/footer/footer.stories.js
@@ -10,10 +10,10 @@ const InternalTemplate = () => {
     <footer class="cbp-footer">
       <nav>
         <ul>
-          <li><a href="#">App Overview</a></li>
-          <li><a href="#">Trainings</a></li>
-          <li><a href="#">FAQ's</a></li>
-          <li><a href="#">Release Notes</a></li>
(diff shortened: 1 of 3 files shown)


packages/vanilla/src/components/footer/footer.stories.js
```diff
@@ -10,10 +10,10 @@ const InternalTemplate = () => {
...
       <nav>
         <ul>
-          <li><a href="#">App Overview</a></li>
-          <li><a href="#">Trainings</a></li>
-          <li><a href="#">FAQ's</a></li>
-          <li><a href="#">Release Notes</a></li>
+          <li><a href="#" class="cbp-btn">App Overview</a></li>
+          <li><a href="#" class="cbp-btn">Trainings</a></li>
+          <li><a href="#" class="cbp-btn">FAQ's</a></li>
+          <li><a href="#" class="cbp-btn">Release Notes</a></li>
```
dannyk3941, line 16: I don't think we should put the `cbp-btn` classes here since they don't look or act like the buttons we have. We're deleting 3-4 css properties by placing the class here and I think it's a little more confusing to place the class in here. I think the only thing that needed fixing was the `letter-spacing` to match the buttons.

packages/vanilla/src/sass/components/footer/_internal.scss
```diff
@@ -161,11 +155,6 @@
     }
   }
 
-  .cbp-footer__info {
-    padding-top: var(--cbp-space-4x);
-    padding-bottom: var(--cbp-space-4x);
-  }
```
dannyk3941, line 167: This spacing probably from before the mock ups changed, looking at mock ups now it looks like there needs to be a padding top for the `.cbp-footer__info` class at `var(--cbp-space-4x)` across all screen widths. The bottom padding however should be verified with Jeremy (@yellowzapp) since that was spaced out for the action bar.

dannyk3941 (changes requested):

packages/vanilla/src/sass/components/footer/_internal.scss
dgibson666: Jeremy's specific comment said that it should be changed to 5x. We can address the action bar issue when we have it and determine its functionality. I definitely wouldn't put a bunch of extra space at the bottom without the floating action bar there.

packages/vanilla/src/components/footer/footer.stories.js
dgibson666: I'll revert it. all those deleted CSS props are part of button, so I feel like it's 6 one way, half a dozen the other. Seems arbitrary we're styling links like buttons.

packages/vanilla/src/sass/components/footer/_internal.scss
dannyk3941: Oh if thats the case then it should what you changed it to.

dannyk3941 (approved):

PR #81, opened 2023-10-06
feat: text and fill divider classes

dannyk3941:
#### What's this PR do?

- Created text and fill divider classes with story


dgibson666 (changes requested):
If I set the divider size, but change the heading level, I see the size changing due to the different line-heights. I think this needs to be normalized/set at each level if we are providing consistent styling based on the divider class regardless of heading level. We should probably run this back to Jeremy for discussion.

Everything else looks fine.

dgibson666 (approved):
Approving this as-is. We can circle back on the line-height issue if needed.

PR #82, opened 2023-10-06
Chore/react pkg re org

dannyk3941:
#### What's this PR do?
- Re-organize `react-components` folder structure
- Remove duplicate Eslint packages and fix eslint configuration in the root `package.json`


dgibson666 (approved):
good start. we can continue iterating from here.

PR #86, opened 2023-10-23
Fix/link and card

dannyk3941:
#### What's this PR do?
1. Remove link font-weight causing bounciness and fix font-color
2. Fix border issue with decision cards
3. Fix bug from disabled/active buttons


dgibson666 (approved):
This appears to fix what you set out to fix. While testing, I did notice that a disabled link-button (anchor styled as a link, e.g,. `<a role="link" aria-disabled="true" class="cbp-btn cbp-btn__primary">`) is still showing up dark blue instead of being styled as a disabled button. That needs to be fixed at some point.

PR #88, opened 2023-10-25
Fix/general card

dannyk3941:
#### What's this PR do?
1. Fix border color for default general card
2. Added warning variant to general card
3. Removed "small" general card story
4. Added show/hide control in general card for the card title

**_Need Help w/_**
Storybook control for body text is added to the general card story, do you think we need this?

### DIFF
--- a/packages/vanilla/src/components/card/generalCard.stories.js
+++ b/packages/vanilla/src/components/card/generalCard.stories.js
@@ -17,17 +17,25 @@ export default {
       description: 'Set the modifier class of the card',
       control: 'select',
       options: ['default', 'info', 'success', 'danger', 'warning']
+    },
+    showIcon: {
+      name: 'Show Icon',
+      description: 'Show an icon in the title',
+      control: 'boolean'
     }
   }
 };
 
-const DefaultGeneralCardTemplate = ({ title, modifier }) => {
+const DefaultGeneralCardTemplate = ({ title, modifier, showIcon }) => {
   return `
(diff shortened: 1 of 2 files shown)


packages/vanilla/src/components/card/generalCard.stories.js
```diff
@@ -16,51 +16,52 @@ export default {
...
-const DefaultGeneralCardTemplate = ({ title, modifier }) => {
+const DefaultGeneralCardTemplate = ({ title, modifier, showIcon, bodyText }) => {
   return `
     <div class="wrapper" style="display: grid; grid-template-columns: 432px">
       <div class="cbp-card ${modifier != 'default' ? `cbp-card--${modifier}` : ''}">
         <div class="cbp-card__content">
           <div class="cbp-card__header">
-            <h4 class="cbp-card__title"><i class="fas fa-info-circle cbp-margin-right-2x"></i>${title}</h4>
+            <h4 class="cbp-card__title">
+              ${showIcon ? '<i class="fas fa-info-circle cbp-margin-right-2x"></i>' : ''}
```
dgibson666, line 41: Sorry this is nitpicky, but make sure to put the icon and the heading/title on the same line next to each other. The line break is showing up as a space and adding to the margin-right making a larger gap than intended. I've fixed a bunch of these when editing stories myself.

dgibson666 (changes requested):
Everything looks good other than the one small code comment.

dgibson666 (approved):
looks good

PR #90, opened 2023-10-27
Fix/banner cards

dannyk3941:
#### What's this PR do?
- Fix banner card title text-color, background colors, and spacing
- Add storybook control for body text
- Add warning variant for banner cards


dgibson666 (approved):
Looks good. 

You didn't change this in this PR, but I noticed that these stories use a control named "variant" and the General Card uses "modifier". They should both be "color" technically. We can leave it here, but for components, it will be both more important to be consistent (as a prop, it's part of the API) and also more obvious to catch these things.

PR #97, opened 2023-11-15
fix: import wrong addon in storybook preview.js for theme

dannyk3941:
#### What's this PR do?
- Storybook throwing an error for `@storybook/styling-addon` that is not even in the project.


dgibson666 (approved):
I wasn't seeing any errors, but this is working for me as well.

PR #150, opened 2024-06-21
Bugfix/panel header

dgibson666:
* Update the way typography sets dark mode colors.
* Update panel and drawer stories, removing cbp-typography context prop in the panel header (no longer needed).
* Remove panel heading color dark mode variable that doesn't change.

Verified expected behavior in Panel, Drawer, and Typography stories, in both light and dark modes and toggling the context property on the parent component (this gets pretty confusing with all the combos).


bagrub (approved):
Looks good

PR #153, opened 2024-06-27
Update card styles to support dark mode theme

bagrub:

### DIFF
--- a/packages/web-components/src/components/cbp-card/cbp-card.scss
+++ b/packages/web-components/src/components/cbp-card/cbp-card.scss
@@ -21,7 +21,9 @@
   --cbp-card-color-border-dark: var(--cbp-color-gray-cool-50);
   --cbp-card-color-title: var(--cbp-color-text-darker);
   --cbp-card-color-title-dark: var(--cbp-color-text-lighter);
-  
+  --cbp-card-color-text: var(--cbp-color-text-darker);
+  --cbp-card-color-text-dark: var(--cbp-color-text-lightest);
+
   --cbp-banner-card-color: var(--cbp-color-gray-cool-60);
   --cbp-banner-card-color-dark: var(--cbp-color-gray-cool-20); 
   --cbp-banner-card-color-title: var(--cbp-color-text-lighter);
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/cbp-card/cbp-card.scss
```diff
@@ -1,19 +1,48 @@
...
+[data-cbp-theme=light] cbp-card[context*=dark],
+[data-cbp-theme=dark] cbp-card:not([context=dark-inverts]):not([context=light-always]) {
+  --cbp-card-color-background: var(--cbp-card-color-background-dark);
+  --cbp-card-color-border: var(--cbp-card-color-border-dark);
+  --cbp-card-color-title: var(--cbp-card-color-title-dark);
+  
+  --cbp-banner-card-color: var(--cbp-banner-card-color-dark);
+  --cbp-banner-card-color-title: var(--cbp-banner-card-color-title-dark);
+  --cbp-banner-card-color-body-background: var(--cbp-banner-card-color-body-background-dark);
+  --cbp-banner-card-border: var(--cbp-banner-card-border-dark)
```
dgibson666: should these be removed or renamed? They shouldn't have "banner" in them.

Edit: wait, this is probably my mistake. I see there's a banner card variation. I was incorrectly assuming copy paste from banner since it was the last one you worked on.

dgibson666 (changes requested):
Good start. You still need to add the context prop to the component and the story though.

dgibson666 (changes requested):
Looking good overall. Last thing missing I think is the dark mode (variables) for the color variants.

packages/web-components/src/components/cbp-card/cbp-card.scss
```diff
@@ -118,38 +168,53 @@ cbp-card {
...
     &[color=info] {
       --cbp-banner-card-color: var(--cbp-color-info-dark);
       --cbp-banner-card-color-body-background: var(--cbp-color-info-lighter);
+      
+      --cbp-banner-card-color-dark: var(--cbp-color-info-light);
+      --cbp-banner-card-color-body-background-dark: var( --cbp-color-info-darker);
     }
-
+    
     &[color=success] {
```
dgibson666, line 192: OK, while I'm in light mode, but change context to dark, the text for every color except warning is not unreadable. So I think you need to set the text colors explicitly. If they are the same for all colors other than warning, set them once under the `&[varaint=banner]` selector.

dgibson666 (changes requested):
Almost there! Banners in light mode with dark context are still not right though. See my code comment.

dgibson666 (approved):
Looks good now.

PR #154, opened 2024-07-03
Application Header dark mode

bagrub:


dgibson666 (changes requested):
It looks like it's working and everything here looks good. However, you need to implement the context property in the component itself. As it is, `context` is just some arbitrary attribute that storybook is adding.

dgibson666 (approved):
looks good

PR #159, opened 2024-07-11
Typography: update dividers per feedback and convert CSS to logical properties

dgibson666:
* Update dividers per feedback
* Convert Typography CSS to logical properties


bagrub (approved):
Changes look good

PR #160, opened 2024-07-11
Structured list dark mode

bagrub:

### DIFF
--- a/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
+++ b/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
@@ -1,5 +1,12 @@
 /** todo: update doc
- * @prop 
+ * @prop --cbp-structured-list-item-padding: var(--cbp-space-4x);
+
+ * @prop  --cbp-structured-list-item-color-background-danger: var(--cbp-color-danger-lighter);
+ * @prop  --cbp-structured-list-item-color-background-danger-dark: var(--cbp-color-danger-darker);
+
+ * @prop  --cbp-structured-list-item-color-background-selected: var(--cbp-color-interactive-selected-light);
+ * @prop  --cbp-structured-list-item-color-background-selected-dark: var(--cbp-color-interactive-selected-dark);
+
 */
 :root {
   --cbp-structured-list-item-padding: var(--cbp-space-4x);
@@ -21,11 +28,11 @@
 cbp-structured-list-item {
   display: block;
 
-  &[color=danger]:not([selected]) { //todo: update to component vars
+  &[color=danger]:not([selected]) { //techdebt: update to component vars
     background-color: var(--cbp-structured-list-item-color-background-danger) !important;
   }
 
-  &[selected] { //todo: update to component vars
+  &[selected] { //techdebt: update to component vars
     background-color: var(--cbp-structured-list-item-color-background-selected) !important;
   }
 }
--- a/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
+++ b/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
@@ -1,11 +1,31 @@
+/** todo: update doc
+ * @prop 
+*/
+:root {
+  --cbp-structured-list-item-padding: var(--cbp-space-4x);
+
+  --cbp-structured-list-item-color-background-danger: var(--cbp-color-danger-lighter);
+  --cbp-structured-list-item-color-background-danger-dark: var(--cbp-color-danger-darker);
+
+  --cbp-structured-list-item-color-background-selected: var(--cbp-color-interactive-selected-light);
+  --cbp-structured-list-item-color-background-selected-dark: var(--cbp-color-interactive-selected-dark);
+
+}
+
+[data-cbp-theme=light] cbp-structured-list[context*=dark]:not([context=light-always]),
+[data-cbp-theme=dark] cbp-structured-list:not([context=dark-inverts]):not([context=light-always]) {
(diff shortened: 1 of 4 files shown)


packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
```diff
@@ -1,11 +1,31 @@
+/** todo: update doc
+ * @prop 
+*/
+:root {
+  --cbp-structured-list-item-padding: var(--cbp-space-4x);
+
+  --cbp-structured-list-item-color-background-danger: var(--cbp-color-danger-lighter);
```
dgibson666, line 14: The color variants shouldn't be up here. They should be modifying the base variable in their selector.

packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
```diff
@@ -1,11 +1,31 @@
...
+  --cbp-structured-list-item-color-background-danger: var(--cbp-structured-list-item-color-background-danger-dark);  
+  --cbp-structured-list-item-color-background-selected: var(--cbp-structured-list-item-color-background-selected-dark);
+}
+
 cbp-structured-list-item {
   display: block;
 
-  &[color=danger]:not([selected]) {
-    background-color: var(--cbp-color-danger-lighter) !important;
+  &[color=danger]:not([selected]) { //todo: update to component vars
```
dgibson666: Let's discuss

dgibson666 (changes requested):
I have a couple of code comments that may need discussion.

The buttons in the footer/action bar need to be `context="dark-inverts"` instead of "dark-always".

I would go ahead and do the other ticket in here as well, as it is not a super-special variant - it's literally just a case with no header, which triggers a top border being applied. You can make a separate story for it, but that's the only difference. It doesn't even need a variant name.

packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
bagrub: discussed on call, marking as resolved based on that discussion

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
```diff
@@ -98,11 +104,18 @@ cbp-structured-list {
...
     &:hover {
       background-color: var(--cbp-structured-list-color-background-hover) !important;
     }
   }
 
+  /* 
+    *Collection 
+  */
+  
+  div[role=list]:only-child > cbp-structured-list-item:first-of-type {
```
dgibson666: This needs to work for regular `li` as well. I would use: 

```
div[role=list]:only-child > *:first-child
```

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -251,13 +252,68 @@ const StructuredListWithGridTemplate = ({ striped, selectable, showHeader, heade
...
+StructuredListCollection.argTypes = {
+  listItems: {
+    description: 'Configure various aspects of the list items within the structured list.',
+    control: 'object',
+  },
+}
+StructuredListCollection.args = {
+  listItems: [
+    {
+      content: "<div style='display:flex;flex-direction: column; gap: 0.5rem;'><span><cbp-icon name='arrow-right' color='var(--cbp-link-color)' size='1rem'></cbp-icon><cbp-link href='#' target='_self' style='margin-inline-start: 0.5rem'>Internal Link</cbp-link></span><span>Description text</span><span> <cbp-icon name='user' size='1rem'></cbp-icon> <i>https://www.text-link.com/help-me</i></span><cbp-tag class='hydrated' style='width: 50px'> Tag </cbp-tag></div>",
```
dgibson666: remove the `class="hydrated"` from these HTML blurbs - that is added by Stencil's componentDidLoad lifecycle method.

The icons being 1rem are larger than the text (the user icon looks especially big), which is 14px. Icons default to 1em, so I think you shouldn't need a size at all.  Verify with Jeremy.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -251,13 +252,68 @@ const StructuredListWithGridTemplate = ({ striped, selectable, showHeader, heade
...
+      selected: false
+    },
+    {
+      content: "<div style='display:flex;flex-direction: column; gap: 0.5rem;'><span><cbp-icon name='arrow-right' color='var(--cbp-link-color)' size='1rem'></cbp-icon><cbp-link href='#' target='_self' style='margin-inline-start: 0.5rem'>Internal Link</cbp-link></span><span>Description text</span><span> <cbp-icon name='user' size='1rem'></cbp-icon> <i>https://www.text-link.com/help-me</i></span><cbp-tag class='hydrated' style='width: 50px'> Tag </cbp-tag></div>",
+      color: 'default',
+      selected: false
+    },
+  ]
+}
+
```
dgibson666, line 318: You also need to set the `showHeader` arg to false for this story because it's set to true at the high level.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
```diff
@@ -98,11 +104,18 @@ cbp-structured-list {
...
+    border-top: 1px solid var(--cbp-structured-list-border-color);
```
dgibson666: You're using a magic number for the border-width instead of the token and they are different. Should be `var(--cbp-border-size-md);`

dgibson666 (changes requested):
Looks good overall. See code comments for changes.

dgibson666 (commented):
looks good now

dgibson666 (approved):

PR #161, opened 2024-07-12
Bugfix/panel darkmode

dgibson666:
* Implement proper dark mode for panel
* Implement User Preferences drawer styles via CSS API for panel


bagrub (approved):
Approving, Looks good to me

PR #162, opened 2024-07-16
Bugfix/universal header logo

dgibson666:
* Add `base` tag to Storybook preview to fix the universal header story broken images (needs verification after merging/deployment)

* Update universal header story to use "HASHIDX"

* Update the `cbp-hide` component to emit an event when its state is toggled - verified via console.logs. Unable to get event listener working in Storybook, however - will revisit later.


bagrub (approved):
looks good to me

PR #163, opened 2024-07-18
Action Bar

bagrub:
#### Create action bar component & replace structured list footer with action bar

### DIFF
--- a/packages/web-components/src/components/cbp-action-bar/cbp-action-bar.scss
+++ b/packages/web-components/src/components/cbp-action-bar/cbp-action-bar.scss
@@ -28,11 +28,7 @@ cbp-action-bar {
   padding: var(--cbp-space-3x);
   color: var(--cbp-action-bar-color);
   background-color: var(--cbp-action-bar-color-background);
-
-
-   & > *{
-    padding-left: 0;
-   }
+  gap: var(--cbp-space-3x);
 
    & > *[slot="cbp-action-bar-info"]{
     margin-inline-end: auto;
@@ -44,7 +40,7 @@ cbp-action-bar {
     position: fixed;
     bottom: 0;
     left: 0;
-    padding: var(--cbp-responsive-spacing-outer);
+    padding-inline: var(--cbp-responsive-spacing-outer);
     box-shadow: var(--cbp-shadow-level-3-up);
    }
 }
--- a/packages/web-components/src/components/cbp-action-bar/cbp-action-bar.scss
+++ b/packages/web-components/src/components/cbp-action-bar/cbp-action-bar.scss
@@ -7,15 +7,15 @@
  --cbp-action-bar-color-background: var(--cbp-color-gray-cool-70);
  --cbp-action-bar-color-background-dark: var(--cbp-color-gray-cool-5);
 
- --cbp-action-bar-info-color: var(--cbp-color-text-lightest);
- --cbp-action-bar-info-color-dark: var(--cbp-color-text-darkest);
+ --cbp-action-bar-color: var(--cbp-color-text-lightest);
+ --cbp-action-bar-color-dark: var(--cbp-color-text-darkest);
   
 }
 
 [data-cbp-theme=light] cbp-action-bar[context*=dark]:not([context=light-always]),
 [data-cbp-theme=dark] cbp-action-bar:not([context=dark-inverts]):not([context=light-always]) {
   --cbp-action-bar-color-background: var(--cbp-action-bar-color-background-dark);
-  --cbp-action-bar-info-color: var(--cbp-action-bar-info-color-dark);
+  --cbp-action-bar-color: var(--cbp-action-bar-color-dark);
   
 }
 
@@ -23,31 +23,28 @@ cbp-action-bar {
   display: flex;
   align-items: center;
   justify-content: right;
-  height: var(--cbp-space-14x);
+  min-height: var(--cbp-space-14x);
   width: 100%;
   padding: var(--cbp-space-3x);
+  color: var(--cbp-action-bar-color);
   background-color: var(--cbp-action-bar-color-background);
-  box-shadow: var(--cbp-shadow-level-3-up);
+
 
    & > *{
     padding-left: 0;
-
-    &:not(:first-child){
-      padding-left: var(--cbp-space-3x);
-    }
    }
 
    & > *[slot="cbp-action-bar-info"]{
     margin-inline-end: auto;
-    color: var(--cbp-action-bar-info-color);
    }
    /*
    *** Sticky variant position
    */
-   &[variant=sticky] {
+   &[variant=floating] {
     position: fixed;
     bottom: 0;
     left: 0;
     padding: var(--cbp-responsive-spacing-outer);
+    box-shadow: var(--cbp-shadow-level-3-up);
    }
 }
--- a/packages/web-components/src/components/cbp-action-bar/cbp-action-bar.scss
+++ b/packages/web-components/src/components/cbp-action-bar/cbp-action-bar.scss
@@ -1,0 +1,53 @@
+/**
+ * @prop --cbp-action-bar-color-background: var(--cbp-color-grey-cool-70);
+ * @prop--cbp-action-bar-color-background-dark: var(--cbp-color-grey-cool-5);
+*/
+
+:root {
+ --cbp-action-bar-color-background: var(--cbp-color-gray-cool-70);
+ --cbp-action-bar-color-background-dark: var(--cbp-color-gray-cool-5);
+
+ --cbp-action-bar-info-color: var(--cbp-color-text-lightest);
+ --cbp-action-bar-info-color-dark: var(--cbp-color-text-darkest);
+  
+}
+
+[data-cbp-theme=light] cbp-action-bar[context*=dark]:not([context=light-always]),
+[data-cbp-theme=dark] cbp-action-bar:not([context=dark-inverts]):not([context=light-always]) {
+  --cbp-action-bar-color-background: var(--cbp-action-bar-color-background-dark);
+  --cbp-action-bar-info-color: var(--cbp-action-bar-info-color-dark);
+  
+}
+
+cbp-action-bar {
+  display: flex;
+  align-items: center;
+  justify-content: right;
+  height: var(--cbp-space-14x);
+  width: 100%;
+  padding: var(--cbp-space-3x);
+  background-color: var(--cbp-action-bar-color-background);
+  box-shadow: var(--cbp-shadow-level-3-up);
+
+   & > *{
+    padding-left: 0;
+
+    &:not(:first-child){
+      padding-left: var(--cbp-space-3x);
+    }
+   }
+
+   & > *[slot="cbp-action-bar-info"]{
+    margin-inline-end: auto;
+    color: var(--cbp-action-bar-info-color);
+   }
+   /*
+   *** Sticky variant position
+   */
+   &[variant=sticky] {
+    position: fixed;
+    bottom: 0;
+    left: 0;
+    padding: var(--cbp-responsive-spacing-outer);
+   }
+}
--- a/packages/web-components/src/components/cbp-action-bar/cbp-action-bar.scss
+++ b/packages/web-components/src/components/cbp-action-bar/cbp-action-bar.scss
@@ -2,14 +2,14 @@
  * @prop --cbp-action-bar-color-background: var(--cbp-color-grey-cool-70);
  * @prop--cbp-action-bar-color-background-dark: var(--cbp-color-grey-cool-5);
 */
-
+//todo: swap light and dark values so context works correctly
 :root {
- --cbp-action-bar-color-background: var(--cbp-color-gray-cool-70);
- --cbp-action-bar-color-background-dark: var(--cbp-color-gray-cool-5);
+--cbp-action-bar-color-background: var(--cbp-color-gray-cool-5);
+--cbp-action-bar-color-background-dark: var(--cbp-color-gray-cool-70);
 
- --cbp-action-bar-color: var(--cbp-color-text-lightest);
- --cbp-action-bar-color-dark: var(--cbp-color-text-darkest);
-  
+--cbp-action-bar-color: var(--cbp-color-text-darkest);
+--cbp-action-bar-color-dark: var(--cbp-color-text-lightest);
+
 }
 
 [data-cbp-theme=light] cbp-action-bar[context*=dark]:not([context=light-always]),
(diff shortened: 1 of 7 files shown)


packages/web-components/src/components/cbp-action-bar/cbp-action-bar.tsx
```diff
@@ -0,0 +1,27 @@
+import { Component, Host, Prop, h } from '@stencil/core';
+
+@Component({
+  tag: 'cbp-action-bar',
+  styleUrl: 'cbp-action-bar.scss',
+})
+export class CbpActionBar {
+
+  /** variant for type: sticky/inline */
+  @Prop({ reflect: true }) variant: 'inline' | 'sticky'
```
dgibson666: This needs a default, which should probably be 'inline'. I would also rename "sticky" to "floating" because technically you could have a sticky element within a table or list container as well.

packages/web-components/src/components/cbp-action-bar/cbp-action-bar.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  /** Supports adding inline styles as an object */
+  @Prop() sx: any = {};
+
```
dgibson666, line 13: You need to bring in the parts that enable the `sx` property. See any other component for an example. This should always be last - move context above it.

packages/web-components/src/components/cbp-action-bar/cbp-action-bar.tsx
```diff
@@ -0,0 +1,27 @@
...
+  /** Specifies the context of the component as it applies to the visual design and whether it inverts when light/dark mode is toggled. Default behavior is "light-inverts" and does not have to be specified. */
+  @Prop({ reflect: true }) context: "light-inverts" | "light-always" | "dark-inverts" | "dark-always";
+  
+  render() {
+    return (
+      <Host variant={this.variant}>
```
dgibson666: you don't need to put `variant` here. It's automatically reflected.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
```diff
@@ -87,19 +87,12 @@ cbp-structured-list {
...
-    background-color: var(--cbp-structured-list-footer-color-background);
-  }
-
   &[striped] [role=list] > *:nth-child(even) {
     background-color: var(--cbp-structured-list-color-striped-background);
   }
 
   // Style all list items regardless of the element used to wrap/define them (don't leave this to cbp-structured-list-item)
-  div[role=list] > * {
+  div[role=list] > *:not(cbp-action-bar) {
```
dgibson666: this addition shouldn't be needed because the action bar is not inside of the list anyway.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -66,16 +66,13 @@ const StructuredListTemplate = ({ listItems, striped, selectable, showHeader, he
...
             ? `
-                <div slot="cbp-structured-list-footer">
-                  <cbp-flex align-items="center" justify-content="space-between">  
-                    <div>0 items selected.</div>
-                    <div>
-                      <cbp-button fill="ghost" context="dark-inverts" accessibility-text="Delete selected items">Delete</cbp-button>
-                      <cbp-button fill="ghost" context="dark-inverts" accessibility-text="Compare selected items">Compare</cbp-button>
-                    </div>
-                </div>
+                <cbp-action-bar variant='inline'>
```
dgibson666: this should still live in and use the `slot="cbp-structured-list-footer"` It is not a list item and cannot be inside the list.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -124,14 +121,11 @@ const StructuredListItemsTemplate = ({ listItems, striped, selectable, showHeade
...
             ? `
-                <div slot="cbp-structured-list-footer">
-                  <cbp-flex align-items="center" justify-content="space-between">  
-                    <div>0 items selected.</div>
-                    <div>
-                      <cbp-button fill="ghost" context="dark-inverts" accessibility-text="Delete selected items">Delete</cbp-button>
-                      <cbp-button fill="ghost" context="dark-inverts" accessibility-text="Compare selected items">Compare</cbp-button>
-                    </div>
-                </div>
+            <cbp-action-bar variant='inline'>
```
dgibson666: this should still live in and use the slot="cbp-structured-list-footer" It is not a list item and cannot be inside the list.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -248,14 +242,11 @@ const StructuredListWithGridTemplate = ({ striped, selectable, showHeader, heade
...
```
dgibson666: this should still live in and use the slot="cbp-structured-list-footer" It is not a list item and cannot be inside the list.

packages/web-components/src/components/cbp-action-bar/cbp-action-bar.scss
```diff
@@ -0,0 +1,53 @@
...
+    color: var(--cbp-action-bar-info-color);
+   }
+   /*
+   *** Sticky variant position
+   */
+   &[variant=sticky] {
+    position: fixed;
+    bottom: 0;
+    left: 0;
+    padding: var(--cbp-responsive-spacing-outer);
```
dgibson666: this responsive token can only be applied inline, not to the block(top and bottom).

packages/web-components/src/components/cbp-action-bar/cbp-action-bar.scss
```diff
@@ -0,0 +1,53 @@
+/**
+ * @prop --cbp-action-bar-color-background: var(--cbp-color-grey-cool-70);
+ * @prop--cbp-action-bar-color-background-dark: var(--cbp-color-grey-cool-5);
+*/
+
+:root {
+ --cbp-action-bar-color-background: var(--cbp-color-gray-cool-70);
+ --cbp-action-bar-color-background-dark: var(--cbp-color-gray-cool-5);
+
+ --cbp-action-bar-info-color: var(--cbp-color-text-lightest);
```
dgibson666: you can remove the "info" from this. It should just be the color for anything in the action bar. You should have a color and a color-bg (shorten it for consistency).

packages/web-components/src/components/cbp-action-bar/cbp-action-bar.scss
```diff
@@ -0,0 +1,53 @@
...
+  --cbp-action-bar-color-background: var(--cbp-action-bar-color-background-dark);
+  --cbp-action-bar-info-color: var(--cbp-action-bar-info-color-dark);
+  
+}
+
+cbp-action-bar {
+  display: flex;
+  align-items: center;
+  justify-content: right;
+  height: var(--cbp-space-14x);
```
dgibson666: this should be a min-height. we don't set fixed heights because they can break if text is resized or wraps.

packages/web-components/src/components/cbp-action-bar/cbp-action-bar.scss
```diff
@@ -0,0 +1,53 @@
...
+    padding-left: 0;
+
+    &:not(:first-child){
+      padding-left: var(--cbp-space-3x);
+    }
+   }
+
+   & > *[slot="cbp-action-bar-info"]{
+    margin-inline-end: auto;
+    color: var(--cbp-action-bar-info-color);
```
dgibson666: move color up to the top level.

packages/web-components/src/components/cbp-action-bar/cbp-action-bar.scss
```diff
@@ -0,0 +1,53 @@
...
+  height: var(--cbp-space-14x);
+  width: 100%;
+  padding: var(--cbp-space-3x);
+  background-color: var(--cbp-action-bar-color-background);
+  box-shadow: var(--cbp-shadow-level-3-up);
+
+   & > *{
+    padding-left: 0;
+
+    &:not(:first-child){
```
dgibson666: this isn't needed and should be replaced with gap.

packages/web-components/src/components/cbp-action-bar/cbp-action-bar.scss
```diff
@@ -0,0 +1,53 @@
...
+
+cbp-action-bar {
+  display: flex;
+  align-items: center;
+  justify-content: right;
+  height: var(--cbp-space-14x);
+  width: 100%;
+  padding: var(--cbp-space-3x);
+  background-color: var(--cbp-action-bar-color-background);
+  box-shadow: var(--cbp-shadow-level-3-up);
```
dgibson666: shadow only applies to floating, not inline.

packages/web-components/src/components/cbp-action-bar/cbp-action-bar.stories.tsx
```diff
@@ -0,0 +1,50 @@
+export default {
+    title: 'Components/Action Bar',
+    tags: ['autodocs'],
+    argTypes: {
+    
+      actionBarContent:{
+        name: 'Action Bar Content',
+        description: 'Buttons, links, or action items to populate the action bar',
+        control: 'string',
```
dgibson666: this should be text. control is broken now.

packages/web-components/src/components/cbp-action-bar/cbp-action-bar.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+          </cbp-typography>
+          ${actionBarContent}
+        </cbp-action-bar>
+      `;
+  };
+
+  export const ActionBar = Template.bind({});
+      ActionBar.args = {
+        actionBarInfo: `0 items selected.`,
+        actionBarContent: ` <cbp-button fill="ghost" context="dark-inverts" accessibility-text="Action 1">Action 1</cbp-button>
```
dgibson666: these buttons need to inherit the same context as the parent component.

dgibson666 (changes requested):
Lots of small changes. Please add requirements.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
bagrub: without this the display block was breaking layout of the action bar on the structured list footer. can look for separate solution but this was least verbose way i could think of ATM

bagrub: updating to use the slot=cbp-structured-list-footer resolved this better, removing this :not()

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
```diff
@@ -87,12 +87,16 @@ cbp-structured-list {
     font-style: italic;
   }
 
+  [slot=cbp-structured-list-footer]{
```
dgibson666: you don't know what exactly will be slotted here (should be the action bar), so you shouldn't be trying to set CSS on it from here. It should be up to the slotted content/component, so just remove this.

packages/web-components/src/components/cbp-action-bar/cbp-action-bar.scss
```diff
@@ -7,47 +7,44 @@
...
-  height: var(--cbp-space-14x);
+  min-height: var(--cbp-space-14x);
   width: 100%;
   padding: var(--cbp-space-3x);
+  color: var(--cbp-action-bar-color);
   background-color: var(--cbp-action-bar-color-background);
-  box-shadow: var(--cbp-shadow-level-3-up);
+
 
    & > *{
```
dgibson666: you removed part of this CSS, but I don't think you need this still and I don't see a gap added.

dgibson666 (changes requested):
still a few small things. The comment about the responsive spacing token was never addressed.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
```diff
@@ -67,8 +67,8 @@ cbp-structured-list {
   display: block; 
   background-color: var(--cbp-structured-list-color-background);
   color: var(--cbp-structured-list-color);
-  
-  [slot] {
+
+  & > [slot] {
```
dgibson666: If this is only targeting the header, target the specific slot name rather than using > for specificity. It's more clear to read and I wasn't clear at first glance why this wasn't targeting the footer as well.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
```diff
@@ -112,7 +108,7 @@ cbp-structured-list {
     *Collection 
   */
 
-  div[role=list]:only-child > *:first-child {
+  div[role=list] > *:first-child {
```
dgibson666: This is now adding the border regardless of the header's existence. It's the same color as the header, but it's adding 2px of border space and affecting the visual alignment (Jeremy will notice).

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.tsx
```diff
@@ -45,21 +45,22 @@ export class CbpStructuredList {
...
         <div role="list"
           aria-label={this.accessibilityText}
           aria-describedby={this.headerId}
         >
           <slot />
         </div>
- 
-        <slot name="cbp-structured-list-footer" />
-      </Host>
+        <div class='cbp-structured-list-footer'>
```
dgibson666: if you're not actually targeting the class name in your CSS, then you don't need it. Do you need the wrapping div at all still? I don't recall the reasoning after realizing the side effects now. Wasn't it just so that you're not applying some CSS to the action bar and doubling up on padding?

dgibson666 (changes requested):
Couple small comments in code, but there are bigger issues:

- I think your colors are reversed because the default story is dark (buttons are low contrast due to the wrong context)
- the context controls doesn't change anything.
- I think the default style should be light and when used in the structured list, it should set `context="dark-inverts"` in the story.
- Make sure your buttons are following the same context as the action bar.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
bagrub: update as per call

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.tsx
bagrub: updates as per call

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
bagrub: updates as per call

dgibson666 (approved):
looks good now!

PR #164, opened 2024-07-19
Feature/tabs darkmode

dgibson666:
* Added dark mode to tabs
* Added `context` prop to `cbp-tabs` and story
* Removed `pointerOnly` prop from button in favor of slotting a custom button (this was only used internally by the responsive tabs controls)

Everything should be good now except for dark mode of the responsive button controls, which requires button CSS to be refactored with a dark mode API (already in backlog).


bagrub (approved):
Looks good to me

PR #169, opened 2024-07-23
Update to footer

bagrub:


packages/web-components/src/components/cbp-footer/cbp-footer.scss
```diff
@@ -22,3 +26,9 @@ cbp-footer {
     padding-top: var(--cbp-space-5x);
   }
 }
+
+@media (min-width: $cbp-breakpoint-md) {
```
dgibson666: I ran a quick test and I don't think you need this. The flex context takes care of this on its own and the buttons look exactly the same when flexed with 100% width as with auto.

packages/web-components/src/components/cbp-footer/cbp-footer.scss
```diff
@@ -10,10 +10,14 @@ cbp-footer {
     --cbp-button-height: var(--cbp-space-11x);
     --cbp-button-border-radius: 0;
     background-color: var(--cbp-color-gray-cool-60);
+     & cbp-flex-item > cbp-button {
```
dgibson666: Too much specificity. I think you just need `cbp-button` (no &).

dgibson666 (changes requested):
Some small tweaks to your CSS and it's good to go.

packages/web-components/src/components/cbp-footer/cbp-footer.scss
```diff
@@ -13,6 +13,9 @@ cbp-footer {
     cbp-button {
       width: 100%; 
       --cbp-button-width: 100%; 
+      & > a{
```
dgibson666: Make this just a nested `*` selector. The button can render a button or anchor, but also accept slotted button/anchor/componets, which means the `>` is fragile.

dgibson666 (approved):
looks good now.

PR #172, opened 2024-07-29
Simple list

bagrub:


dgibson666 (changes requested):
I feel like this one requires a lot of work. Some of the main show-stoppers right now:

* The list item component was not supposed to be committed and doesn't appear to be used - please delete the files from your branch and commit the deletion so they are not in your PR (you can make a copy of them outside of the project if you want to save the work)
* tag should be a prop with values of ul, ol, dl. This eliminates the translation and shortens your code.
* Your HTML structure is not correct for your nested list, which is causing you to do more work like the indent value in your story. Nested lists should live inside of a list item. The only allowable content inside of a list element is "Zero or more li, script, and template elements."
* Your story is nesting same-type lists, but probably should allow specifying the nested list type as well for better testing. (Your CSS selectors reflect this and don't account for mixed nested lists).

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -0,0 +1,50 @@
...
+
+  li {
+    color: var(--cbp-list-item-color);
+    font-size: var(--cbp-font-size-body);
+    font-weight: var(--cbp-font-weight-bold);
+    display: list-item;
+  }
+
+  //Unordered nested lists
+  li + ul{
```
dgibson666: This is wrong. Nested lists need to be inside of list items. They cannot be siblings. Your nested selectors should look like `ul ol, ol ol {...}`

packages/web-components/src/components/cbp-list/cbp-list.stories.tsx
```diff
@@ -0,0 +1,104 @@
+export default {
+    title: 'Components/Simple List',
```
dgibson666: Can drop the "simple" - just call it "list".

packages/web-components/src/components/cbp-list/cbp-list.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+            },
+            {//example of ol sublist
+              content: `
+                <ol>
+                  <li>List Sub-Item 1</li>
+                  <li>List Sub-Item 2</li>
+                  <li>List Sub-Item 3</li>
+                  <li>List Sub-Item 4</li>
+                  <li>List Sub-Item 5</li>
+                  <ol>
```
dgibson666, line 82: Like in the CSS, this is wrong. Nested lists need to be inside of list items. They cannot be siblings.

packages/web-components/src/components/cbp-list/cbp-list.tsx
```diff
@@ -0,0 +1,45 @@
...
+  tag: 'cbp-list',
+  styleUrl: 'cbp-list.scss',
+})
+export class CbpList {
+  @Element() host: HTMLElement;
+  
+  private renderedTag: HTMLElement;
+
+  /** Specifies the semantic tag to be rendered. */
+  @Prop() tag!: 'ul' | 'ol' ;
```
dgibson666: Make this default to "ul"

packages/web-components/src/components/cbp-list/cbp-list.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  /** Supports adding inline styles as an object */
+  @Prop() sx: any = {};
+
+  /** Specifies the context of the component as it applies to the visual design and whether it inverts when light/dark mode is toggled. Default behavior is "light-inverts" and does not have to be specified. */
+  @Prop({ reflect: true }) context: "light-inverts" | "light-always" | "dark-inverts" | "dark-always";
+
```
dgibson666, line 24: This component also needs an `accessibilityText` prop that renders and `aria-label` on the list. See structured list reqs and add it the same guidance to these requirements.

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -0,0 +1,50 @@
...
+    padding-inline-start: 3ch;
```
dgibson666: Is this correct? I would think it would be specified using a design token (not a magic number).

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -0,0 +1,50 @@
...
+}
+
+cbp-list {
+  line-height: var(--cbp-line-height-sm);
+
+  li {
+    color: var(--cbp-list-item-color);
+    font-size: var(--cbp-font-size-body);
+    font-weight: var(--cbp-font-weight-bold);
+    display: list-item;
```
dgibson666: Is this necessary? It should be the default behavior already.

dgibson666 (changes requested):
See code comments.

packages/web-components/src/components/cbp-list/cbp-list.scss
bagrub: updated to css var

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -0,0 +1,58 @@
...
+      list-style-type: '- ';
+    }
+  }
+  //Ordered nested lists
+    ol ol{
+      padding-inline-start: 3ch;
+        li{ //Ordered list 1st sublist
+        list-style-type: upper-alpha; 
+        }
+        ol li{ //Ordered list 2nd sublist
```
dgibson666: This level is supposed to be "#)" formatting. I'm not sure if that's possible - need to look into it. Right now it's identical to the top level list, which is not the intention.

packages/web-components/src/components/cbp-list/cbp-list.specs.mdx
```diff
@@ -0,0 +1,31 @@
+import { Meta } from '@storybook/addon-docs';
+
+<Meta title="Components/List/Specifications" />
+
+# cbp-list
+
+## Purpose
+    the simple list renders unordered and ordered lists in accordance with the design system
```
dgibson666: Please remove all "simple list" instances

packages/web-components/src/components/cbp-list/cbp-list.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+          options: ['ul', 'ol'],
+        },
+        context : {
+        control: 'select',
+        options: [ "light-inverts", "light-always", "dark-inverts", "dark-always"]
+        },
+        sx: {
+        description: 'Supports adding inline styles as an object of key-value pairs comprised of CSS properties and values. Values should reference design tokens when possible.',
+        control: 'object',
+        },
```
dgibson666, line 21: need a control for accessibility text.

packages/web-components/src/components/cbp-list/cbp-list.tsx
```diff
@@ -0,0 +1,54 @@
...
+  @Element() host: HTMLElement;
+  
+  private renderedTag: HTMLElement;
+
+  /** Specifies the semantic tag to be rendered. */
+  @Prop() tag: 'ul' | 'ol' = 'ul';
+
+  /** 
+  * Specifies an accessible label for the list as an `aria-label`, similar to a table `caption`. 
+  * Since the structured list contains significant amount of data, it is advised to always specify a label describing the list.
```
dgibson666: remove the second line copied from structured list

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -0,0 +1,58 @@
...
+  // //Unordered nested lists
+  ul li{
+    list-style-type: disc;
+   
+    ul li{ //Unordered list 1st sublist
+      list-style-type: '- ';
+    }
+  }
+  //Ordered nested lists
+    ol ol{
```
dgibson666: I think you need to set the default marker for ordered lists because the first one under list item 5 has bullets instead of numbers when the top level list is `ul`.

dgibson666 (changes requested):
Some additional code comments. I think we need to discuss the 3ch padding because I think it's being misinterpreted from the design specs.

packages/web-components/src/components/cbp-list/cbp-list.scss
dgibson666: First demo here does this exactly: https://developer.mozilla.org/en-US/docs/Web/CSS/counter

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -0,0 +1,68 @@
...
+  li {
+    color: var(--cbp-list-item-color);
+    font-size: var(--cbp-font-size-body);
+    font-weight: var(--cbp-font-weight-bold);
+    padding-block: var(--cbp-list-item-padding);
+  }
+
+  
+  // //Unordered nested lists
+  ul > li{
```
dgibson666, line 36: Do you need the >? I would think the nested ones would automatically have higher specificity.

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -0,0 +1,68 @@
...
+    list-style-type: disc;
+   
+    ul li{ //Unordered list 1st sublist
+      list-style-type: '- ';
+    }
+  }
+    
+  //Ordered nested lists
+  ol > li{
```
dgibson666, line 45: Do you need the >? I would think the nested ones would automatically have higher specificity.

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -0,0 +1,68 @@
...
+      ol li{ //Ordered list 2nd sublist
+        list-style-type: decimal; 
+    
+        ol{
+          counter-reset: count;
+    
+          li{//Ordered list 3rd sublist
+            counter-increment: count;
+             
+            &::marker{
```
dgibson666, line 67: This one is too deep. It's currently on 4th-level, but should be on 3rd. 4th is small alpha.

packages/web-components/src/components/cbp-list/cbp-list.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+    });
+    return html.join('');
+  }
+  
+const Template = ({listItems, tag, context, accessibilityText }) => {
+    return ` 
+    <cbp-list
+      ${tag ? `tag=${tag}` : ''}
+      ${context && context != 'light-inverts' ? `context=${context}` : ''}
+      ${accessibilityText ? `accessibilityText=${accessibilityText}` : ''}
```
dgibson666: this isn't working because it needs to be dash-case in HTML/template literal format. Just the HTML part - not the  ternary check.

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -0,0 +1,68 @@
+/**
+ * @prop --cbp-list-item-color: var(--cbp-color-text-darkest);
+ * @prop --cbp-list-item-color-dark: var(--cbp-color-text-lightest);
+*/
+
+:root {
+  --cbp-list-item-color: var(--cbp-color-text-darkest);
+  --cbp-list-item-color-dark: var(--cbp-color-text-lightest);
+  --cbp-list-padding: var(--cbp-space-4x);
+  --cbp-list-item-padding: 2px;
```
dgibson666: this is --cbp-space-half-x

dgibson666 (changes requested):
Couple things not working still.

packages/web-components/src/components/cbp-list/cbp-list.scss
bagrub: direct child there is so that when they are nested in the other we start at the first marker in type. I was having some cross pollination between ul and ol for sub lists

packages/web-components/src/components/cbp-list/cbp-list.scss
bagrub: direct child there is so that when they are nested in the other we start at the first marker in type. I was having some cross pollination between ul and ol for sub lists

packages/web-components/src/components/cbp-list/cbp-list.scss
bagrub: ah, didn't see that as an option. updating to this var now

packages/web-components/src/components/cbp-list/cbp-list.stories.tsx
bagrub: removed ternary check here

dgibson666 (approved):
I think you got it!

PR #173, opened 2024-07-30
Expand dark

bagrub:

### DIFF
--- a/packages/web-components/src/components/cbp-expand/cbp-expand.scss
+++ b/packages/web-components/src/components/cbp-expand/cbp-expand.scss
@@ -1,6 +1,3 @@
-/**
- *  @Prop  --cbp-accordion-item-color: var(--cbp-color-text-darker);
- */
 :root {
   --cbp-accordion-item-control-font-size: var(--cbp-font-size-heading-sm);
   --cbp-accordion-item-control-font-weight: var(--cbp-font-weight-semibold);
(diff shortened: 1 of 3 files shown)


dgibson666 (changes requested):
This looks good overall, but the fact that there is no dark mode+context selector in the main component CSS throws up a big red flag. While it looks like it works in most cases, there are two where it won't work:

If the label is slotted, there's no control over that text. You're only testing the case where the label is plain text passed as a prop and gets rendered in a `cbp-typography` component with its own context.

When using light-always or dark-always in opposing light/dark modes, the content can get out of sync with the heading (the content is unreadable on the proper background).

So you need some color variable (+dark variant) pushed up to the `:root` selector and a dark mode toggle.

bagrub:
> This looks good overall, but the fact that there is no dark mode+context selector in the main component CSS throws up a big red flag. While it looks like it works in most cases, there are two where it won't work:
> 
> If the label is slotted, there's no control over that text. You're only testing the case where the label is plain text passed as a prop and gets rendered in a `cbp-typography` component with its own context.
> 
> When using light-always or dark-always in opposing light/dark modes, the content can get out of sync with the heading (the content is unreadable on the proper background).
> 
> So you need some color variable (+dark variant) pushed up to the `:root` selector and a dark mode toggle.

adding the css vars back in the :root{} and the dark selector back in and will focus on the label slot to confirm it is getting correct context state

dgibson666 (approved):
This looks good now. I would like any instances of "accordion" changed to "expand" in the CSS vars, but you didn't change those... up to you if you want to update them in this branch or we can update it later.

PR #174, opened 2024-08-06
tagUpdate

bagrub:

### DIFF
--- a/packages/web-components/src/components/cbp-tag/cbp-tag.scss
+++ b/packages/web-components/src/components/cbp-tag/cbp-tag.scss
@@ -26,7 +26,7 @@ cbp-tag {
   cursor: default;
   padding: 0 var(--cbp-space-3x);
   text-align: center;
-  font-weight: var(--cbp-font-weight-medium);
+  font-weight: var(--cbp-font-size-subhead);
   line-height: var(--cbp-space-7x);
   max-width: 100%;
   overflow: hidden;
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/cbp-tag/cbp-tag.scss
```diff
@@ -26,33 +26,33 @@ cbp-tag {
   cursor: default;
   padding: 0 var(--cbp-space-3x);
   text-align: center;
-  font-weight: var(--cbp-font-weight-medium);
+  font-weight: var(--cbp-font-size-subhead);
```
dgibson666: this is incorrect. This variable is a font size, which needs to be specified properly.

packages/web-components/src/components/cbp-tag/cbp-tag.stories.tsx
```diff
@@ -44,7 +44,7 @@ const Template = ({ label, color, withIcon, context, sx }) => {
       ${context && context != 'light-inverts' ? `context=${context}` : ''}
       ${sx ? `sx=${JSON.stringify(sx)}` : ''}
     >
-      ${Icon ? `<cbp-icon name="${Icon}" sx='{"margin-right":"var(--cbp-space-2x)"}'></cbp-icon>` : ''}
+      ${Icon ? `<cbp-icon name="${Icon}" size="0.75rem"sx='{"margin-right":"var(--cbp-space-1x)"}'></cbp-icon>` : ''}
```
dgibson666: Because the icon is the same size as the font size, you shouldn't need this. The default icon size is 1em, which is relative to the surrounding context. Fix the font size (above) and this fixes itself.

dgibson666 (changes requested):
Just a couple small comments.

packages/web-components/src/components/cbp-tag/cbp-tag.scss
dgibson666: The old font-weight is still needed though.

dgibson666 (approved):
Looks good now. Confirmed that the icon is the same size as the text.

PR #176, opened 2024-08-06
Link list

bagrub:


packages/web-components/src/components/cbp-icon/cbp-icon.tsx
```diff
@@ -95,6 +95,11 @@ export class CbpIcon {
       viewbox: "0 0 512 512",
       path: "M256 512A256 256 0 1 0 256 0a256 256 0 1 0 0 512zm0-384c13.3 0 24 10.7 24 24V264c0 13.3-10.7 24-24 24s-24-10.7-24-24V152c0-13.3 10.7-24 24-24zM224 352a32 32 0 1 1 64 0 32 32 0 1 1 -64 0z",    
     },
+    "external-link-alt": {
```
dgibson666, line 98: Please update the icon story (control dropdown) with any icons you've added (in this PR and recently). It needs to reflect this list.

packages/web-components/src/components/cbp-list/cbp-list.tsx
```diff
@@ -10,9 +10,15 @@ export class CbpList {
...
   private renderedTag: HTMLElement;
 
+  /** Specifies the variant of list (simple, link, special) */
+  @Prop() variant: 'simple' | 'link'; 
+  
   /** Specifies the semantic tag to be rendered. */
   @Prop() tag: 'ul' | 'ol' = 'ul';
 
+  /** Specifies the font size for the list */
+  @Prop() size: 'sm' | 'lg';
```
dgibson666: I would make this "normal" (it's not actually small) and "large" , defaulting to normal. It should also be reflected if you are using it as a CSS hook.

packages/web-components/src/components/cbp-list/cbp-list.tsx
```diff
@@ -10,9 +10,15 @@ export class CbpList {
   
   private renderedTag: HTMLElement;
 
+  /** Specifies the variant of list (simple, link, special) */
+  @Prop() variant: 'simple' | 'link';
```
dgibson666: needs to be reflected if you're using it as a CSS hook. You're still missing the "unstyled" list as well. I would remove "simple" and just list the actual variants of "link" and "unstyled".

packages/web-components/src/components/cbp-list/cbp-list.tsx
```diff
@@ -35,17 +41,18 @@ export class CbpList {
...
-        >
-          <slot />
-        </Tag>
-      </Host>
-    );
+      return (
+        <Host>
+          <Tag 
+            ref={(el) => this.renderedTag = el}
+            // aria-size={this.size}
```
dgibson666: remove

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -6,20 +6,53 @@
...
 [data-cbp-theme=dark] cbp-list:not([context=dark-inverts]):not([context=light-always]) {
   --cbp-list-item-color: var(--cbp-list-item-color-dark);
+  --cbp-link-list-icon-color: var(--cbp-link-list-icon-color-dark);
+  --cbp-link-list-description-color: var(--cbp-link-list-description-color-dark);
 }
 
 cbp-list {
   line-height: var(--cbp-line-height-xs);
 
+  &[size="sm"]:not([variant="link-internal"]):not([variant="link-external"]){
```
dgibson666: I'm not a fan of this tight coupling. I don't even think those are actual variants now. You should have a default normal list size and just change it for the large size.

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -6,20 +6,53 @@
...
+    font-size: var(--cbp-list-font-small);
+  
+    li{
+      line-height: var(--cbp-list-item-small-line-height);
+      padding-block: var(--cbp-list-item-small-padding);
+    }
+  }
+
+  &[size="lg"]:not([variant="link-internal"]):not([variant="link-external"]){
```
dgibson666: shouldn't need all these nots. The cascade should work fine since you're specifying the variant after the size.

In fact, look at how many times you're setting these? The CSS properties should be set once in the li selector and updated via variables in each of these selectors.

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -73,4 +106,24 @@ cbp-list {
...
+
+  &[variant="link"] li{
+    list-style-type: none;
+
+    .linkListIcon{
+      color: var(--cbp-link-list-icon-color);
+    }          
+  }
+
+  &[variant="link-external"]{
```
dgibson666: this is no longer a variant and shouldn't need to be.

dgibson666 (changes requested):
Lots of CSS cleanup and naming comments. 

Lastly, the link list doesn't behave like a list if you have wrapping text. We can discuss, but I would give the list a gutter via padding and then pull the icons to the left with a negative margin to make them faux list markers. This should still work with your list items with 2 icons (which I find wierd - I would think you'd mimic the list item marker with a single icon).

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -33,38 +40,29 @@
...
   
-    li{
-      line-height: var(--cbp-list-item-small-line-height);
-      padding-block: var(--cbp-list-item-small-padding);
+    li{  
+      color: var(--cbp-list-item-color);
+      font-weight: var(--cbp-font-weight-bold);
+      padding-block: var(--cbp-list-item-padding);
+      line-height: var(--cbp-list-item-line-height);
+      padding-block: var(--cbp-list-item-padding-block);
```
dgibson666, line 44: you've listed this twice with different values

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -108,22 +106,26 @@ cbp-list {
 }
 
   //Link List
+  &[variant="link"]{
+    padding-inline-start: var(--cbp-space-8x);
+    
+    li{
+      list-style-type: none;
+  
+      .linkListIcon,
```
dgibson666: I would simply target `cbp-icon` here. Forcing the end-user/developer to use a specific class name for slotted content isn't ideal. They may get the code from storybook, but otherwise there's no indication what classes they are supposed to use. This is a pretty simple use case and they shouldn't be sprinkling random icons within their lists otherwise.

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -108,22 +106,26 @@ cbp-list {
...
-    }          
-  }
-
-  &[variant="link-external"]{
-
-    .linkListDescription{
-      color: var(--cbp-link-list-description-color);
-      font-style: italic;
-      font-size: var(--cbp-font-size-body);
+      .linkListDescription{
```
dgibson666: This is bothering me for the same reason as above. Maybe link lists should be handled by a component instead? Too much structure is expected without providing that structure.

In any case, you shouldn't use mixed case in CSS classes. convert to dashed or BEM and we can discuss/revisit later.

packages/web-components/src/components/cbp-list/cbp-list.stories.tsx
```diff
@@ -5,7 +5,7 @@ export default {
         // variant: {
         //   control: 'select',
         //   description: 'Variant of List',
-        //   options: ['simple' , 'link-internal' , 'link-external'],
+        //   options: ['unstyled' , 'link-internal' , 'link-external'],
```
dgibson666: why do we need separate variants for link-internal and link-external? I don't think we do.

packages/web-components/src/components/cbp-list/cbp-list.tsx
```diff
@@ -10,14 +10,14 @@ export class CbpList {
...
-  @Prop() variant: 'simple' | 'link'; 
+  /** Specifies the variant of list (unstyled, link, special) */
+  @Prop() variant: 'unstyled' | 'link' = 'unstyled'; 
   
   /** Specifies the semantic tag to be rendered. */
   @Prop() tag: 'ul' | 'ol' = 'ul';
 
   /** Specifies the font size for the list */
-  @Prop() size: 'sm' | 'lg';
+  @Prop() size: 'normal' | 'lg' = 'normal';
```
dgibson666: sorry to nitpick, but spell out large for consistency.

packages/web-components/src/components/cbp-list/cbp-list.tsx
```diff
@@ -44,9 +44,10 @@ export class CbpList {
       return (
         <Host>
           <Tag 
-            ref={(el) => this.renderedTag = el}
-            // aria-size={this.size} 
+            ref={(el) => this.renderedTag = el} 
             aria-label={this.accessibilityText}
+            aria-size={this.size}
```
dgibson666: this isn't a thing and should be deleted.

packages/web-components/src/components/cbp-list/cbp-list.tsx
```diff
@@ -44,9 +44,10 @@ export class CbpList {
       return (
         <Host>
           <Tag 
-            ref={(el) => this.renderedTag = el}
-            // aria-size={this.size} 
+            ref={(el) => this.renderedTag = el} 
             aria-label={this.accessibilityText}
+            aria-size={this.size} 
+            aria-variant={this.variant}
```
dgibson666: this isn't a thing and should be deleted. This is what props are for. ARIA is not for making up fake attributes.

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -33,38 +40,29 @@
...
+      padding-block: var(--cbp-list-item-padding-block);
     }
-  }
 
-  &[size="lg"]:not([variant="link-internal"]):not([variant="link-external"]){
-    font-size: var(--cbp-list-font-large);
+  &[size="lg"]:not([variant="link"]){
+    --cbp-list-font-size: var(--cbp-font-size-heading-sm);
   
     li{
```
dgibson666: You can set all of these variables at the same level. They don't need to be on the LI selector to inherit down to the LI from the list's selector. This actually complicates overriding the values.

dgibson666 (changes requested):
I put in more code comments. Let's review tomorrow. I think there are a few things in my previous review missing still.

packages/web-components/src/components/cbp-list/cbp-list.tsx
```diff
@@ -17,7 +17,7 @@ export class CbpList {
   @Prop() tag: 'ul' | 'ol' = 'ul';
 
   /** Specifies the font size for the list */
-  @Prop() size: 'normal' | 'lg' = 'normal';
+  @Prop() size: 'normal' | 'large' = 'normal';
```
dgibson666: This needs to read `@Prop({reflect:true})` to use as a CSS hook. The only reason it works in Storybook is because it's treating your template literal as HTML + attributes (not props), so they are already in the DOM.

dgibson666 (changes requested):
It's looking better. Let go ahead and renamed the "link" variant to "icon" and break the stories out into 3 separate ones:

1. Internal Link list
2. External Link List
3. Icon list (with different icons as needed)

Also, please include the unstyled variant, which can be its own story since it doesn't make much sense to toggle unstyled/icon variants in a story (unless you want to make it work). I think this should wrap up all the list tickets. I think breaking them down too granular by variant was actually causing problems.

packages/web-components/src/components/cbp-list/cbp-list.tsx
```diff
@@ -10,14 +10,14 @@ export class CbpList {
   
   private renderedTag: HTMLElement;
 
-  /** Specifies the variant of list (unstyled, link, special) */
-  @Prop() variant: 'unstyled' | 'link' = 'unstyled'; 
+  /** Specifies the variant of list (unstyled, icon, special) */
+  @Prop() variant: 'unstyled' | 'icon' = 'unstyled';
```
dgibson666: This should not have a default. The default it bulleted/numbered. Unstyled removes the bullets.

packages/web-components/src/components/cbp-icon/cbp-icon.stories.tsx
```diff
@@ -18,6 +18,7 @@ export default {
         'circle-xmark',
         'ellipsis-vertical',
         'envelope',
+        'external-link-alt',
```
dgibson666, line 21: still missing globe

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -1,37 +1,64 @@
...
+   @prop --cbp-list-item-line-height: var(--cbp-line-height-sm);
+   @prop --cbp-list-item-padding-block: var(--cbp-space-half-x);
+
+   @prop --cbp-list-item-color: var(--cbp-color-text-darkest);
+   @prop --cbp-list-item-color-dark: var(--cbp-color-text-lightest);
+  
+   @prop --cbp-link-list-icon-color: var(--cbp-color-interactive-primary-dark);
+   @prop --cbp-link-list-icon-color-dark: var(--cbp-color-interactive-primary-dark);
+
+   @prop --cbp-link-list-description-color: var(--cbp-color-text-darkest);
```
dgibson666: These can be removed because they are the same color as the base list items/text.

packages/web-components/src/components/cbp-list/cbp-list.scss
```diff
@@ -73,4 +100,29 @@ cbp-list {
...
+    }
+
+    li{
+      list-style-type: none;
+  
+      cbp-icon{
+        margin-inline-start: calc(var(--cbp-space-4x) * -1);
+      }
+
+      .linkListDescription{
```
dgibson666: shouldn't need this. It should be handled by the typography tag.

packages/web-components/src/components/cbp-list/cbp-list.stories.tsx
```diff
@@ -31,13 +36,16 @@ function generateItems(items) {
     });
     return html.join('');
   }
+
+
   
-const Template = ({listItems, tag, context, accessibilityText }) => {
+const Template = ({listItems, tag, size, accessibilityText, context,}) => {
```
dgibson666: you're not passing in `sx` or applying it to the component. Please verify for each story.

packages/web-components/src/components/cbp-list/cbp-list.stories.tsx
```diff
@@ -100,4 +108,131 @@ export const List = Template.bind({});
...
+                ${parentVariant == 'link-external' ? '<br /><cbp-icon class="linkListDescriptionIcon" name="globe" /> </cbp-icon><span class="linkListDescription"> ' + content + ' description </span>' : ''}
+              </li>`;      
+    });
+    return html.join('');
+  } 
+}
+
+const InternalLinkListTemplate = ({linkListItems, size, accessibilityText, context,}) => {
+  return ` 
+  <cbp-typography tag="h3"> Internal Link List: </cbp-typography>
```
dgibson666: Drop the heading from the story. We just want the relevant list code.

packages/web-components/src/components/cbp-list/cbp-list.stories.tsx
```diff
@@ -100,4 +108,131 @@ export const List = Template.bind({});
...
+          },
+          {
+            content: "List Link Item 3",
+          },
+        ]
+};
+
+const ExternalLinkListTemplate = ({linkListItems, size, accessibilityText, context,}) => {
+  return ` 
+  <cbp-typography tag="h3"> External Link List: </cbp-typography>
```
dgibson666: Drop the heading from the story. We just want the relevant list code.

packages/web-components/src/components/cbp-list/cbp-list.stories.tsx
```diff
@@ -100,4 +108,131 @@ export const List = Template.bind({});
...
+                <cbp-icon name='${icon}' color='${color}'></cbp-icon>
+                ${content}
+              </li>`;      
+  });
+  return html.join('');
+}
+
+const IconListTemplate = ({linkListItems, accessibilityText, context,}) => {
+  return ` 
+  <cbp-typography tag="h3"> Icon Link List: </cbp-typography>
```
dgibson666: Drop the heading from the story. We just want the relevant list code.

dgibson666 (changes requested):
See code comments. It's mostly cleanup. Some additional notes:

* You're still using a span with a class for the link description. Replace this with cbp-typography to make the text italics.
* Still need an "unstyled" variant story and CSS to handle it.
* I'm not sure if the design specs address this, but the list padding works for normal lists - the bullets are just within the padding/box. But for large lists, the bullet is actually partially outside of the bounding box. This will affect alignment with other text around it.

dgibson666 (approved):
I think the gutter still has some issues, but I'm approving this with the intent to revisit the CSS and do some cleanup still when we have design feedback from Jeremy.

PR #182, opened 2024-08-19
update list doc

bagrub:


dgibson666 (approved):
Looks good. I may add to it later, but I don't feel like it fits in the asynchronous feedback loop here.

PR #184, opened 2024-08-19
Bugfix/form fields a11y 202408

dgibson666:
* Update form field CSS for optgroup and remove autofill bg color
* Update field wrapper to add overlays to aria-describedby
* Update the way form field IDs are handled (allow setting on the native element, using component prop as a fallback)
* Update stories for testing purposes
* update docs


bagrub (commented):
lgtm

PR #186, opened 2024-08-23
updates for badge, chip, panel, and card

bagrub:
#### Updates to badge, chip, panel, & card as detailed below via tickets

TCS-2422 chip component updates:

General
*font size is incorrect - should be $cbp-font-size-subhead (12px)
*icon size is incorrect - should be 12px x 12px (area occupied by icon)
*Dark mode - text color should be text-darkest
*dark mode - icon should be text-darkest in enabled state
*Dark mode - background circle - focus - circle size should be reduce 24 x24px see screenshot
*dark mode - background color- selected - should be selected- light (reads as #a1d3ff currently, not sure what color this is coming from)

TCS-2417 Card decision variant

Button Card:
All variants
*can we add in toggle to show icon in title? missing from storybook. see anatomy sheet for guidance //already completed before work started
*vertical spacing between title and body text should be 8px (spacing 2x) currently no spacing //already completed before work started
*there is a horizontal padding added of 12px (im ok with this but i never prescribed padding as the button widths were determined by card width - might want to reduce to 8px to minimize risk of wrapping text)

Button Arrangement is out of order: (serial positioning effect)
*on single button cards: just primary
*on 2 button cards: secondary (on left) primary (on right)
*on 3 button cards: tertiary (on left) secondary (in middle) primary (on right)

*with danger cards swap the primary button with danger (can be swapped by developers depending on context this is just for the default)
*On the Actions layout on storybook can we add options for single (with icon) double (with icon) triple (with icon) so developers can see button options that have icons attached?

Danger
[*]Add in danger button type (cant check otherwise)

TCS-2423- panel updates
*Panel header font weight should be $cbp-font-weight-medium (500) not regular
*Icon size should be 24x24px (area) 
*header area grows when icon is added? should stay at 72px tall.

Changes requested during design review checkin:
*badge dark mode text color (default) needs to be --cbp-badge-color-dark: var(--cbp-color-info-darker) 
*Card- margin update off title should specific to the decision variant
*card(banner): lightmode danger background. title color should be text-color-lighter. not darkest

### DIFF
--- a/packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
+++ b/packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
@@ -40,11 +40,11 @@ const renderActions = (layout, color, context, withIcon, { btn1, btn2, btn3 }) =
     return `
       <div slot="cbp-card-actions">
         <cbp-button tag="${btn2.tag}" ${btn2.tag == 'a' ? `href="#"` : ''} fill="solid" color="${btn2.color}" context="${context}" aria-describedby="card-heading-1">
-          ${withIcon ? `<cbp-icon name='arrow-right' sx='{"display":"inline", "padding-inline-end":"0.5rem"}'></cbp-icon>` : ''}
+          ${withIcon ? `<cbp-icon name='arrow-right' sx='{"display":"inline", "padding-inline-end":"var(--cbp-space-2x)"}'></cbp-icon>` : ''}
           ${btn2.label}
         </cbp-button>
         <cbp-button tag="${btn1.tag}" ${btn1.tag == 'a' ? `href="#"` : ''} fill="solid" color="${color == 'danger' ? 'danger' : btn1.color}" context="${context}" aria-describedby="card-heading-1">
-          ${withIcon ? `<cbp-icon name='check' sx='{"display":"inline", "padding-inline-end":"0.5rem"}'></cbp-icon>` : ''}
+          ${withIcon ? `<cbp-icon name='check' sx='{"display":"inline", "padding-inline-end":"var(--cbp-space-2x)"}'></cbp-icon>` : ''}
           ${btn1.label}
         </cbp-button>
       </div>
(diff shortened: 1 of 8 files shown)


packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
```diff
@@ -33,26 +33,46 @@ export default {
...
+const renderActions = (layout, color, context, withIcon, { btn1, btn2, btn3 }) => {
+  
+  
   if (layout === 'double') {
     return `
       <div slot="cbp-card-actions">
-        <cbp-button tag="${btn1.tag}" ${btn1.tag == 'a' ? `href="#"` : ''} fill="solid" color="${btn1.color}" context="${context}" aria-describedby="card-heading-1">${btn1.label}</cbp-button>
-        <cbp-button tag="${btn2.tag}" ${btn2.tag == 'a' ? `href="#"` : ''} fill="solid" color="${btn2.color}" context="${context}" aria-describedby="card-heading-1">${btn2.label}</cbp-button>
+        <cbp-button tag="${btn2.tag}" ${btn2.tag == 'a' ? `href="#"` : ''} fill="solid" color="${btn2.color}" context="${context}" aria-describedby="card-heading-1">
+          ${withIcon ? `<cbp-icon name='arrow-right' sx='{"display":"inline", "padding-inline-end":"0.5rem"}'></cbp-icon>` : ''}
```
dgibson666: Your padding inline should use a token/variable. No magic numbers. Same for each instance below.

packages/web-components/src/components/cbp-chip/cbp-chip.scss
```diff
@@ -128,8 +129,9 @@ cbp-chip {
 
       > .cbp-chip__label + cbp-icon {
         background-color: var(--cbp-chip-icon-color-background-focus);
-        border-color: var(--cbp-color-white);
-        outline: var(--cbp-border-size-md) solid var(--cbp-chip-icon-color-outline-focus);
+        // border-color: var(--cbp-color-white);
```
dgibson666: your changes here have broken the focus state in light mode. The icon should have a circular white outline. It looks like it's different in dark mode so you might have to refactor a bit to hoist them up as variables and just set the variables here.

packages/web-components/src/components/cbp-chip/cbp-chip.tsx
```diff
@@ -70,7 +70,7 @@ export class CbpChip {
           <span class="cbp-chip__label">
             <slot />
           </span>
-          <cbp-icon name="plus"></cbp-icon>
+          <cbp-icon name="plus" size='0.75rem'></cbp-icon>
```
dgibson666: I think you can use a token/css variable here.

packages/web-components/src/components/cbp-panel/cbp-panel.stories.tsx
```diff
@@ -62,7 +62,7 @@ const PanelTemplate = ({ role, headingLevel, header, headerId, content, ariaLabe
         variant="heading-lg"
         ${headerId ? `id="${headerId}"` : ''}
       >
-        ${showIcon ? '<cbp-icon name="star" sx=\'{"margin-right":"var(--cbp-space-4x)"}\'></cbp-icon>' : ''}${header}
+        ${showIcon ? '<cbp-icon name="star" size="1.5rem" sx=\'{"margin-right":"var(--cbp-space-4x)", "vertical-align":"text-top"}\'></cbp-icon>' : ''}${header}
```
dgibson666: use a token for size. And can you use backticks here so you don't have to escape the single quotes? (or it won't let you since you're already in a template literal?)

dgibson666 (changes requested):
see code comments

dgibson666 (approved):
looks good

PR #188, opened 2024-08-26
updates to Modal and icons

bagrub:
#### Dialog updates as requested in TCS-2433

Dialog:
*z-index of modal should be $cbp-z-index-level-top
*dark mode tertiary button: enabled state - text color should be text-lightest
    default button order should be as follows:
*double button - secondary on the left, primary on the right
*triple button - tertiary on left, secondary on center, primary on right

 Icons:
*add icons to map as requested by Josue on the teams chat(pen to square, addressbook)
*add icons to story for testing


dgibson666 (approved):
looks good

PR #192, opened 2024-09-03
create the Media structured list sample

bagrub:


dgibson666 (approved):
Approving, but note that this breaks on a small viewport due to the non-breaking URLs. This is something we can discuss as a team for desired behavior.

PR #194, opened 2024-09-04
Tab update

bagrub:
#### Updates to Tab:
*Light Mode danger tab - change text color to danger-dark
*Dark Mode focus state - change outline color to black instead of white
*add in control to show icon before text
*add in control to show badge after text

Updates discussed in meeting:
*Drawer close button needs to be 16px instead of 1em in drawer
*Dialog, z-index has a fixed value. needs to be z-index-top

### DIFF
--- a/packages/web-components/src/components/cbp-tabs/cbp-tabs.stories.tsx
+++ b/packages/web-components/src/components/cbp-tabs/cbp-tabs.stories.tsx
@@ -10,6 +10,12 @@ export default {
     accessibilityText: {
       control: 'text',
     },
+    withIcon: {
+      control: 'boolean',
+    },
+    withBadge: {
+      control: 'boolean',
+    },
     context : {
       control: 'select',
       options: [ "light-inverts", "light-always", "dark-inverts", "dark-always"]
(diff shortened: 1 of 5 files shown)


packages/web-components/src/components/cbp-tabs/cbp-tabs.stories.tsx
```diff
@@ -71,12 +79,14 @@ Tabs.args = {
...
       selected: false,
+      withIcon: true,
+      withBadge: true
     },
     {
       name: 'tab2',
       label: 'Tab 2',
       accessibilityText: '',
-      color: 'default',
+      color: 'danger',
```
dgibson666: Can you change this back to default for the story? It's exposed for testing, but copying the default story code should not have test cases to fix/edit.

dgibson666 (changes requested):
Looks good - one minor change.

packages/web-components/src/components/cbp-tabs/cbp-tabs.stories.tsx
bagrub: updated

dgibson666 (approved):

PR #197, opened 2024-09-09
cbp-notice

bagrub:

### DIFF
--- a/packages/web-components/src/components/cbp-notice/cbp-notice.scss
+++ b/packages/web-components/src/components/cbp-notice/cbp-notice.scss
@@ -1,0 +1,60 @@
+/**
+ * @prop --cbp-list-padding: var(--cbp-space-4x);
+*/
+
+:root {  
+  --cbp-notice-background: var(--cbp-color-info-lighter);
+  --cbp-notice-background-dark: var(--cbp-color-info-dark);
+  --cbp-notice-border-left-color: var(--cbp-color-info-base);
+  --cbp-notice-border-left-color-dark: var(--cbp-color-info-light);
+}
+
+
+[data-cbp-theme=light] cbp-notice[context*=dark]:not([context=light-always]),
+[data-cbp-theme=dark] cbp-notice:not([context=dark-inverts]):not([context=light-always]) {
+  --cbp-notice-background: var(--cbp-notice-background-dark);
+  --cbp-notice-border-left-color: var(--cbp-notice-border-left-color-dark);
+}
+
+cbp-notice {
+  display: block;
+
+
+  .notice-container{
+    background-color: var(--cbp-notice-background);
+    border-left: var(--cbp-space-2x) solid var(--cbp-notice-border-left-color);
+    border-radius: var(--cbp-border-radius-softer); //0px 0px var(--cbp-border-radius-softer) var(--cbp-border-radius-softer);
+
+    padding: var(--cbp-space-3x);
+
+    [slot=cbp-notice-title]{
+      display: inline-block;
+      margin-block-end: var(--cbp-space-2x); //todo: not working as expected but rendering in DOM
+    }
+  }
+
+  &[color="success"]{
+    --cbp-notice-background: var(--cbp-color-success-lighter);
+    --cbp-notice-background-dark: var(--cbp-color-success-dark);   
+    
+    --cbp-notice-border-left-color: var(--cbp-color-success-base);
+    --cbp-notice-border-left-color-dark: var(--cbp-color-success-light);
+  }
+
+  
+  &[color="warning"]{
+    --cbp-notice-background: var(--cbp-color-warning-lighter);
+    --cbp-notice-background-dark: var(--cbp-color-warning-darker);   
+    
+    --cbp-notice-border-left-color: var(--cbp-color-warning-base);
+    --cbp-notice-border-left-color-dark: var(--cbp-color-warning-base);
+  }
+
+  &[color="danger"]{
+    --cbp-notice-background: var(--cbp-color-danger-lighter);
+    --cbp-notice-background-dark: var(--cbp-color-danger-dark);   
+    
+    --cbp-notice-border-left-color: var(--cbp-color-danger-base);
+    --cbp-notice-border-left-color-dark: var(--cbp-color-danger-light);
+  }
+}
--- a/packages/web-components/src/components/cbp-notice/cbp-notice.scss
+++ b/packages/web-components/src/components/cbp-notice/cbp-notice.scss
@@ -23,13 +23,13 @@ cbp-notice {
   .notice-container{
     background-color: var(--cbp-notice-background);
     border-left: var(--cbp-space-2x) solid var(--cbp-notice-border-left-color);
-    border-radius: var(--cbp-border-radius-softer); //0px 0px var(--cbp-border-radius-softer) var(--cbp-border-radius-softer);
+    border-radius: var(--cbp-border-radius-softer);
 
     padding: var(--cbp-space-3x);
 
     [slot=cbp-notice-title]{
       display: inline-block;
-      margin-block-end: var(--cbp-space-2x); //todo: not working as expected but rendering in DOM
+      margin-block-end: var(--cbp-space-2x); 
     }
   }
 
--- a/packages/web-components/src/components/cbp-notice/cbp-notice.scss
+++ b/packages/web-components/src/components/cbp-notice/cbp-notice.scss
@@ -5,24 +5,24 @@
 :root {  
   --cbp-notice-color-text: var(--cbp-color-text-darkest);
   --cbp-notice-color-text-dark: var(--cbp-color-text-lightest);
-  --cbp-notice-background: var(--cbp-color-info-lighter);
-  --cbp-notice-background-dark: var(--cbp-color-info-dark);
-  --cbp-notice-border-color: var(--cbp-color-info-base);//todo: rename to match naming convention better
-  --cbp-notice-border-color-dark: var(--cbp-color-info-light);
+  --cbp-notice-color-background: var(--cbp-color-info-lighter);
+  --cbp-notice-color-background-dark: var(--cbp-color-info-dark);
+  --cbp-notice-color-border: var(--cbp-color-info-base);
+  --cbp-notice-color-border--dark: var(--cbp-color-info-light);
 }
 
 
 [data-cbp-theme=light] cbp-notice[context*=dark]:not([context=light-always]),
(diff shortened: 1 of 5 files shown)


packages/web-components/src/components/cbp-notice/cbp-notice.scss
```diff
@@ -0,0 +1,60 @@
+/**
+ * @prop --cbp-list-padding: var(--cbp-space-4x);
+*/
+
+:root {
```
dgibson666, line 5: Your naming isn't consistent with other components. Please do --cbp-notice-color -bg -border, etc. You don't need to use "left" the border color.

packages/web-components/src/components/cbp-notice/cbp-notice.scss
```diff
@@ -0,0 +1,60 @@
...
+[data-cbp-theme=dark] cbp-notice:not([context=dark-inverts]):not([context=light-always]) {
+  --cbp-notice-background: var(--cbp-notice-background-dark);
+  --cbp-notice-border-left-color: var(--cbp-notice-border-left-color-dark);
+}
+
+cbp-notice {
+  display: block;
+
+
+  .notice-container{
```
dgibson666: I don't think you need this wrapper. Your component is the wrapper.

packages/web-components/src/components/cbp-notice/cbp-notice.specs.mdx
```diff
@@ -0,0 +1,29 @@
+import { Meta } from '@storybook/addon-docs';
+
+<Meta title="Components/Notice/Specifications" />
+
+# cbp-notice
+
```
dgibson666, line 6: Review for punctuation consistency (add missing periods). I would grab the text for responsiveness and accessibility from another component for consistent wording. This sounds a little clumsy/not as succinct and clear as it could be.

packages/web-components/src/components/cbp-notice/cbp-notice.stories.tsx
```diff
@@ -0,0 +1,54 @@
+export default {
+    title: 'Components/Notice',
+    tags: ['autodocs'],
+    argTypes: {
+      label: {
```
dgibson666: The Label control is not being used. Hoist the title and content up here in argTypes and name them with (slotted) so it's clear they aren't mapping to props.

packages/web-components/src/components/cbp-notice/cbp-notice.scss
```diff
@@ -0,0 +1,60 @@
...
+:root {  
+  --cbp-notice-background: var(--cbp-color-info-lighter);
+  --cbp-notice-background-dark: var(--cbp-color-info-dark);
+  --cbp-notice-border-left-color: var(--cbp-color-info-base);
+  --cbp-notice-border-left-color-dark: var(--cbp-color-info-light);
+}
+
+
+[data-cbp-theme=light] cbp-notice[context*=dark]:not([context=light-always]),
+[data-cbp-theme=dark] cbp-notice:not([context=dark-inverts]):not([context=light-always]) {
```
dgibson666, line 16: You'll need variables for the text (heading and content if they are not the same), because the text isn't changing when you change the context alone (not the OS dark/light mode).

packages/web-components/src/components/cbp-notice/cbp-notice.tsx
```diff
@@ -0,0 +1,32 @@
...
+  /** Supports adding inline styles as an object */
+  @Prop() sx: any = {};
+
+  /** Specifies the context of the component as it applies to the visual design and whether it inverts when light/dark mode is toggled. Default behavior is "light-inverts" and does not have to be specified. */
+  @Prop({ reflect: true }) context: "light-inverts" | "light-always" | "dark-inverts" | "dark-always";
+    
+  render() {
+    return (
+      <Host>
+        <div class='notice-container'>
```
dgibson666: As in the CSS, I don't think you need this wrapper. The component is the wrapper.

packages/web-components/src/components/cbp-notice/cbp-notice.tsx
```diff
@@ -0,0 +1,32 @@
+import { Component, Host, Prop, h } from '@stencil/core';
+
+@Component({
+  tag: 'cbp-notice',
+  styleUrl: 'cbp-notice.scss',
+
+})
+export class CbpNotice {
+
+  /** Optionally specifies a card color (different from the default color) based on predefined design token values. */
```
dgibson666, line 13: I feel like this should default to "info". Can you check some other components with the color prop and see if they default?

packages/web-components/src/components/cbp-notice/cbp-notice.tsx
```diff
@@ -0,0 +1,32 @@
...
+  @Prop({ reflect: true }) color: 'info' | 'success' | 'warning' | 'danger';
+    
+  /** Supports adding inline styles as an object */
+  @Prop() sx: any = {};
+
+  /** Specifies the context of the component as it applies to the visual design and whether it inverts when light/dark mode is toggled. Default behavior is "light-inverts" and does not have to be specified. */
+  @Prop({ reflect: true }) context: "light-inverts" | "light-always" | "dark-inverts" | "dark-always";
+
```
dgibson666, line 21: You've got an `sx` prop, but it's not hooked up in the component, so it won't work.

dgibson666 (changes requested):
Decent first cut. See code comments.

packages/web-components/src/components/cbp-notice/cbp-notice.scss
bagrub: removed

packages/web-components/src/components/cbp-notice/cbp-notice.stories.tsx
bagrub: updated

packages/web-components/src/components/cbp-notice/cbp-notice.tsx
bagrub: removed

packages/web-components/src/components/cbp-notice/cbp-notice.tsx
bagrub: updated

packages/web-components/src/components/cbp-notice/cbp-notice.tsx
bagrub: updated for default as 'info'. Looks like we might have some inconsistency on default's for color prop, looks like card doesn't have defaults but button does. might want to make a tech debt ticket or something on that later

packages/web-components/src/components/cbp-notice/cbp-notice.specs.mdx
bagrub: update. pulled from cbp-container for the text, can update/refine as needed

packages/web-components/src/components/cbp-notice/cbp-notice.scss
bagrub: updated

packages/web-components/src/components/cbp-notice/cbp-notice.scss
bagrub: updated

packages/web-components/src/components/cbp-notice/cbp-notice.scss
```diff
@@ -3,58 +3,57 @@
 */
 
 :root {  
+  --cbp-notice-color-text: var(--cbp-color-text-darkest);
+  --cbp-notice-color-text-dark: var(--cbp-color-text-lightest);
   --cbp-notice-background: var(--cbp-color-info-lighter);
   --cbp-notice-background-dark: var(--cbp-color-info-dark);
-  --cbp-notice-border-left-color: var(--cbp-color-info-base);
-  --cbp-notice-border-left-color-dark: var(--cbp-color-info-light);
+  --cbp-notice-border-color: var(--cbp-color-info-base);//todo: rename to match naming convention better
```
dgibson666: they should be --cbp-notice-color-bg, --cbp-notice-color-border. Color is the main part after the component name. Sorry if I wasn't clear.

packages/web-components/src/components/cbp-notice/cbp-notice.tsx
```diff
@@ -1,30 +1,41 @@
...
+import { setCSSProps } from '../../utils/utils';
 
 @Component({
   tag: 'cbp-notice',
   styleUrl: 'cbp-notice.scss',
 
 })
 export class CbpNotice {
+  
+  private renderedTag: HTMLElement;
```
dgibson666: This isn't what you need and it's not set to anything. SX references this.host, which uses the @Element decorator.

dgibson666 (changes requested):
couple things still not right. see code comments.

packages/web-components/src/components/cbp-notice/cbp-notice.scss
bagrub: ah, that makes sense. Updated

packages/web-components/src/components/cbp-notice/cbp-notice.tsx
bagrub: updated

packages/web-components/src/components/cbp-notice/cbp-notice.scss
```diff
@@ -5,24 +5,24 @@
...
   --cbp-notice-color-text: var(--cbp-color-text-darkest);
   --cbp-notice-color-text-dark: var(--cbp-color-text-lightest);
-  --cbp-notice-background: var(--cbp-color-info-lighter);
-  --cbp-notice-background-dark: var(--cbp-color-info-dark);
-  --cbp-notice-border-color: var(--cbp-color-info-base);//todo: rename to match naming convention better
-  --cbp-notice-border-color-dark: var(--cbp-color-info-light);
+  --cbp-notice-color-background: var(--cbp-color-info-lighter);
+  --cbp-notice-color-background-dark: var(--cbp-color-info-dark);
+  --cbp-notice-color-border: var(--cbp-color-info-base);
+  --cbp-notice-color-border--dark: var(--cbp-color-info-light);
```
dgibson666: you have 2 dashes here. --dark

dgibson666 (changes requested):
1 typo, which is causing no border to show in dark mode for info.

packages/web-components/src/components/cbp-notice/cbp-notice.scss
bagrub: updated

dgibson666 (approved):

PR #198, opened 2024-09-11
updates for button and segmented button

bagrub:
#### Updates to button and segmented button

Button:
*Light - Primary - Outline - Hover state - Border color should be primary-dark (not darker)
*Dark - Secondary - Outline - hover state - border color should be secondary-lighter (not darker)

Segmented Button:
*Added story for Icon only segmented button
*Light Mode: Disabled state - colors are inverted. text should be disabled-dark, bg color should be disabled-light, border color should remain secondary-dark
*Dark Mode: enabled - text color should be secondary-lighter not text lightest //Done before i started

### DIFF
--- a/packages/web-components/src/components/cbp-segmented-button-group/cbp-segmented-button-group.scss
+++ b/packages/web-components/src/components/cbp-segmented-button-group/cbp-segmented-button-group.scss
@@ -54,5 +54,16 @@ cbp-segmented-button-group {
       --cbp-button-color-dark: var(--cbp-color-text-darkest);
       --cbp-button-color-bg-dark: var(--cbp-color-interactive-selected-light);
     }
+
+    
+    //Disabled
+    [type='button'][disabled]  {
+      --cbp-button-color: var(--cbp-color-interactive-disabled-dark);
+      --cbp-button-color-bg: var(--cbp-color-interactive-disabled-light);
+      --cbp-button-color-border: var(--cbp-color-interactive-secondary-dark);
+    }
+
+    //read-only
+    
   }
 }
--- a/packages/web-components/src/components/cbp-segmented-button-group/cbp-segmented-button-group.scss
+++ b/packages/web-components/src/components/cbp-segmented-button-group/cbp-segmented-button-group.scss
@@ -56,14 +56,13 @@ cbp-segmented-button-group {
     }
 
     
-    //Disabled
-    [type='button'][disabled]  {
+    // Disabled
+    [disabled]  {
       --cbp-button-color: var(--cbp-color-interactive-disabled-dark);
       --cbp-button-color-bg: var(--cbp-color-interactive-disabled-light);
-      --cbp-button-color-border: var(--cbp-color-interactive-secondary-dark);
-    }
 
-    //read-only
-    
+      --cbp-button-color-dark: var(--cbp-color-interactive-disabled-dark);
+      --cbp-button-color-bg-dark: var(--cbp-color-interactive-disabled-light);
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/cbp-segmented-button-group/cbp-segmented-button-group.scss
```diff
@@ -54,5 +54,16 @@ cbp-segmented-button-group {
       --cbp-button-color-dark: var(--cbp-color-text-darkest);
       --cbp-button-color-bg-dark: var(--cbp-color-interactive-selected-light);
     }
+
+    
+    //Disabled
+    [type='button'][disabled]  {
```
dgibson666: I'm not sure about this selector. I think it should be `cbp-button[disabled]` - the way you've written it could be applying it to both the component and the button - and the component application may not work outside of storybook as written.

Also, if you are overriding colors via the CSS API, you should need to specify the dark mode colors too. (Verified this doesn't look right in dark mode)

packages/web-components/src/components/cbp-segmented-button-group/cbp-segmented-button-group.stories.tsx
```diff
@@ -65,3 +65,40 @@ SegmentedButtonGroup.args = {
...
+          ${generateButtons(buttons)}
+        </cbp-segmented-button-group>
+      `;
+};
+
+export const SegmentedButtonGroupIcons = IconTemplate.bind({});
+SegmentedButtonGroupIcons.args = {
+  buttons: [
+    {
+      label: '<cbp-icon name="user"></cbp-icon>',
```
dgibson666, line 86: I think you need to add `variant="square"` to each of these buttons. They are not rendering as square icon buttons still because the default button padding is more on the horizontal axis than vertical.

dgibson666 (changes requested):
Couple small issues. See code comments.

packages/web-components/src/components/cbp-segmented-button-group/cbp-segmented-button-group.stories.tsx
bagrub: updated

packages/web-components/src/components/cbp-segmented-button-group/cbp-segmented-button-group.scss
bagrub: Added the dark mode variables to match. removed the [type=button] from selector as it is unnecessary, also scss tree has the cbp-button already included so selector will render as (cbp-segmented-button-group cbp-button[color][fill] [disabled])

packages/web-components/src/components/cbp-segmented-button-group/cbp-segmented-button-group.stories.tsx
```diff
@@ -23,8 +23,8 @@ export default {
 };
 
 function generateButtons(buttons) {
-  const html = buttons.map(({ label, value, pressed, disabled }) => {
-    return `<cbp-button type="button" value="${value}" ${pressed == true ? 'pressed' : ''} ${disabled == true ? 'disabled' : ''}>${label}</cbp-button>`;
+  const html = buttons.map(({ label, value, pressed, disabled, variant }) => {
+    return `<cbp-button type="button" value="${value}" ${pressed == true ? 'pressed' : ''} ${disabled == true ? 'disabled' : ''} variant="${variant ? variant : 'default'}">${label}</cbp-button>`;
```
dgibson666: I don't think we want `variant="default"` output in code to copy-paste. Just output an empty string if it's not specified (like in the first story).

dgibson666 (changes requested):
1 story-related change. everything else looks good.

packages/web-components/src/components/cbp-segmented-button-group/cbp-segmented-button-group.stories.tsx
bagrub: updated

dgibson666 (approved):

PR #199, opened 2024-09-11
First cut of cbp-checkbox component

dgibson666:
* First cut of cbp-checkbox component
* Basic story for single checkbox. Checklists are a different beast that I'll be addressing separately in a later push (requires updates to `cbp-form-field`)

The majority of this effort was getting the CSS working for all those states and checked combinations. Base component functionality works very similar to `cbp-form-field` but may be revisited for code sharing with radio buttons.


bagrub (commented):
lgtm

PR #201, opened 2024-09-13
Add indeterminate state to checkbox, other tweaks.

dgibson666:
* Add indeterminate state to checkbox and story.
* Fix `checked` property so that it stays in sync with the checked state and is part of the emitted event.


bagrub (approved):
lgtm

PR #203, opened 2024-09-16
Card flag variant

bagrub:


dgibson666 (approved):
looks good. I would have you remove the commented code, but I know you'll be in here for another couple iterations and can do it then. Also, you don't need to pass the image tag as an argument. It can be coded directly into the story instead.

PR #205, opened 2024-09-23
Example of card layout in internal template

dgibson666:
Created a new template story with a responsive grid of cards.


bagrub (approved):
lgtm

PR #206, opened 2024-09-23
Interactive cards

bagrub:

### DIFF
--- a/packages/web-components/src/components/cbp-card/cbp-card.scss
+++ b/packages/web-components/src/components/cbp-card/cbp-card.scss
@@ -28,7 +28,7 @@
 
 :root {
   --cbp-card-color-background: var(--cbp-color-white);
-  --cbp-card-color-background-dark: var(--cbp-color-gray-cool-60);
+  --cbp-card-color-background-dark: var(--cbp-color-gray-cool-70);
   --cbp-card-color-border: var(--cbp-color-gray-cool-10);
   --cbp-card-color-border-dark: var(--cbp-color-gray-cool-50);
   --cbp-card-color-title: var(--cbp-color-text-darker);
@@ -309,14 +309,11 @@ cbp-card {
   &[interactive="clickable"],
   // &[select],
   &[interactive="multi"]{
-
+    --cbp-checkbox-min-height: var(--cbp-space-6x); //adjust spacing for checkbox in title for multi variant
+    
     a{
       color: var(--cbp-card-color-text); 
       flex-grow: 1;
-    }
-
-    cbp-checkbox > label{ //todo: look to define via the * @prop --cbp-checkbox-min-height: var(--cbp-space-11x);
-    min-height: var(--cbp-space-2x);
     }
 
     &:hover{
@@ -325,7 +322,7 @@ cbp-card {
       --cbp-card-color-title: var(--cbp-color-text-darker);
       --cbp-card-color-text: var(--cbp-color-text-darker);
   
-      --cbp-card-color-background-dark: var(--cbp-color-gray-cool-50); //todo: Confirm value change is good with Visual design
+      --cbp-card-color-background-dark: var(--cbp-color-gray-cool-60);
       --cbp-card-color-border-dark: var(--cbp-color-interactive-secondary-lighter);
       --cbp-card-color-title-dark: var(--cbp-color-text-lighter);
       --cbp-card-color-text-dark: var(--cbp-color-text-lighter);
--- a/packages/web-components/src/components/cbp-card/cbp-card.scss
+++ b/packages/web-components/src/components/cbp-card/cbp-card.scss
@@ -312,6 +312,11 @@ cbp-card {
 
     a{
       color: var(--cbp-card-color-text); 
+      flex-grow: 1;
+    }
+
+    cbp-checkbox > label{ //todo: look to define via the * @prop --cbp-checkbox-min-height: var(--cbp-space-11x);
+    min-height: var(--cbp-space-2x);
     }
 
     &:hover{
@@ -320,13 +325,14 @@ cbp-card {
       --cbp-card-color-title: var(--cbp-color-text-darker);
       --cbp-card-color-text: var(--cbp-color-text-darker);
   
-      --cbp-card-color-background-dark: var(--cbp-color-gray-cool-60);
+      --cbp-card-color-background-dark: var(--cbp-color-gray-cool-50); //todo: Confirm value change is good with Visual design
       --cbp-card-color-border-dark: var(--cbp-color-interactive-secondary-lighter);
       --cbp-card-color-title-dark: var(--cbp-color-text-lighter);
       --cbp-card-color-text-dark: var(--cbp-color-text-lighter);
     }      
 
-    &:focus{
+    &:focus,
+    &:focus-within{
       --cbp-card-color-background: var(--cbp-color-interactive-focus-dark);
       --cbp-card-color-border: var(--cbp-color-interactive-secondary-lighter);
       --cbp-card-color-title: var(--cbp-color-text-lighter);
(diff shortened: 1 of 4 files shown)


packages/web-components/src/components/cbp-card/cbp-card.scss
```diff
@@ -40,16 +46,23 @@
...
 
+  //flag
   --cbp-flag-card-min-width: 7rem;
   --cbp-flag-card-color: var(--cbp-color-text-lighter);
   --cbp-flag-card-color-dark: var(--cbp-color-text-darker);
   --cbp-flag-card-color-background: var(--cbp-color-gray-cool-60);
   --cbp-flag-card-color-background-dark: var(--cbp-color-gray-cool-30);
+
+  //interactive
+  --cbp-interactive-card-outline-focus: var(--cbp-color-white);
```
dgibson666: these are colors - you're missing the word "color" in them. Should be --cbp-interactive-card-color-outline-focus.

packages/web-components/src/components/cbp-card/cbp-card.scss
```diff
@@ -40,16 +46,23 @@
...
+  --cbp-interactive-card-outline-focus-dark: var(--cbp-color-black);
+  --cbp-interactive-card-danger-outline-focus: var(--cbp-color-white);
+  --cbp-interactive-card-danger-outline-focus-dark: var(--cbp-color-black);
 }
 
 
 [data-cbp-theme=light] cbp-card[context*=dark]:not([context=light-always]),
-[data-cbp-theme=dark] cbp-card:not([context=dark-inverts]):not([context=light-always]) {
+[data-cbp-theme=dark][theme="dark"] cbp-card:not([context=dark-inverts]):not([context=light-always]) {
```
dgibson666, line 65: why is this added? It's redundant, technically.

packages/web-components/src/components/cbp-card/cbp-card.scss
```diff
@@ -85,6 +101,17 @@ cbp-card {
...
       margin-bottom: 0;
     }
+
+    .cbp-card-select{
+      label{
+        min-height: var(--cbp-space-4x);
+      }
+    
+      input{
+        display: flex;
```
dgibson666: this shouldn't be needed. display flex is for a parent and an input cannot have children. I toggled it off and didn't see any effect.

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
```diff
@@ -315,4 +336,37 @@ FlagCard.argTypes = {
...
+};
+
+InteractiveCard.argTypes = {
+  interactive: {
+    name: 'Interactive',
+    description: 'Set the interactivity of the card',
+    control: 'select',
+    options: ['clickable', 'select', 'multi'],
+  },
+  href:{
```
dgibson666, line 371: This should be defaulted to something. If you make it a clickable card, the anchor is not interactive without an href, so it's not focusable (and you can't see focus states) and you also can't see that the body text changes color as a live link (should be fixed probably)

packages/web-components/src/components/cbp-card/cbp-card.tsx
```diff
@@ -43,17 +49,52 @@ export class CbpCard {
...
-        <div class="cbp-card-body">
-          {this.variant !== 'banner' && <slot name="cbp-card-title" />}
-          <slot />
-        </div>
+        {this.interactive !== 'clickable' && <div>
+          {this.variant === 'flag' && <div class='cbp-card-flag'><slot name='cbp-card-flag'/></div>}
+            <div class="cbp-card-body">
+              {(this.variant !== 'banner' && (this.interactive !='multi' && this.interactive !='select')) && <slot name="cbp-card-title" />}
+              {this.interactive === 'multi' && 
+                <cbp-checkbox
```
dgibson666: I think this needs a lot more thought/discussion. I don't think the component should render the checkbox because it removes control of the form element from the developer/app. Your story doesn't have an input name, which means it wouldn't be submitted in the formData(). The reason the native inputs are slotted is so frameworks can access them directly and apply other attributes/bindings, but being rendered by the component will prevent that.

packages/web-components/src/components/cbp-card/cbp-card.tsx
```diff
@@ -19,8 +19,14 @@ export class CbpCard {
...
-  /** When present, the card will have a slot avaliable to display an img/icon flag for the card  */
-  // @Prop({ reflect: true }) flag: boolean = false;
+  /** Specifies the interactivity of the card */
+  @Prop({ reflect: true}) interactive: 'clickable' | 'select' | 'multi';
+
+  /** Specifies the href for the 'clickable' setting of interactive cards */
+  @Prop() href: string
+
+  /** Marks the rendered card in a disabled state when specified. */
+  @Prop({ reflect: true}) disabled: boolean;
```
dgibson666, line 29: you're using disabled for styling, but the anchor is still active (if an hred was provided). You need to tie that functionality together. Take a look at the cbp-link and see how it handles disabled links as well.

dgibson666 (changes requested):
Some significant updates needed and probably some discussion. See comments in code.

packages/web-components/src/components/cbp-card/cbp-card.scss
bagrub: added this so the specificy was higher than the danger hover dark mode. was having issues with the the variables loading the light mode colors in dark for that instance, Can look for other avenues if we don't want to resolve it this way

packages/web-components/src/components/cbp-card/cbp-card.scss
bagrub: updated variable name to outline-color

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
bagrub: set the default to point to github instance of storybook (can change this if you want, just first address i could think of that would be 'internal' & added style for the color of the a tag to map to the --cbp-card-color-text driving color on component

packages/web-components/src/components/cbp-card/cbp-card.scss
```diff
@@ -398,6 +388,12 @@ cbp-card {
     }
 
     &[disabled]{
+
+      a{
+        pointer-events: none;
+        cursor: not-allowed;
```
dgibson666: if you made the anchor aria-disabled, the cursor would not be needed here.

packages/web-components/src/components/cbp-card/cbp-card.tsx
```diff
@@ -47,56 +47,31 @@ export class CbpCard {
   }
 
   render() {
-    return (
-      <Host>
-        {this.interactive === 'clickable' && 
+    if(this.interactive === 'clickable'){
+      return(
+        <Host>
           <a class='cbp-card-clickable' href={this.href}>
```
dgibson666: you need to add aria-disabled="true" (string) here, if it's disabled. This will apply the cursor automatically.

dgibson666 (changes requested):
Couple in-code comments, but here are my overall testing results:

Clickable:

Need to apply aria-disabled="true" (string) to the anchor when disabled=true. This will handle the cursor.

If disabled, you need to pass "disabled" down to the checkbox in the story as well. I'm on the fence if this should be done by the component...

The anchor needs flex-grow: 1. If your viewport is large enough that the text does not reach the end of the card, you can see the hover/focus outline does not stretch the full width of the card.

I don't see the background color (and text) changing on hover/focus (only active/pressed).

Checkbox variation:

It's strange how if you click anywhere in the card, the whole card gets an active style (but it's not focusable), but it doesn't toggle the checkbox unless you've clicked the label or checkbox. The user feedback isn't lining up and I think clicking anywhere in the card should toggle the checkbox. I'm ok with deferring this, however, as it might be significant and you'll have to revisit with the radio version anyway.

Since the entire card is not focusable, I don't see the focus state at all and hover is not changing the background. (related to above)

If I view a regular card and change it to the checkbox variation, the spacing changes significantly due to the checkbox. The padding or min-height should be exposed in a CSS variable, as it gets overridden for the multi-select dropdown too.

dgibson666 (changes requested):
The only issue I'm seeing now is that for the checkbox variation the heading (label) is not inverting on click/focus like the text and is super-low contrast/unreadable.

dgibson666 (approved):
looks good. Note: still needs iterating on the check/radio cards to make them entirely clickable.

PR #207, opened 2024-09-25
add the slot for card actions back for non interactive variants

bagrub:

### DIFF
--- a/packages/web-components/src/components/cbp-card/cbp-card.tsx
+++ b/packages/web-components/src/components/cbp-card/cbp-card.tsx
@@ -70,6 +70,7 @@ export class CbpCard {
                 {(this.variant !== 'banner') && <slot name="cbp-card-title" />}
                 <slot />
             </div>
+            <slot name='cbp-card-actions' />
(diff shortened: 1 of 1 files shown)


packages/web-components/src/components/cbp-card/cbp-card.tsx
```diff
@@ -70,6 +70,7 @@ export class CbpCard {
                 {(this.variant !== 'banner') && <slot name="cbp-card-title" />}
                 <slot />
             </div>
+            <slot name='cbp-card-actions' />
```
dgibson666: I don't see this filtering out interactive cards. I think {!this.interactive &&  <slot name='cbp-card-actions' />} is needed

dgibson666 (changes requested):

packages/web-components/src/components/cbp-card/cbp-card.tsx
bagrub: updated

dgibson666 (approved):

PR #216, opened 2024-10-17
Structured list select

bagrub:

### DIFF
--- a/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.tsx
+++ b/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.tsx
@@ -28,7 +28,6 @@ export class CbpStructuredListItem {
     });
   }
 
-  //TODO: check just all of this, coded w/ a head cold
   @Listen('stateChanged')
   stateChangedHandler(){
     console.log('state toggle for structured list item selected');
--- a/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.tsx
+++ b/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.tsx
@@ -29,9 +29,8 @@ export class CbpStructuredListItem {
   }
 
   @Listen('stateChanged')
-  stateChangedHandler(){
-    console.log('state toggle for structured list item selected');
-    this.selected = !this.selected;
+  stateChangedHandler({detail: {checked}}){
+    this.selected = checked;
   }
 
   render() {
--- a/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.tsx
+++ b/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.tsx
@@ -1,4 +1,4 @@
-import { Component, Prop, Element, Host, h } from '@stencil/core';
+import { Component, Prop, Element, Host, h, Listen } from '@stencil/core';
 import { setCSSProps } from '../../utils/utils';
 
 @Component({
@@ -26,6 +26,13 @@ export class CbpStructuredListItem {
     setCSSProps(this.host, {
       ...this.sx,
     });
+  }
+
+  //TODO: check just all of this, coded w/ a head cold
+  @Listen('stateChanged')
+  stateChangedHandler(){
+    console.log('state toggle for structured list item selected');
+    this.selected = !this.selected;
   }
 
   render() {
--- a/packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
+++ b/packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
@@ -80,7 +80,7 @@ cbp-structured-list {
     font-weight: var(--cbp-font-weight-bold);
 
     // Nesting 2 flex contexts will not take up 100% of the width
-    *:only-child {
+    *:only-child:not(cbp-checkbox):not(input[type="checkbox"]) { //todo: confirm these :nots are needed
       flex-basis: 100%;
     }
   }
@@ -108,5 +108,24 @@ cbp-structured-list {
 
   div[role=list]:only-child > *:first-child {
     border-top: var(--cbp-border-size-md) solid var(--cbp-structured-list-border-color);
+  }
+
+  /** Selectable */
+  cbp-structured-list-item cbp-checkbox{
+    display: inline-flex;
+  }
+
+  cbp-structured-list-item[selected]{
+    --cbp-structured-list-color-background: var(--cbp-color-interactive-selected-light);
+    --cbp-structured-list-border-color: var(--cbp-color-gray-cool-30);
+    --cbp-structured-list-color-background-dark: var(--cbp-color-interactive-selected-dark);
+    --cbp-structured-list-border-color-dark: var(--cbp-color-gray-cool-50);
+
+    &:hover{
+    --cbp-structured-list-color-background: var(--cbp-color-gray-cool-10);
+    --cbp-structured-list-border-color: var(--cbp-color-gray-cool-30);
+    --cbp-structured-list-color-background-dark: var(--cbp-color-gray-cool-90);
+    --cbp-structured-list-border-color-dark: var(--cbp-color-gray-cool-50);
+    }
   }
 }
--- a/packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
+++ b/packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
@@ -80,7 +80,7 @@ cbp-structured-list {
     font-weight: var(--cbp-font-weight-bold);
 
     // Nesting 2 flex contexts will not take up 100% of the width
-    *:only-child:not(cbp-checkbox):not(input[type="checkbox"]) { //todo: confirm these :nots are needed
+     *:only-child:not(cbp-checkbox):not(input[type="checkbox"]) {
       flex-basis: 100%;
     }
   }
@@ -113,9 +113,21 @@ cbp-structured-list {
   /** Selectable */
   cbp-structured-list-item cbp-checkbox{
     display: inline-flex;
+
+    input[type=checkbox]{
+      margin-inline-end: var(--cbp-space-4x);
+    }
+  }
+
+  &[selectable] cbp-structured-list-item{
+    display: flex;
+    flex-direction: row; 
+    align-items: center;
+
   }
 
   cbp-structured-list-item[selected]{
+  
     --cbp-structured-list-color-background: var(--cbp-color-interactive-selected-light);
     --cbp-structured-list-border-color: var(--cbp-color-gray-cool-30);
     --cbp-structured-list-color-background-dark: var(--cbp-color-interactive-selected-dark);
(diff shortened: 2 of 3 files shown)


packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.tsx
```diff
@@ -28,6 +28,12 @@ export class CbpStructuredListItem {
     });
   }
 
+  @Listen('stateChanged')
+  stateChangedHandler(){
+    console.log('state toggle for structured list item selected');
+    this.selected = !this.selected;
```
dgibson666: You should be keying the selected value off of the checked value emitted in that event. Let me know if you want to talk through this. I think pagination is a good example where I'm using destructuring of custom events to get at a specific key.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -365,4 +365,64 @@ StructuredListMedia.args = {
...
+          selectable
+          header-id="list-header"
+          ${context && context != 'light-inverts' ? `context=${context}` : ''}   
+          ${sx ? `sx=${JSON.stringify(sx)}` : ''}
+        >
+        ${showHeader ? `<div slot="cbp-structured-list-header" id="${headerId}"><cbp-checkbox value='1'> <input type='checkbox' name='checkbox' value='1' /></cbp-checkbox> <span>XX results - Y filters Applied Updated: XX/XX/XXXX XX:XX:XX EST</span></div>` : ''}
+    
+        ${generateItems(listItems)}
+
+
```
dgibson666: Selectable lists are the usecase that require the footer most and it's not implemented in your story.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
```diff
@@ -109,4 +109,35 @@ cbp-structured-list {
   div[role=list]:only-child > *:first-child {
     border-top: var(--cbp-border-size-md) solid var(--cbp-structured-list-border-color);
   }
+
+  /** Selectable */
+  cbp-structured-list-item cbp-checkbox{
+    display: inline-flex;
+
+    input[type=checkbox]{
```
dgibson666: This doesn't feel right - you shouldn't be overriding CSS on these elements in this component. Leverage flex and use a gap, ideally. Even if you had to add a margin, it could be done on the component rather than the nested input checkbox.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -365,4 +365,64 @@ StructuredListMedia.args = {
...
+const StructuredListSelectableTemplate = ({ listItems, striped, showHeader, headerId, context, sx }) => {
+  return ` 
+       <cbp-structured-list
+          ${striped ? `striped` : ''}
+          selectable
+          header-id="list-header"
+          ${context && context != 'light-inverts' ? `context=${context}` : ''}   
+          ${sx ? `sx=${JSON.stringify(sx)}` : ''}
+        >
+        ${showHeader ? `<div slot="cbp-structured-list-header" id="${headerId}"><cbp-checkbox value='1'> <input type='checkbox' name='checkbox' value='1' /></cbp-checkbox> <span>XX results - Y filters Applied Updated: XX/XX/XXXX XX:XX:XX EST</span></div>` : ''}
```
dgibson666: your checkbox doesn't have a label. It also doesn't need a name for the select all - you wouldn't want it submitted in a post. I don't think it needs a value either technically or what it should be.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -365,4 +365,64 @@ StructuredListMedia.args = {
...
+    description: 'Configure various aspects of the list items within the structured list.',
+    control: 'object',
+  },
+  // selectable: true
+}
+StructuredListSelectable.args = {
+  showHeader: false,
+  listItems: [
+    {
+      content: "<cbp-checkbox value='1'> <input type='checkbox' name='checkbox' value='1' /></cbp-checkbox> <span>Structured List Selectable Item 1</span>",
```
dgibson666: You need labels for each of the checkboxes. Each of the values should be different (make them sequential?).

And finally because you're passing the checkbox as content, you can't pass in the context from the parent. If I change the context to light-always or dark-always, the checkboxes remain in their current state and don't match the structured list context. I think this probably requires modifying your story and adding a new function to generate selectable list items.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
```diff
@@ -80,7 +80,7 @@ cbp-structured-list {
     font-weight: var(--cbp-font-weight-bold);
 
     // Nesting 2 flex contexts will not take up 100% of the width
-    *:only-child {
+     *:only-child:not(cbp-checkbox):not(input[type="checkbox"]) {
```
dgibson666: The checkbox should never be the only child. I'm not really sure why this is needed.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
```diff
@@ -109,4 +109,35 @@ cbp-structured-list {
   div[role=list]:only-child > *:first-child {
     border-top: var(--cbp-border-size-md) solid var(--cbp-structured-list-border-color);
   }
+
+  /** Selectable */
+  cbp-structured-list-item cbp-checkbox{
+    display: inline-flex;
```
dgibson666: This shouldn't be needed. If I uncheck it in the browser tools, I see no change.

dgibson666 (commented):
See code comments.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
bagrub: adding the label to the check box seems to have made this change unnecessary, removing the not()'s and reverting to original selector. 

originally the global selector was applying it to the checkbox, & input of the checkbox so the checkbox was stretching to ~40% of the available width

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: wrote new function generateSelectableItems to address this. is applying context of parent to checkbox as well as adding the 'hidden' labels for the checkbox

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: removed the name & value here, added 'hidden' label in same fashion as the generateSelectableItems is applying them to the list items checkbox

packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.tsx
bagrub: update to use map the selected prop to the checked value of provided by the event. Used the   @Listen('buttonClick') listener from the pagination as an example for the deconstructing of a custom event

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
bagrub: removed this style block and applied gap to the &[selectable] cbp-structured-list-item{} selector below this

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -49,7 +49,28 @@ function generateItems(items) {
...
+      value= ${index}
+      ${context && context != 'light-inverts' ? `context=${context}` : ''}   
+    >
+      <label>
+        <input
+          type="checkbox"
+          name="checkbox"
+          value=${index}
+        />
+        <span style='display: none'>Checkbox ${index}</span>
```
dgibson666: just slap a `hidden` attribute on here instead. That's what it's for.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -49,7 +49,28 @@ function generateItems(items) {
...
-
+function generateSelectableItems(items, context){
+
+  const html = items.map(({ content, color, selected }, index) => {
+    return `<cbp-structured-list-item ${color != 'default' ? `color="${color}"` : ''} ${selected ? `selected` : ''} >
+    <cbp-checkbox 
+      value= ${index}
+      ${context && context != 'light-inverts' ? `context=${context}` : ''}   
+    >
+      <label>
```
dgibson666: the checkbox component renders the `label` tag, so you shouldn't add it here. You just needed to add the text.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -379,47 +400,53 @@ const StructuredListSelectableTemplate = ({ listItems, striped, showHeader, head
           ${context && context != 'light-inverts' ? `context=${context}` : ''}   
           ${sx ? `sx=${JSON.stringify(sx)}` : ''}
         >
-        ${showHeader ? `<div slot="cbp-structured-list-header" id="${headerId}"><cbp-checkbox value='1'> <input type='checkbox' name='checkbox' value='1' /></cbp-checkbox> <span>XX results - Y filters Applied Updated: XX/XX/XXXX XX:XX:XX EST</span></div>` : ''}
+        ${showHeader ? `<div slot="cbp-structured-list-header" id="${headerId}"><cbp-checkbox><label><input type="checkbox" /><span style='display: none'>check all</span></label></cbp-checkbox> <span>XX results - Y filters Applied Updated: XX/XX/XXXX XX:XX:XX EST</span></div>` : ''}
```
dgibson666: the checkbox component renders the `label` tag, so you shouldn't add it here. You just needed to add the text.

dgibson666 (commented):
Few small things - see code comments.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: as discussed in IM, keeping the display: none due to bug with hidden attribute

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: removed label tag here

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: removed label tag here

dgibson666 (approved):
looks good

PR #217, opened 2024-10-24
Toast component

bagrub:

### DIFF
--- a/packages/web-components/src/components/cbp-toast/cbp-toast.scss
+++ b/packages/web-components/src/components/cbp-toast/cbp-toast.scss
@@ -1,0 +1,149 @@
+/**
+ * @Prop   --cbp-toast-background: var(--cbp-color-info-dark);
+ * @Prop   --cbp-toast-background-dark: var(--cbp-color-info-lighter);
+ * @Prop   --cbp-toast-color: var(--cbp-color-text-lighter);
+ * @Prop   --cbp-toast-color-dark: var(--cbp-color-text-darker);
+ * @Prop   --cbp-toast-color-content: var(--cbp-color-text-lighter);
+ * @Prop   --cbp-toast-color-content-dark: var(--cbp-color-text-darkest);
+ * @Prop   --cbp-toast-sidebar: var(--cbp-color-info-darker);
+ * @Prop   --cbp-toast-sidebardark: var(--cbp-color-info-light);
+ * @Prop   --cbp-toast-sidebar-icon: var(--cbp-color-text-light);
+ * @Prop   --cbp-toast-sidebar-icon-dark: var(--cbp-color-text-darker);
+ */
+ :root{
+ 
+  --cbp-toast-background: var(--cbp-color-info-dark);
+  --cbp-toast-background-dark: var(--cbp-color-info-lighter);
+  --cbp-toast-color: var(--cbp-color-text-lighter);
+  --cbp-toast-color-dark: var(--cbp-color-text-darker);
+  --cbp-toast-color-content: var(--cbp-color-text-lighter);
+  --cbp-toast-color-content-dark: var(--cbp-color-text-darkest);
+  --cbp-toast-sidebar: var(--cbp-color-info-darker);
+  --cbp-toast-sidebardark: var(--cbp-color-info-light);
+  --cbp-toast-sidebar-icon: var(--cbp-color-text-light);
+  --cbp-toast-sidebar-icon-dark: var(--cbp-color-text-darker);
+
+}
+
+[data-cbp-theme=light] cbp-toast[context*=dark]:not([context=light-always]),
+[data-cbp-theme=dark] cbp-toast:not([context=dark-inverts]):not([context=light-always]) {
+
+  --cbp-toast-background: var(--cbp-toast-background-dark);
+  --cbp-toast-color: var(--cbp-toast-color-dark);
+  --cbp-toast-color-content: var(--cbp-toast-color-content-dark);
+  --cbp-toast-sidebar: var(--cbp-toast-sidebar-dark);
+  --cbp-toast-sidebar-icon: var(--cbp-toast-sidebar-icon-dark);
+
+}
+
+
+@keyframes show {
+  0%{
+    opacity: 1;
+    transform: translateX(100%);
+  }
+
+}
+
+@keyframes dismiss {
+  99%{
+    transform: translateX(200%);
+  }
+  100%{
+    visibility: hidden;
+  }
+}
+
+cbp-toast {
+  display: flex;
+  margin-left: auto;
+  margin-bottom: var(--cbp-space-6x);
+  width: max-content;
+  max-width: 50%;
+  background-color: var(--cbp-toast-background);
+  color: var(--cbp-toast-color);
+  box-shadow: 0px 0px 18px 2px rgba(0, 0, 0, 0.3);
+  border-radius: 5px;
+
+  animation: show 600ms 100ms ease-in-out forwards;
+  
+  &:not([open]){
+    animation: dismiss 1s ease-in-out forwards;
+  }
+
+  .cbp-toast-sidebar{
+    display: flex;
+    padding: var(--cbp-space-4x);
+    color: var(--cbp-toast-sidebar-icon);
+    background-color: var(--cbp-toast-sidebar);
+    border-top-left-radius: 5px;
+    border-bottom-left-radius: 5px;
+  }
+
+  .cbp-toast-title{
+    font-weight: var(--cbp-font-weight-medium);
+    font-size: var(--cbp-font-size-heading-sm);
+    line-height: var(--cbp-line-height-sm);
+  }
+
+  .cbp-toast-container{
+    width: max-content;
+    padding:  var(--cbp-space-3x);
+  }
+
+  .cbp-toast-content{
+    font-weight: var(--cbp-font-weight-regular);
+    font-size: var(--cbp-font-size-body);
+    line-height: var(--cbp-line-height-xs);
+    color: var(--cbp-toast-color-content);
+  }
+
+  .cbp-toast-button-bar{
+    display: flex;
+    justify-content: center;
+
+    & > div{
+      display: flex;
+      gap: var(--cbp-space-4x);
+    }
+
+    cbp-button[fill=ghost][color=secondary]{ //match button color to text color for toast
+      --cbp-button-color: var(--cbp-toast-color);
+    }
+  }
+
+  &[color='info']{
+    --cbp-toast-background: var(--cbp-color-info-dark);
+    --cbp-toast-background-dark: var(--cbp-color-info-lighter);
+    --cbp-toast-sidebar: var(--cbp-color-info-darker);
+    --cbp-toast-sidebar-dark: var(--cbp-color-info-light); 
+  }
+
+  &[color='success']{
+    --cbp-toast-background: var(--cbp-color-success-base);
+    --cbp-toast-background-dark: var(--cbp-color-success-lighter);
+    --cbp-toast-sidebar: var(--cbp-color-success-darker);
+    --cbp-toast-sidebar-dark: var(--cbp-color-success-light);
+  }
+
+  &[color='warning']{
+    --cbp-toast-background: var(--cbp-color-warning-base);
+    --cbp-toast-background-dark: var(--cbp-color-warning-base);
+    --cbp-toast-sidebar: var(--cbp-color-warning-darker);
+    --cbp-toast-sidebar-dark: var(--cbp-color-warning-darker);
+    --cbp-toast-color: var(--cbp-color-text-darker);
+    --cbp-toast-color-dark: var(--cbp-color-text-darker);
+    --cbp-toast-color-content: var(--cbp-color-text-darker);
+    --cbp-toast-color-content-dark: var(--cbp-color-text-darker);
+    --cbp-toast-sidebar-icon-dark: var(--cbp-color-text-light);
+  }
+  
+  &[color='danger']{
+
+    --cbp-toast-background: var(--cbp-color-danger-dark);
+    --cbp-toast-background-dark: var(--cbp-color-danger-lighter);
+    --cbp-toast-sidebar: var(--cbp-color-danger-darker);
+    --cbp-toast-sidebar-dark: var(--cbp-color-danger-light);
+  }
+
+}
--- a/packages/web-components/src/components/cbp-toast/cbp-toast.scss
+++ b/packages/web-components/src/components/cbp-toast/cbp-toast.scss
@@ -90,7 +90,6 @@ cbp-toast {
   }
 
   .cbp-toast-container{
-    width: max-content;
     padding:  var(--cbp-space-3x);
   }
 
--- a/packages/web-components/src/components/cbp-toast/cbp-toast.scss
+++ b/packages/web-components/src/components/cbp-toast/cbp-toast.scss
@@ -75,8 +75,8 @@ cbp-toast {
     align-items: center;
     padding: var(--cbp-space-4x);
     background-color: var(--cbp-toast-color-bg-sidebar);
-    border-top-left-radius: 5px;
-    border-bottom-left-radius: 5px;
+    border-top-left-radius: var(--cbp-border-radius-softer);;
+    border-bottom-left-radius: var(--cbp-border-radius-softer);;
 
     cbp-icon{
           color: var(--cbp-toast-color-icon-sidebar);
--- a/packages/web-components/src/components/cbp-toast/cbp-toast.scss
+++ b/packages/web-components/src/components/cbp-toast/cbp-toast.scss
@@ -1,38 +1,38 @@
 /**
- * @Prop   --cbp-toast-background: var(--cbp-color-info-dark);
- * @Prop   --cbp-toast-background-dark: var(--cbp-color-info-lighter);
+ * @Prop   --cbp-toast-color-bg: var(--cbp-color-info-dark);
+ * @Prop   --cbp-toast-color-bg-dark: var(--cbp-color-info-lighter);
  * @Prop   --cbp-toast-color: var(--cbp-color-text-lighter);
  * @Prop   --cbp-toast-color-dark: var(--cbp-color-text-darker);
  * @Prop   --cbp-toast-color-content: var(--cbp-color-text-lighter);
  * @Prop   --cbp-toast-color-content-dark: var(--cbp-color-text-darkest);
- * @Prop   --cbp-toast-sidebar: var(--cbp-color-info-darker);
- * @Prop   --cbp-toast-sidebardark: var(--cbp-color-info-light);
- * @Prop   --cbp-toast-sidebar-icon: var(--cbp-color-text-light);
- * @Prop   --cbp-toast-sidebar-icon-dark: var(--cbp-color-text-darker);
+ * @Prop   --cbp-toast-color-bg-sidebar: var(--cbp-color-info-darker);
+ * @Prop   --cbp-toast-color-bg-sidebar-dark: var(--cbp-color-info-light);
+ * @Prop   --cbp-toast-color-icon-sidebar: var(--cbp-color-text-light);
+ * @Prop   --cbp-toast-color-icon-sidebar-dark: var(--cbp-color-text-darker);
  */
  :root{
  
-  --cbp-toast-background: var(--cbp-color-info-dark);
-  --cbp-toast-background-dark: var(--cbp-color-info-lighter);
+  --cbp-toast-color-bg: var(--cbp-color-info-dark);
+  --cbp-toast-color-bg-dark: var(--cbp-color-info-lighter);
   --cbp-toast-color: var(--cbp-color-text-lighter);
   --cbp-toast-color-dark: var(--cbp-color-text-darker);
   --cbp-toast-color-content: var(--cbp-color-text-lighter);
   --cbp-toast-color-content-dark: var(--cbp-color-text-darkest);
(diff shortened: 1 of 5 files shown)


packages/web-components/src/components/cbp-toast/cbp-toast.scss
```diff
@@ -0,0 +1,149 @@
...
+ * @Prop   --cbp-toast-color-content: var(--cbp-color-text-lighter);
+ * @Prop   --cbp-toast-color-content-dark: var(--cbp-color-text-darkest);
+ * @Prop   --cbp-toast-sidebar: var(--cbp-color-info-darker);
+ * @Prop   --cbp-toast-sidebardark: var(--cbp-color-info-light);
+ * @Prop   --cbp-toast-sidebar-icon: var(--cbp-color-text-light);
+ * @Prop   --cbp-toast-sidebar-icon-dark: var(--cbp-color-text-darker);
+ */
+ :root{
+ 
+  --cbp-toast-background: var(--cbp-color-info-dark);
```
dgibson666: these need "color" in them to be consistent. And I have been abbreviating background as "bg".

packages/web-components/src/components/cbp-toast/cbp-toast.scss
```diff
@@ -0,0 +1,149 @@
...
+  --cbp-toast-background-dark: var(--cbp-color-info-lighter);
+  --cbp-toast-color: var(--cbp-color-text-lighter);
+  --cbp-toast-color-dark: var(--cbp-color-text-darker);
+  --cbp-toast-color-content: var(--cbp-color-text-lighter);
+  --cbp-toast-color-content-dark: var(--cbp-color-text-darkest);
+  --cbp-toast-sidebar: var(--cbp-color-info-darker);
```
dgibson666: all of these are colors and need color in the name.

packages/web-components/src/components/cbp-toast/cbp-toast.scss
```diff
@@ -0,0 +1,149 @@
...
+    transform: translateX(200%);
+  }
+  100%{
+    visibility: hidden;
+  }
+}
+
+cbp-toast {
+  display: flex;
+  margin-left: auto;
```
dgibson666: Use logical properties. We've been using them a lot, so let's try to be consistent and stick with them.

packages/web-components/src/components/cbp-toast/cbp-toast.scss
```diff
@@ -0,0 +1,149 @@
...
+  margin-bottom: var(--cbp-space-6x);
+  width: max-content;
+  max-width: 50%;
```
dgibson666: This will not scale well on mobile. You'd want it to be allowed to take most of the width on a small screen. Maybe think about using min() here or on the width itself. See https://www.smashingmagazine.com/2024/10/css-min-all-the-things/

packages/web-components/src/components/cbp-toast/cbp-toast.scss
```diff
@@ -0,0 +1,149 @@
...
+  background-color: var(--cbp-toast-background);
+  color: var(--cbp-toast-color);
+  box-shadow: 0px 0px 18px 2px rgba(0, 0, 0, 0.3);
+  border-radius: 5px;
+
+  animation: show 600ms 100ms ease-in-out forwards;
+  
+  &:not([open]){
```
dgibson666, line 69: Is this working? Usually when you update a prop and cause a re-render, animations break or are halted.

packages/web-components/src/components/cbp-toast/cbp-toast.scss
```diff
@@ -0,0 +1,149 @@
...
+    line-height: var(--cbp-line-height-sm);
+  }
+
+  .cbp-toast-container{
+    width: max-content;
+    padding:  var(--cbp-space-3x);
+  }
+
+  .cbp-toast-content{
+    font-weight: var(--cbp-font-weight-regular);
```
dgibson666, line 97: if this is all slotted content, you shouldn't need to explicitly define values that are the default text values.

packages/web-components/src/components/cbp-toast/cbp-toast.specs.mdx
```diff
@@ -0,0 +1,41 @@
+import { Meta } from '@storybook/addon-docs';
+
+<Meta title="Components/Toast/Specifications" />
+
+# cbp-toast
+
+## Purpose
+
+* Toast is intended to temporarialy display messages to users such as error or warning messages about an form input, or confrimation of a file upload. It is the intent of the designer that for things
```
dgibson666, line 9: We need to refine this. You should never use toast for form validation. Confirmation misspelled.

packages/web-components/src/components/cbp-toast/cbp-toast.specs.mdx
```diff
@@ -0,0 +1,41 @@
...
+like error messages to also be populated elsewhere in app so that they don't get 'lost'. 
+
+## Functional Requirements
+
```
dgibson666, line 13: We don't want to explicitly list props (more maintenance tying it to the implementation). Just defined things it does.

packages/web-components/src/components/cbp-toast/cbp-toast.specs.mdx
```diff
@@ -0,0 +1,41 @@
...
+    *Slots:
+        *title: Slot for the title section of the Toast, intended for text or typography tags
+        *content: Slot for the content of the toast, used for description of error or progress bar for example. 
+        *buttons: Slot intended for 1 or 2 buttons, UX design intended for these to be a dismiss (of the toast) and/or a button to link to log/more detailed section explaining the toast
+
+## Technical Specifications
+
+### User Interactions
+
+* Toast is intended to temporarialy draw attention to items/actions for the user as such they are meant to have an internal timer to display & have the displayed content in a log elsewhere for user to
```
dgibson666, line 29: This isn't a user interaction. User interaction would be dismissing a toast or acting on other slotted links/buttons.

packages/web-components/src/components/cbp-toast/cbp-toast.specs.mdx
```diff
@@ -0,0 +1,41 @@
...
+    review after the toast is gone. 
+
+### Responsiveness
+
+* Toast's width is set to the max-content of the toast, with a max-width on the toast of 50% of the viewportS 
+
+### Accessibility
+
+* Toast being on a timer is an Accessibility concern, which is why it is advised to have a log of toasts perserved elsewhere in the app as mentioned above
+* Slotted content is Accessibile via keyboard, primarily use being the slotted buttons
```
dgibson666, line 39: I meant to mention this previously, but toast should use `aria-live`. They are typically out of the flow of the page and would take forever to keyboard navigate to. You don't want to send focus and interrupt what the user is doing, but screen readers should read the toast automoatically.

packages/web-components/src/components/cbp-toast/cbp-toast.stories.tsx
```diff
@@ -0,0 +1,107 @@
...
+    </cbp-toast>
+          `;
+  };
+
+  export const MultipleToast = MultiTemplate.bind({});
+
+  MultipleToast.args = {
+    icon: 'user',
+    content: 'Notification Description - A rule you are following just fired.',
+    buttons: '<cbp-button type="button" fill="ghost" color="secondary"> Default </cbp-button> <cbp-button type="button" fill="ghost" color="secondary"> Default 2</cbp-button>'
```
dgibson666: I would just slot the buttons. This isn't providing much value allowing them to edit this block of HTML and you have an issue with context that may need to be passed down to them. (they are not visible in dark mode).

packages/web-components/src/components/cbp-toast/cbp-toast.tsx
```diff
@@ -0,0 +1,63 @@
...
+  render() {
+
+    if(this.open && this.timer){
+      setTimeout(() => { this.open = false }, this.timer * 1000)
+    }
+    
+    return (
+      <Host>
+        <div class='cbp-toast-sidebar'>
+          <cbp-icon
```
dgibson666: This locks them in to our icons. I would just provide a named slot so they can slot an icon component or include an external icon if needed.

packages/web-components/src/components/cbp-toast/cbp-toast.tsx
```diff
@@ -0,0 +1,63 @@
...
+export class CbpToast {
+
+  /** specifies the icon loaded into the sidebar */
+  @Prop() icon: string = 'user';
+
+  /** specifies the color for the toast */
+  @Prop({ reflect: true }) color: 'info' | 'danger' | 'success' | 'warning' = 'info';
+
+  /** specifies time in seconds for the toast to be displayed */
+  @Prop() timer: 3 | 5 | 10;
```
dgibson666: I'd call this duration - it's more accurate.

packages/web-components/src/components/cbp-toast/cbp-toast.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  /** When set, specifies that the toast is open */
+  @Prop({ reflect: true }) open: boolean = true;
```
dgibson666: booleans default to false and you should never default them to true because in plain HTML, the absence of a boolean attribute is false. So how do you set it to false? Maybe it makes sense in this case if the component is handling the dismissal but we should talk it through.

packages/web-components/src/components/cbp-toast/cbp-toast.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  /** Specifies the context of the component as it applies to the visual design and whether it inverts when light/dark mode is toggled. Default behavior is "light-inverts" and does not have to be specified. */
+  @Prop({ reflect: true }) context: "light-inverts" | "light-always" | "dark-inverts" | "dark-always";
+  
+  /** Supports adding inline styles as an object */
+  @Prop() sx: any = {};
```
dgibson666, line 25: SX is not implemented in the component.

packages/web-components/src/components/cbp-toast/cbp-toast.scss
```diff
@@ -0,0 +1,149 @@
...
+    display: flex;
+    justify-content: center;
+
+    & > div{
+      display: flex;
+      gap: var(--cbp-space-4x);
+    }
+
+    cbp-button[fill=ghost][color=secondary]{ //match button color to text color for toast
+      --cbp-button-color: var(--cbp-toast-color);
```
dgibson666, line 113: Maybe this needs the dark color as well? Is that why dark mode is breaking?

packages/web-components/src/components/cbp-toast/cbp-toast.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return (
+      <Host>
+        <div class='cbp-toast-sidebar'>
+          <cbp-icon
+            size='2rem'
+            name={this.icon}
+            />
+        </div>
+        <div class='cbp-toast-container'>
+          <span class='cbp-toast-title'>
```
dgibson666: This span should be a div I think, for consistency.

dgibson666 (commented):
Lots of code comments. 

As noted, the buttons aren't readable in dark mode.

Also, there's no way to manually dismiss them. Is one of the buttons supposed to always be "dismiss"? Is it possible to have no buttons, since they are slotted? Probably should discuss some of this stuff.

packages/web-components/src/components/cbp-toast/cbp-toast.scss
bagrub: updated to --cbp-toast-color-bg

packages/web-components/src/components/cbp-toast/cbp-toast.scss
bagrub: updated to --cbp-toast-color-bg-sidebar & --cbp-toast-color-icon-sidebar

packages/web-components/src/components/cbp-toast/cbp-toast.scss
bagrub: updated

packages/web-components/src/components/cbp-toast/cbp-toast.tsx
bagrub: updated name from 'timer' to 'duration'

packages/web-components/src/components/cbp-toast/cbp-toast.tsx
bagrub: added

packages/web-components/src/components/cbp-toast/cbp-toast.scss
bagrub: added variable mapping for dark mode

packages/web-components/src/components/cbp-toast/cbp-toast.tsx
bagrub: updated to <div>

packages/web-components/src/components/cbp-toast/cbp-toast.tsx
bagrub: updated to remove the default setting on the toast.tsx and add the storybook control to start the sample with this populated as true, as discussed on call

packages/web-components/src/components/cbp-toast/cbp-toast.scss
bagrub: tested & working correctly from what i am seeing

packages/web-components/src/components/cbp-toast/cbp-toast.scss
bagrub: removed max-width and used a min() on width to account for mobile as discussed

packages/web-components/src/components/cbp-toast/cbp-toast.scss
```diff
@@ -56,11 +56,10 @@
...
 cbp-toast {
   display: flex;
-  margin-left: auto;
-  margin-bottom: var(--cbp-space-6x);
-  width: max-content;
-  max-width: 50%;
-  background-color: var(--cbp-toast-background);
+  margin-inline-start: auto;
+  margin-block-end: var(--cbp-space-6x);
+  width: min(100%, 500px);
```
dgibson666: I'd still use relative units instead of px. Convert to rems in a round number like 30rem *(480px) or similar.

packages/web-components/src/components/cbp-toast/cbp-toast.scss
```diff
@@ -74,8 +73,8 @@ cbp-toast {
   .cbp-toast-sidebar{
     display: flex;
     padding: var(--cbp-space-4x);
-    color: var(--cbp-toast-sidebar-icon);
-    background-color: var(--cbp-toast-sidebar);
+    color: var(--cbp-toast-color-icon-sidebar);
+    background-color: var(--cbp-toast-color-bg-sidebar);
     border-top-left-radius: 5px;
```
dgibson666: we should have tokens for border radius.

packages/web-components/src/components/cbp-toast/cbp-toast.stories.tsx
```diff
@@ -22,11 +22,12 @@ export default {
...
   };
   
-  const Template = ({ icon, content, buttons, timer, color, context, sx }) => {
+  const Template = ({ open, icon, content, buttons, duration, color, context, sx }) => {
     return ` 
           <cbp-toast
+            ${open ? `open=${open}` : ''}
             color=${color}
-            timer=${timer}
+            duration=${duration}
```
dgibson666: you need to use a ternary for these anytime they are able to be null. Right now anyone copying the code will get these attributes, which is not idea:
  color="undefined"
  duration="undefined"

packages/web-components/src/components/cbp-toast/cbp-toast.stories.tsx
```diff
@@ -22,11 +22,12 @@ export default {
     },
   };
   
-  const Template = ({ icon, content, buttons, timer, color, context, sx }) => {
+  const Template = ({ open, icon, content, buttons, duration, color, context, sx }) => {
     return ` 
           <cbp-toast
+            ${open ? `open=${open}` : ''}
```
dgibson666: yout don't need the  "=${open}" here. Just the "open" attribute being present.

packages/web-components/src/components/cbp-toast/cbp-toast.stories.tsx
```diff
@@ -41,16 +42,18 @@ export default {
   export const Toast = Template.bind({});
   
   Toast.args = {
+    open: true,
     icon: 'user',
     content: 'Notification Description - A rule you are following just fired.',
```
dgibson666, line 49: You should have a control for the title as well. This is hardcoded into your stories.

packages/web-components/src/components/cbp-toast/cbp-toast.tsx
```diff
@@ -47,9 +59,9 @@ export class CbpToast {
             />
```
dgibson666: This is still being rendered instead of slotted? Let's discuss.

dgibson666 (commented):
Some more code comments, but looking good.

packages/web-components/src/components/cbp-toast/cbp-toast.scss
bagrub: > updated to 30rem

packages/web-components/src/components/cbp-toast/cbp-toast.scss
bagrub: updated to use the --cbp-border-radius-softer token for border radius

packages/web-components/src/components/cbp-toast/cbp-toast.stories.tsx
bagrub: updated to use ternaries as discussed

packages/web-components/src/components/cbp-toast/cbp-toast.stories.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-toast/cbp-toast.stories.tsx
bagrub: added control for title

packages/web-components/src/components/cbp-toast/cbp-toast.tsx
bagrub: updated icon to slotted and set in story as discussed in call

packages/web-components/src/components/cbp-toast/cbp-toast.scss
```diff
@@ -72,11 +72,15 @@ cbp-toast {
...
   .cbp-toast-sidebar{
     display: flex;
+    align-items: center;
     padding: var(--cbp-space-4x);
-    color: var(--cbp-toast-color-icon-sidebar);
     background-color: var(--cbp-toast-color-bg-sidebar);
     border-top-left-radius: 5px;
     border-bottom-left-radius: 5px;
+
+    cbp-icon{
```
dgibson666: ideally, you would target the slot by name and set the color and the icon will inherit it (because it defaults to currentColor)

packages/web-components/src/components/cbp-toast/cbp-toast.scss
```diff
@@ -72,11 +72,15 @@ cbp-toast {
 
   .cbp-toast-sidebar{
     display: flex;
+    align-items: center;
     padding: var(--cbp-space-4x);
-    color: var(--cbp-toast-color-icon-sidebar);
     background-color: var(--cbp-toast-color-bg-sidebar);
     border-top-left-radius: 5px;
```
dgibson666: Still some hardcoded border radius values here

packages/web-components/src/components/cbp-toast/cbp-toast.stories.tsx
```diff
@@ -22,17 +22,18 @@ export default {
...
+            ${open ? `open` : ''}
+            ${color ? `color=${color}` : ''}
+            ${duration ? `duration=${duration}` : ''}
             ${icon ? `icon=${icon}` : ''}
             ${context && context != 'light-inverts' ? `context=${context}` : ''}
             ${sx ? `sx=${JSON.stringify(sx)}` : ''}
-          >
-            <div slot="cbp-toast-title"> Test Toast Title</div>
+          >  
+            <div slot="cbp-toast-icon">${icon}</div>
```
dgibson666: I would just pass the icon name here, not the whole icon component. Just code the icon component into the story. And remember that you cannot self-close web components/custom elements (only in JSX). You need a full closing tag.

packages/web-components/src/components/cbp-toast/cbp-toast.stories.tsx
```diff
@@ -43,12 +44,13 @@ export default {
   
   Toast.args = {
     open: true,
-    icon: 'user',
+    icon: `<cbp-icon size='2rem' name='user'/>`,
```
dgibson666: revert this per previous comment. Just pass the name, render the icon in the story.

dgibson666 (commented):
See code comments

packages/web-components/src/components/cbp-toast/cbp-toast.stories.tsx
bagrub: updated as requested, this seems to be where the rogue > was coming from as well

packages/web-components/src/components/cbp-toast/cbp-toast.scss
bagrub: updated to style by slot not by cbp-icon

dgibson666 (approved):

PR #223, opened 2024-10-31
Usa banner and structured list update

bagrub:
#### Addressing 2 tickets for USA banner update & structured list update as outlined below:

USA banner update:
[*]Not sure what is determining the height of the overall component, for example in the XL breakpoint the text line height is 16px with padding of 8px on the y axis. so that should mean the component is 32px tall. but it is rendering at 36px for some reason. cant determine how.
[*]the font weight is incorrect. should be medium.
[*]the hover state of the text "heres how you know..." is incorrect. should be styled exactly like a text link (except smaller). remove the background fill, change text color to primary-lighter on hover.
[*]the focus state of the text is also incorrect. remove the background fill, change text color to focus-light, apply the outline like the text link has.

Structured list update: 
[*]the calendar collection is missing still. 
[*]the image sizing on the media version of the collection is off. the image is not expected to grow with the collection.
[*]the header should be displayed by default with the selectable structured list. 
[*]i know doug is currently working on some fixes with the checkbox but its overall area it occupies with the hover/focus halo is 44x44px. 
[*]based on this the right and left padding on the checkbox area should be 4px on each side. (the 44px focus/hover area not the 24px checkbox itself) same with the header checkbox so the stay aligned with each other.

### DIFF
--- a/packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
+++ b/packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
@@ -102,14 +102,47 @@ cbp-structured-list {
     }
   }
 
-  /* 
-    *Collection 
-  */
+  /**Collection */
 
   div[role=list]:only-child > *:first-child {
     border-top: var(--cbp-border-size-md) solid var(--cbp-structured-list-border-color);
   }
 
+  .cbp-structured-list-calendar-item{
+    width: 6.25rem;
+    height: 6.25rem;
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+
+    .cbp-structured-list-calendar-item-month{
+      width: 100%;
+      text-align: center;
+      border-top-left-radius: var(--cbp-border-radius-soft);
+      border-top-right-radius: var(--cbp-border-radius-soft);
+      padding-block: var(--cbp-space-2x);
+      background-color: var(--cbp-color-red-50);
+      border: 1px solid var(--cbp-color-red-50);
+      color: var(--cbp-color-white);
+      text-transform: uppercase;
+      font-size: var(---cbp-font-size-heading-xl);
+      font-weight: var(--cbp-font-weight-regular);
+    }
+
+    .cbp-structured-list-calendar-item-day{
+      width: 100%;
+      height: 4.25rem;
+      text-align: center;
+      border-bottom-left-radius: var(--cbp-border-radius-soft);
+      border-bottom-right-radius: var(--cbp-border-radius-soft);
+      border: 2px solid var(--cbp-color-gray-cool-40);
+      border-top: none;
+      color: var(--cbp-structured-list-color);
+      font-size: var(--cbp-font-size-masthead-1);
+      font-weight: var(--cbp-font-weight-black);
+      line-height: 1.75;
+    }
+  }
   /** Selectable */
 
   &[selectable] [slot=cbp-structured-list-header] {
--- a/packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
+++ b/packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
@@ -107,42 +107,7 @@ cbp-structured-list {
   div[role=list]:only-child > *:first-child {
     border-top: var(--cbp-border-size-md) solid var(--cbp-structured-list-border-color);
   }
-
-  .cbp-structured-list-calendar-item{
-    width: 6.25rem;
-    height: 6.25rem;
-    display: flex;
-    flex-direction: column;
-    align-items: center;
-
-    .cbp-structured-list-calendar-item-month{
-      width: 100%;
-      text-align: center;
-      border-top-left-radius: var(--cbp-border-radius-soft);
-      border-top-right-radius: var(--cbp-border-radius-soft);
-      padding-block: var(--cbp-space-2x);
-      background-color: var(--cbp-color-red-50);
-      border: 1px solid var(--cbp-color-red-50);
-      color: var(--cbp-color-white);
-      text-transform: uppercase;
-      font-size: var(---cbp-font-size-heading-xl);
-      font-weight: var(--cbp-font-weight-regular);
-    }
-
-    .cbp-structured-list-calendar-item-day{
-      width: 100%;
-      height: 4.25rem;
-      text-align: center;
-      border-bottom-left-radius: var(--cbp-border-radius-soft);
-      border-bottom-right-radius: var(--cbp-border-radius-soft);
-      border: 2px solid var(--cbp-color-gray-cool-40);
-      border-top: none;
-      color: var(--cbp-structured-list-color);
-      font-size: var(--cbp-font-size-masthead-1);
-      font-weight: var(--cbp-font-weight-black);
-      line-height: 1.75;
-    }
-  }
+  
   /** Selectable */
 
   &[selectable] [slot=cbp-structured-list-header] {
--- a/packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
+++ b/packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
@@ -363,7 +363,12 @@ StructuredListMedia.args = {
   showHeader: false,
   listItems: [
     {
-      content: "<cbp-grid grid-template-columns='1fr 3fr' gap='1rem'><div><img src='./assets/images/cbp-seal.svg' /></div><div><cbp-typography tag='p'><cbp-icon name='arrow-right' color='var(--cbp-link-color)' size='1rem'></cbp-icon><cbp-link href='#' target='_self' style='margin-inline-start: 0.5rem'>CBP Seizes over 2 Million Pounds of Cocaine</cbp-link></cbp-typography><cbp-typography tag='p'>CBP officers out of the Port of Baltimore Narcotics task force seized more than 2 million pounds of cocaine tuesday. It is one of the largest single seizures of cocaine ever recorded</cbp-typography><cbp-typography tag='p'><cbp-icon name='globe' size='1rem'></cbp-icon> <i>https://www.cbp.dhs.gov/news/cocaine</i></cbp-typography><cbp-tag> News</cbp-tag></div><cbp-grid>",
+      content: "<cbp-grid grid-template-columns='1fr 5fr' gap='1rem'><div><img src='./assets/images/cbp-seal.svg' style='width: 100px'/></div><div><cbp-typography tag='p'><cbp-icon name='arrow-right' color='var(--cbp-link-color)' size='1rem'></cbp-icon><cbp-link href='#' target='_self' style='margin-inline-start: 0.5rem'>CBP Seizes over 2 Million Pounds of Cocaine</cbp-link></cbp-typography><cbp-typography tag='p'>CBP officers out of the Port of Baltimore Narcotics task force seized more than 2 million pounds of cocaine tuesday. It is one of the largest single seizures of cocaine ever recorded</cbp-typography><cbp-typography tag='p'><cbp-icon name='globe' size='1rem'></cbp-icon> <i>https://www.cbp.dhs.gov/news/cocaine</i></cbp-typography><cbp-tag> News</cbp-tag></div><cbp-grid>",
+      color: 'default',
+      selected: false
+    },
+    {
+      content: "<cbp-grid grid-template-columns='1fr 5fr' gap='1rem'><div><img src='./assets/images/cbp-seal.svg' style='width: 100px'/></div><div><cbp-typography tag='p'><cbp-icon name='arrow-right' color='var(--cbp-link-color)' size='1rem'></cbp-icon><cbp-link href='#' target='_self' style='margin-inline-start: 0.5rem'>CBP Seizes over 2 Million Pounds of Cocaine</cbp-link></cbp-typography><cbp-typography tag='p'>CBP officers out of the Port of Baltimore Narcotics task force seized more than 2 million pounds of cocaine tuesday. It is one of the largest single seizures of cocaine ever recorded</cbp-typography><cbp-typography tag='p'><cbp-icon name='globe' size='1rem'></cbp-icon> <i>https://www.cbp.dhs.gov/news/cocaine</i></cbp-typography><cbp-tag> News</cbp-tag></div><cbp-grid>",
       color: 'default',
       selected: false
     },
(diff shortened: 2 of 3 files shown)


packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -363,12 +363,12 @@ StructuredListMedia.args = {
...
+      content: "<cbp-grid grid-template-columns='1fr 5fr' gap='1rem'><div><img src='./assets/images/cbp-seal.svg' style='width: 100px'/></div><div><cbp-typography tag='p'><cbp-icon name='arrow-right' color='var(--cbp-link-color)' size='1rem'></cbp-icon><cbp-link href='#' target='_self' style='margin-inline-start: 0.5rem'>CBP Seizes over 2 Million Pounds of Cocaine</cbp-link></cbp-typography><cbp-typography tag='p'>CBP officers out of the Port of Baltimore Narcotics task force seized more than 2 million pounds of cocaine tuesday. It is one of the largest single seizures of cocaine ever recorded</cbp-typography><cbp-typography tag='p'><cbp-icon name='globe' size='1rem'></cbp-icon> <i>https://www.cbp.dhs.gov/news/cocaine</i></cbp-typography><cbp-tag> News</cbp-tag></div><cbp-grid>",
```
dgibson666: It looks like you've updated the first 2, but not the last 2, so the images are different sizes. It doesn't make sense that you're sizing the img to pixels but the grid it using fractional units. There's no "flex-shrink: 0" in grid, so the images shrink to 0 if you resize it down. Use a fixed unit in the first column (e.g., grid-template-columns='100px 1fr') for the demo and then you can remove the style attribute from the img tag.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
```diff
@@ -102,14 +102,47 @@ cbp-structured-list {
...
-  /* 
-    *Collection 
-  */
+  /**Collection */
 
   div[role=list]:only-child > *:first-child {
     border-top: var(--cbp-border-size-md) solid var(--cbp-structured-list-border-color);
   }
 
+  .cbp-structured-list-calendar-item{
```
dgibson666: We need to discuss this. None of these classes that belong to slotted content should live in the component CSS. You've got a bunch of values that should be tokens too.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -385,6 +385,51 @@ StructuredListMedia.args = {
...
+  listItems: {
+    description: 'Configure various aspects of the list items within the structured list.',
+    control: 'object',
+  },
+}
+StructuredListMediaCalendar.args = { //todo: update text in samples to match calendar text
+  showHeader: false,
+  listItems: [
+    {
+      content: "<cbp-grid grid-template-columns='1fr 5fr' gap='1rem'><div class='cbp-structured-list-calendar-item'><span class='cbp-structured-list-calendar-item-month'>apr</span><span class='cbp-structured-list-calendar-item-day'>31</span></div><div><cbp-typography tag='p'><cbp-icon name='arrow-right' color='var(--cbp-link-color)' size='1rem'></cbp-icon><cbp-link href='#' target='_self' style='margin-inline-start: 0.5rem'>Find the Hidden Beetle Training</cbp-link></cbp-typography><cbp-typography tag='p'>This training session will help you find those pesky hiding beetles. Inside boxes, on top of containers, or in someone's pocket, these bugs don't stand a chance of getting to America's sweet, delicious forests</cbp-typography><cbp-typography tag='p'><cbp-icon name='globe' size='1rem'></cbp-icon> <i>https://www.cbp.dhs.gov/news/beetles</i></cbp-typography></div><cbp-grid>",
```
dgibson666: These have the same issue with the grid as the media story. Use a fixed (relative units in this case since it's text) first column. I'm thinking more about the faux calendar item. This could probably be made as a "pattern" using the flex components and the sx prop to hook in all the CSS. It may be ugly with the amount of CSS, but that CSS doesn't belong to the component.

packages/web-components/src/components/cbp-usa-banner/cbp-usa-banner.scss
```diff
@@ -36,11 +40,13 @@ cbp-usa-banner {
...
+  cbp-button.cbp-usa-banner-expand[fill='ghost'][color='primary']:not(#fakeId)  { 
+    //addition of fill & color in the selector is to up specificity to override button's dark mode var remapping correctly
+    
     --cbp-button-height: 1rem;
     --cbp-button-min-height: 1rem;
 
-    --cbp-button-padding: var(--cbp-space-2x);
+    --cbp-button-padding: var(--cbp-space-1x);
 
     --cbp-button-color: var(--cbp-color-interactive-secondary-light);
```
dgibson666: these don't look right. The control is supposed to be blue like a link. See comments in the Jira ticket for focus values.

dgibson666 (commented):
See code comments.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: updated columns to '100px 1fr' as requested

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
bagrub: As discussed on call updated the storybook content blocks to use flex components & sx to style the calendar items

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: see above

packages/web-components/src/components/cbp-usa-banner/cbp-usa-banner.scss
bagrub: updated color mappings to for enabled to interactive-primary-light and focus to interactive-focus-light (note: this var was explicitly requested for this mapping and results in focus being light blue not purple as in mockups)

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -409,28 +409,206 @@ StructuredListMediaCalendar.argTypes = {
...
   showHeader: false,
   listItems: [
     {
-      content: "<cbp-grid grid-template-columns='1fr 5fr' gap='1rem'><div class='cbp-structured-list-calendar-item'><span class='cbp-structured-list-calendar-item-month'>apr</span><span class='cbp-structured-list-calendar-item-day'>31</span></div><div><cbp-typography tag='p'><cbp-icon name='arrow-right' color='var(--cbp-link-color)' size='1rem'></cbp-icon><cbp-link href='#' target='_self' style='margin-inline-start: 0.5rem'>Find the Hidden Beetle Training</cbp-link></cbp-typography><cbp-typography tag='p'>This training session will help you find those pesky hiding beetles. Inside boxes, on top of containers, or in someone's pocket, these bugs don't stand a chance of getting to America's sweet, delicious forests</cbp-typography><cbp-typography tag='p'><cbp-icon name='globe' size='1rem'></cbp-icon> <i>https://www.cbp.dhs.gov/news/beetles</i></cbp-typography></div><cbp-grid>",
+      content: `<cbp-grid grid-template-columns='1fr 5fr' gap='1rem'>
+                  <cbp-flex
+                  sx='{
+                        "width":"6.25rem",
+                        "height":"6.25rem",
+                        "display":"flex",
```
dgibson666: you don't need display: flex since you are using the flex component. flex-direction and align-items are implemented as props on this component, so use them there (better readability).

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -409,28 +409,206 @@ StructuredListMediaCalendar.argTypes = {
...
+                        "flex-direction":"column",
+                        "align-items":"center"
+                      }'
+                >
+                  <cbp-flex-item
+                  sx='{
+                        "width":"100%",
+                        "text-align":"center",
+                        "border-top-left-radius":"var(--cbp-border-radius-soft)",
```
dgibson666: combine border-radius into 1 line of shorthand: var(--cbp-border-radius-soft) var(--cbp-border-radius-soft) 0 0;

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -409,28 +409,206 @@ StructuredListMediaCalendar.argTypes = {
...
+                        "border-top-right-radius":"var(--cbp-border-radius-soft)",
+                        "padding-block":"var(--cbp-space-2x)",
+                        "background-color":"var(--cbp-color-red-50)",
+                        "border":"1px solid var(--cbp-color-red-50)",
```
dgibson666: border size 1px should have a token value.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -409,28 +409,206 @@ StructuredListMediaCalendar.argTypes = {
...
+                  </cbp-flex-item>
+                  
+                  <cbp-flex-item
+                    sx='{
+                          "width":"100%",
+                          "height":"4.25rem",
+                          "text-align":"center",
+                          "border-bottom-left-radius":"var(--cbp-border-radius-soft)",
+                          "border-bottom-right-radius":"var(--cbp-border-radius-soft)",
+                          "border-width":"2px",
```
dgibson666: token and why are these broken out into separate props instead of using border shorthand?

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -409,28 +409,206 @@ StructuredListMediaCalendar.argTypes = {
...
+                  > 
+                    apr
+                  </cbp-flex-item>
+                  
+                  <cbp-flex-item
+                    sx='{
+                          "width":"100%",
+                          "height":"4.25rem",
+                          "text-align":"center",
+                          "border-bottom-left-radius":"var(--cbp-border-radius-soft)",
```
dgibson666: use shorthand

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -409,28 +409,206 @@ StructuredListMediaCalendar.argTypes = {
...
+                          "border-bottom-right-radius":"var(--cbp-border-radius-soft)",
+                          "border-width":"2px",
+                          "border-style":"solid",
+                          "border-color":"var(-cbp-color-gray-cool-40)",
+                          "border-top":"none",
+                          "color":"var(--cbp-structured-list-color)",
+                          "font-size":"var(--cbp-font-size-masthead-1)",
+                          "font-weight":"var(--cbp-font-weight-black)",
+                          "line-height":"1.75"
```
dgibson666: token?

packages/web-components/src/components/cbp-usa-banner/cbp-usa-banner.scss
```diff
@@ -48,17 +48,17 @@ cbp-usa-banner {
...
+    --cbp-button-color-focus: var(--cbp-color-interactive-focus-light);
 
     --cbp-button-color-border-hover: transparent;
     --cbp-button-color-border-focus: transparent;
     --cbp-button-color-border-active: transparent;
 
     --cbp-button-border-radius: 0;
 
-    --cbp-button-color-outline-focus: var(--cbp-color-violet-30);
+    --cbp-button-color-outline-focus: var(--cbp-color-interactive-focus-light);
```
dgibson666, line 63: The active/clicked color needs updating too. Right now, while the mouse is pressed down on the control, the text turns black and essentially disappears. There should be a var for that as well (active I think).

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -409,28 +409,206 @@ StructuredListMediaCalendar.argTypes = {
     control: 'object',
   },
 }
-StructuredListMediaCalendar.args = { //todo: update text in samples to match calendar text
+StructuredListMediaCalendar.args = {
   showHeader: false,
   listItems: [
     {
-      content: "<cbp-grid grid-template-columns='1fr 5fr' gap='1rem'><div class='cbp-structured-list-calendar-item'><span class='cbp-structured-list-calendar-item-month'>apr</span><span class='cbp-structured-list-calendar-item-day'>31</span></div><div><cbp-typography tag='p'><cbp-icon name='arrow-right' color='var(--cbp-link-color)' size='1rem'></cbp-icon><cbp-link href='#' target='_self' style='margin-inline-start: 0.5rem'>Find the Hidden Beetle Training</cbp-link></cbp-typography><cbp-typography tag='p'>This training session will help you find those pesky hiding beetles. Inside boxes, on top of containers, or in someone's pocket, these bugs don't stand a chance of getting to America's sweet, delicious forests</cbp-typography><cbp-typography tag='p'><cbp-icon name='globe' size='1rem'></cbp-icon> <i>https://www.cbp.dhs.gov/news/beetles</i></cbp-typography></div><cbp-grid>",
+      content: `<cbp-grid grid-template-columns='1fr 5fr' gap='1rem'>
```
dgibson666: This grid should still not have a 1fr for the first column. It should be fixed, but in relative units (rems). If you view this on a large screen, the gap between the calendar and text grows, which is not ideal.

dgibson666 (commented):
More code comments after testing.

packages/web-components/src/components/cbp-usa-banner/cbp-usa-banner.scss
bagrub: updated to override the --cbp-button-color-active token to the focus color to avoid the described behavior

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: updated to 6.25rem 1fr

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: update display, flex-direction, & align to props, removed the SX since width is being set on grid and I noticed some weirdness having the above set via prop the SX seemed to override those inline-styles instead of adding to

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: updated

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: updated to use var(--cbp-border-size-sm)

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: unitless values for line height don't have vars, discussed on call to confirm

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: discussed on stand up this morning, pushing with values broken out separately due to how the sx is parsing the border values

dgibson666 (approved):
looks good.

PR #225, opened 2024-11-04
Bugfix/check radio

dgibson666:
* Updated the methods of displaying checkbox and radio checked states. It didn't actually fix the off-by-1px issue when base font size is set to "large" (1.25 multiplier), but it is less CSS overall.
* Updated setCSSProps to skip properties with value of undefined and only console.log when a prop is being set (!undefined).
* Added JSDocs for `cbp-flex` and `cbp-flex-item`, tweaked CSS.

After extensive testing of setCSSProps() and sx, I'm fairly certain that the issues seen are due to browser implementation and a side effect of using shorthand properties. I was able to reproduce the many phantom props being set to empty, but in my tests the visual output always displayed as intended still.

### DIFF
--- a/packages/web-components/CHANGELOG.md
+++ b/packages/web-components/CHANGELOG.md
@@ -4,13 +4,17 @@ This CHANGELOG.md tracks the updates to the web components package of the CBP de
 
 The React components are wrappers generated from this package and will share the same changes.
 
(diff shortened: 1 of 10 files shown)


bagrub (approved):
lgtm

PR #226, opened 2024-11-14
Tooltip

bagrub:
#### Creation of Tooltip component

### DIFF
--- a/packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
+++ b/packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
@@ -39,6 +39,8 @@
   --cbp-tooltip-control-color-focus-dark: var(--cbp-color-text-darkest);
   --cbp-tooltip-control-color-bg-focus: var(--cbp-color-interactive-focus-dark);
   --cbp-tooltip-control-color-bg-focus-dark: var(--cbp-color-interactive-focus-light);
+  --cbp-tooltip-control-outline-focus: 2px solid var(--cbp-color-white);
+  --cbp-tooltip-control-outline-focus-dark: 2px solid var(--cbp-color-black);
 
   --cbp-tooltip-definition-color: var(--cbp-color-interactive-primary-dark);
   --cbp-tooltip-definition-color-dark: var(--cbp-color-interactive-primary-light);
@@ -61,6 +63,7 @@
 
   --cbp-tooltip-control-color-focus: var(--cbp-tooltip-control-color-focus-dark);
   --cbp-tooltip-control-color-bg-focus: var(--cbp-tooltip-control-color-bg-focus-dark);
+  --cbp-tooltip-control-outline-focus: var(--cbp-tooltip-control-outline-focus-dark);
 
   --cbp-tooltip-definition-color: var(--cbp-tooltip-definition-color-dark);
   --cbp-tooltip-definition-color-hover: var(--cbp-tooltip-definition-color-hover-dark);
@@ -89,6 +92,7 @@ cbp-tooltip {
   &:focus-within{
     color: var(--cbp-tooltip-control-color-focus);
     background: var(--cbp-tooltip-control-color-bg-focus);
+    outline:var(--cbp-tooltip-control-outline-focus);
   }
 
   /** definition link*/
--- a/packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
+++ b/packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
@@ -8,17 +8,16 @@
  * @Prop   --cbp-tooltip-padding: var(--cbp-space-inset-4x, 1rem);
  * @Prop   --cbp-tooltip-gap: 0px;
  * @Prop   --cbp-tooltip-offset: 0px;
- * @Prop   --cbp-tooltip-link-decoration: none;
  * @Prop   --cbp-tooltip-caret-size: 15px;
  * @Prop   --cbp-tooltip-caret-offset: calc(var(--cbp-tooltip-caret-size) / 2);
- * @Prop   --cbp-tooltip-color-hover: var(--cbp-color-interactive-secondary-darker);
- * @Prop   --cbp-tooltip-color-hover-dark: var(--cbp-color-interactive-secondary-lighter);
- * @Prop   --cbp-tooltip-color-bg-hover: var(--cbp-color-interactive-secondary-lighter);
- * @Prop   --cbp-tooltip-color-bg-hover-dark: var(--cbp-color-interactive-disabled-light);
- * @Prop   --cbp-tooltip-color-focus: var(--cbp-color-text-lightest);
- * @Prop   --cbp-tooltip-color-focus-dark: var(--cbp-color-text-darkest);
- * @Prop   --cbp-tooltip-color-bg-focus: var(--cbp-color-interactive-focus-dark);
- * @Prop   --cbp-tooltip-color-bg-focus-dark: var(--cbp-color-interactive-focus-light);
+ * @Prop   --cbp-tooltip-control-color-hover: var(--cbp-color-interactive-secondary-darker);
+ * @Prop   --cbp-tooltip-control-color-hover-dark: var(--cbp-color-interactive-secondary-lighter);
+ * @Prop   --cbp-tooltip-control-color-bg-hover: var(--cbp-color-interactive-secondary-lighter);
+ * @Prop   --cbp-tooltip-control-color-bg-hover-dark: var(--cbp-color-interactive-disabled-light);
+ * @Prop   --cbp-tooltip-control-color-focus: var(--cbp-color-text-lightest);
+ * @Prop   --cbp-tooltip-control-color-focus-dark: var(--cbp-color-text-darkest);
+ * @Prop   --cbp-tooltip-control-color-bg-focus: var(--cbp-color-interactive-focus-dark);
+ * @Prop   --cbp-tooltip-control-color-bg-focus-dark: var(--cbp-color-interactive-focus-light);
  */
  :root{
  
@@ -28,20 +27,27 @@
   --cbp-tooltip-color-bg-dark: var(--cbp-color-gray-cool-5);
   --cbp-tooltip-gap: 0px;
   --cbp-tooltip-offset: 0px;
-  --cbp-tooltip-link-decoration: none;
   --cbp-tooltip-caret-size: 15px;
   --cbp-tooltip-caret-offset: calc(var(--cbp-tooltip-caret-size) / 2);
 
-  --cbp-tooltip-color-hover: var(--cbp-color-interactive-secondary-darker);
-  --cbp-tooltip-color-hover-dark: var(--cbp-color-interactive-secondary-lighter);
-  --cbp-tooltip-color-bg-hover: var(--cbp-color-interactive-secondary-lighter);
-  --cbp-tooltip-color-bg-hover-dark: var(--cbp-color-interactive-disabled-light);
+  --cbp-tooltip-control-color-hover: var(--cbp-color-interactive-secondary-darker);
+  --cbp-tooltip-control-color-hover-dark: var(--cbp-color-interactive-secondary-lighter);
+  --cbp-tooltip-control-color-bg-hover: var(--cbp-color-interactive-secondary-lighter);
+  --cbp-tooltip-control-color-bg-hover-dark: var(--cbp-color-interactive-disabled-light);
 
-  --cbp-tooltip-color-focus: var(--cbp-color-text-lightest);
-  --cbp-tooltip-color-focus-dark: var(--cbp-color-text-darkest);
-  --cbp-tooltip-color-bg-focus: var(--cbp-color-interactive-focus-dark);
-  --cbp-tooltip-color-bg-focus-dark: var(--cbp-color-interactive-focus-light);
-  
+  --cbp-tooltip-control-color-focus: var(--cbp-color-text-lightest);
+  --cbp-tooltip-control-color-focus-dark: var(--cbp-color-text-darkest);
+  --cbp-tooltip-control-color-bg-focus: var(--cbp-color-interactive-focus-dark);
+  --cbp-tooltip-control-color-bg-focus-dark: var(--cbp-color-interactive-focus-light);
+
+  --cbp-tooltip-definition-color: var(--cbp-color-interactive-primary-dark);
+  --cbp-tooltip-definition-color-dark: var(--cbp-color-interactive-primary-light);
+  --cbp-tooltip-definition-color-hover: var(--cbp-color-interactive-primary-darker);
+  --cbp-tooltip-definition-color-hover-dark: var(--cbp-color-interactive-primary-lighter);
+  --cbp-tooltip-definition-color-focus: var(--cbp-color-interactive-focus-dark);
+  --cbp-tooltip-definition-color-focus-dark: var(--cbp-color-interactive-focus-light);
+  --cbp-tooltip-definition-outline-focus: 2px solid var(--cbp-color-interactive-focus-dark);
+  --cbp-tooltip-definition-outline-focus-dark: 2px solid var(--cbp-color-interactive-focus-light);
 }
 
 [data-cbp-theme=light] cbp-tooltip[context*=dark]:not([context=light-always]),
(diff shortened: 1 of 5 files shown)


packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
```diff
@@ -0,0 +1,248 @@
...
+ * @Prop   --cbp-tooltip-color-hover-dark: var(--cbp-color-interactive-secondary-lighter);
+ * @Prop   --cbp-tooltip-color-bg-hover: var(--cbp-color-interactive-secondary-lighter);
+ * @Prop   --cbp-tooltip-color-bg-hover-dark: var(--cbp-color-interactive-disabled-light);
+ * @Prop   --cbp-tooltip-color-focus: var(--cbp-color-text-lightest);
+ * @Prop   --cbp-tooltip-color-focus-dark: var(--cbp-color-text-darkest);
+ * @Prop   --cbp-tooltip-color-bg-focus: var(--cbp-color-interactive-focus-dark);
+ * @Prop   --cbp-tooltip-color-bg-focus-dark: var(--cbp-color-interactive-focus-light);
+ */
+ :root{
+
```
dgibson666, line 23: these variables are not clearly named. I think you might want to specify --cbp-tooltip-content and --cbp-tooltip-control to make it clear what they represent.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
```diff
@@ -0,0 +1,248 @@
...
+  --cbp-tooltip-color-focus: var(--cbp-tooltip-color-focus-dark);
+  --cbp-tooltip-color-bg-focus: var(--cbp-tooltip-color-bg-focus-dark);
+}
+
+cbp-tooltip {
+  display: inline-block;
+  position: relative;
+  --cbp-tooltip-placement-gap: var(--cbp-tooltip-gap);
+  --cbp-tooltip-placement-offset: var(--cbp-tooltip-offset);
+  --cbp-link-decoration: var(--cbp-tooltip-link-decoration);
```
dgibson666: this is not acting on a cbp-link (or should not be) component so it should not be named as such.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+        sx: {
+            description: 'Supports adding inline styles as an object of key-value pairs comprised of CSS properties and values. Values should reference design tokens when possible.',
+            control: 'object',
+        },
+    },
+  };
+
+const Template = ({ open, uid, alignment, title, content, buttonContent, context, sx }) => {
+return ` 
+        <a href='#' style='display: inline-block; margin-left:40%; margin-top:20%'> item #1</a>
```
dgibson666: clean this story up. don't need the anchors anymore.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+};
+  
+export const Tooltip = Template.bind({});
+  
+Tooltip.args = {
+    uid: 'tooltip',
+    alignment: 'top-left',
+    title: 'Test Tooltip Title',
+    content: 'Stub text for tooltip.',
+    buttonContent: '<cbp-icon name="user"></cbp-icon>',
```
dgibson666: this should be renamed as "tooltip control" or the like.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -0,0 +1,74 @@
...
+  
+  @Element() host: HTMLElement;
+  /** When set, specifies that the tooltip is open */
+  @Prop({ reflect: true }) open: boolean = false;
+ 
+  //todo: update from alignment to placement
+  @Prop({ reflect: true}) alignment: "top-left" | "top-center" | "top-left" | "right-top" | "right-center" | "right-bottom" | "bottom-left" | "bottom-center" | "bottom-right" | "left-top" | "left-center" | "left-bottom";
+  
+  /** Specifies a unique `ID` for the tooltip, used to wire up the controls and accessibility features. */
+  @Prop() uid: string;
```
dgibson666: this should have a default using createUUID() (?) function coming from utils. I'm not sure it needs to be a prop, as it's being implemented in the wrong place.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -0,0 +1,74 @@
...
+    console.log('oh no it blew up');
+    this.host.focus();
+    this.open =false;
+  }
+
+  render() {
+    
+    return (
+      <Host 
+        id={this.uid}
```
dgibson666: The id serves no purpose here and could be applied by the developer as a native `id` if desired. This should go on the `role=tooltip` element and an `aria-describedby={this.uid}` would go here.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -0,0 +1,74 @@
...
+    if (typeof this.sx == 'string') {
+      this.sx = JSON.parse(this.sx) || {};
+    }
+    setCSSProps(this.host, {
+      ...this.sx,
+    });
+  }
+
+  dismissTooltip(){
+    console.log('oh no it blew up');
```
dgibson666: remove

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -0,0 +1,74 @@
...
+          </div>
+          <cbp-button
+            class='cbp-tooltip-close'
+            type="button"
+            fill="ghost"
+            color="secondary" 
+            context={(this.context == 'dark-always' || this.context == 'dark-inverts')? "light-always":"dark-always"}
+            variant="square"
+            onClick={() => this.dismissTooltip()}
+            onFocusout={() => this.dismissTooltip()}
```
dgibson666: I was wondering how you were doing this, but this is not correct. If I shift-tab to go backwards in the tab order, it still closes, even though it should be focusing on tooltip focusable content or the control again. I think we have tab-out detection working somewhere but will need to look for it.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.specs.mdx
```diff
@@ -0,0 +1,31 @@
+import { Meta } from '@storybook/addon-docs';
+
+<Meta title="Components/Tooltip/Specifications" />
+
```
dgibson666, line 4: Specs need to be cleaned up and fleshed out. Use proper capitalization and punctuation. 

The purpose should probably be discussed for specific wording with clarifying guidance from Jeremy.

The accessibility section especially should reference the role, tabindex, keyboard nav, etc.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+            alignment=${alignment}
+            ${context && context != 'light-inverts' ? `context=${context}` : ''}
+            ${sx ? `sx=${JSON.stringify(sx)}` : ``}
+
+            >  
+            <div>
+                ${buttonContent}
+            </div>
+            <div slot="cbp-tooltip-content">
+                <div style='font-weight: var(--cbp-font-weight-bold)'>${title}</div>
```
dgibson666, line 44: I feel like this text is too tight using divs with no line-height applied. I don't know if you can put a line height on the `role=tooltip` element and it inherits down or if you should use paragraph tags on them and remove bottom margin on the last child.

dgibson666 (commented):
- See code comments. 
- Still needs ESC to close.
- The dismiss button isn't visible in dark mode.
- Need a story for definition tooltip and a way to style text instead of a button-like control. There's no good reason to do this separately since it's known up front.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
bagrub: update var names to distinguish between  content & control

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
bagrub: removed this var, looks like it was set to none regardless, i think this was carry over from the original style sheet

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.stories.tsx
bagrub: removed <a> tags

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.stories.tsx
bagrub: renamed to tooltip control

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: removed

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.stories.tsx
bagrub: added line-height to this

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: Worked through this on call to include  handleFocusOut() to manage this event

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: updated to use createNamespaceKey as discussed

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.specs.mdx
bagrub: update doc based on UX spec details & reviewed for capitalization & punctuation

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
```diff
@@ -50,34 +56,59 @@
...
   position: relative;
   --cbp-tooltip-placement-gap: var(--cbp-tooltip-gap);
   --cbp-tooltip-placement-offset: var(--cbp-tooltip-offset);
-  --cbp-link-decoration: var(--cbp-tooltip-link-decoration);
 
   &[open] [role='tooltip']{
       display: flex;
   }
     
+  &:not([definitionlinkstyle="true"]):hover{
```
dgibson666: don't make this a `:not()`. Just assign the default with no attribute selector and override with the attribute selector.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -59,10 +70,10 @@ export class CbpTooltip {
             type="button"
             fill="ghost"
             color="secondary" 
-            context={(this.context == 'dark-always' || this.context == 'dark-inverts')? "light-always":"dark-always"}
+            context={(document.getElementsByTagName("cbp-app")[0].theme == 'dark' || this.context == 'dark-always' || this.context == 'dark-inverts')? "light-always":"dark-always"}
```
dgibson666: I've never had to do this sort of logic before. I think `dark-inverts` should cover it. This represents the default state (it's a dark context, so the button will appear in dark mode when the user is in light mode) but it inverts (to light) when the user switches to dark mode.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -1,30 +1,32 @@
...
 
-  
   @Element() host: HTMLElement;
   /** When set, specifies that the tooltip is open */
   @Prop({ reflect: true }) open: boolean = false;
- 
-  //todo: update from alignment to placement
+
+  /** used to set styles for the definition link for text controls*/
+  @Prop({ reflect: true }) definitionLinkStyle: boolean = false;
```
dgibson666: I know naming is hard but this is way too verbose. I try to keep things to a single word to act more like HTML and not have a bunch of hyphens. I think I'd go with `variant: 'default' | 'definition' = 'default'` (then you don't need to use the "default" attribute selector when setting styles, only `[variant=definition]`. This also leaves room for expansion in case we do want to add a true link style (or something else) later.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.stories.tsx
```diff
@@ -56,9 +54,45 @@ Tooltip.args = {
     alignment: 'top-left',
     title: 'Test Tooltip Title',
     content: 'Stub text for tooltip.',
-    buttonContent: '<cbp-icon name="user"></cbp-icon>',
-    // sx:{
-    //     "margin-left":"40%",
-    //     "margin-top":"20%"
-    // }   
+    tooltipControl: '<cbp-icon name="user"></cbp-icon>',
+    sx:{
```
dgibson666: you don't want this in your story (I was wondering why the `sx` controls was reordered higher in the list). You can center your story with: 

```
 parameters: {
    layout: 'centered',
  },
```

(will apply to both stories)

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.specs.mdx
```diff
@@ -6,26 +6,34 @@ import { Meta } from '@storybook/addon-docs';
 
 ## Purpose
 
-* Tooltip is intended to display additional content to clarify or provide detail for users
+* the tooltip component allows the development team to disclose supplemental, not essential or nessecary, information to a user via a
```
dgibson666: Capitalize all your bullets. Most should be proper sentences with punctuation.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
```diff
@@ -0,0 +1,283 @@
...
+  --cbp-tooltip-placement-left: calc(0% - var(--cbp-tooltip-placement-gap, 5px));
+  --cbp-tooltip-placement-transform-x: -100%;
+  --cbp-tooltip-placement-transform-y: 0%;
+  --cbp-tooltip-caret-rotation: 225deg;
+  --cbp-tooltip-caret-top: calc(75% - var(--cbp-tooltip-caret-offset));
+  --cbp-tooltip-caret-left: calc(100% - var(--cbp-tooltip-caret-offset))
+}
+}
+
+cbp-tooltip div[role=tooltip] {
```
dgibson666: This is confusing to me since it's nested inside of the main custom tag, it should be in the same block as the other CSS, not a new block/nesting context. Move it up top before the positioning - it doesn't matter that the placement isn't set until after it. Same for the subsequent blocks starting with `cbp-tooltip`.

dgibson666 (commented):
See code comments.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
bagrub: removed the :not()

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: discussed in call & marked this as tech debt

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
```diff
@@ -74,24 +74,25 @@ cbp-tooltip {
...
     color: var(--cbp-tooltip-control-color-hover);
     background: var(--cbp-tooltip-control-color-bg-hover);
   }
 
-  &:not([definitionlinkstyle="true"]):focus-visible,
-  &:not([definitionlinkstyle="true"]):focus-within{
+  &:focus-visible,
+  &:focus-within{
     color: var(--cbp-tooltip-control-color-focus);
     background: var(--cbp-tooltip-control-color-bg-focus);
```
dgibson666, line 101: where is your outline for the faux-button? I see a white outline in dark mode when it should be black(ish - same color as the icon).

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.stories.tsx
```diff
@@ -55,19 +55,16 @@ Tooltip.args = {
     title: 'Test Tooltip Title',
     content: 'Stub text for tooltip.',
     tooltipControl: '<cbp-icon name="user"></cbp-icon>',
-    sx:{
-        "margin-left":"40%",
-        "margin-top":"20%"
-    }   
 }
+Tooltip.parameters= { layout: 'centered'} //todo: not vertically aligned due to the cbp-app{min-height: 100vh}, same with both samples
```
dgibson666: This can go in the default export block as a key before argTypes.

dgibson666 (commented):
See code comments.

I'm also seeing some issues with interactive colors. For the icon, the hover color and bg are the same in dark mode. On focus, the outline is white but should be black.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
```diff
@@ -73,16 +78,29 @@
...
   --cbp-tooltip-placement-offset: var(--cbp-tooltip-offset);
   --cbp-tooltip-gap: var(--cbp-space-4x, 1rem);
   
   &[open] [role='tooltip']{
       display: flex;
+
+  }
+   
+  //set default tooltip control to look like square ghost button
+  &{
```
dgibson666: why is this in a separate selector? Just put it on the host selector.

dgibson666 (commented):
the button styles are bleeding into the definition styles. The focus ring on definition is square. You probably need to unset height, width, and padding on the definition variant.

dgibson666 (approved):
looks good.

PR #229, opened 2024-12-02
App Header Nav

bagrub:


packages/web-components/src/components/cbp-app-header-nav/cbp-app-header-nav.tsx
```diff
@@ -0,0 +1,69 @@
...
+    });
+  }
+
+  render() {
+    
+    if(this.variant === 'multi'){
+      return (
+        <Host>
+          <nav>
+            <cbp-button
```
dgibson666: When you move the buttons to the story, make sure you are coding them correctly. A type=button cannot have an href. This should be tag=a (no type) to have an href.

dgibson666 (commented):
Please rename your component to `cbp-nav-item`. This is more precise for what it is.

This component should be a wrapper that encapsulates styles and manages aria-current. Don't render a link or button. 

See code comment about how to use the button/links. Don't include the button with caret version since it's not working.

Add the new readme to the API docs (you can see any other component with children as an example)

packages/web-components/src/components/cbp-app-header-nav/cbp-app-header-nav.tsx
bagrub: updated the type to tag

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.scss
```diff
@@ -0,0 +1,68 @@
...
+ * @Prop   --cbp-nav-item-button-color-bg-hover: var(--cbp-color-interactive-secondary-lighter);
+ * @Prop   --cbp-nav-item-button-color-bg-hover-dark: var(--cbp-color-interactive-secondary-darker);
+    
+ * @Prop   --cbp-nav-item-button-border-bottom-hover: var(--cbp-color-interactive-secondary-darker);
+ * @Prop   --cbp-nav-item-button-border-bottom-hover-dark: var(--cbp-color-interactive-secondary-lighter);
+
+ */
+
+ :root {
+    --cbp-nav-item-button-color: var(--cbp-color-interactive-secondary-darker);
```
dgibson666: I don't think you need "button" in all of these names. Especially since these are links. You're not styling anything else, so it's really just extra.

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.scss
```diff
@@ -0,0 +1,68 @@
...
+
+cbp-nav-item {
+
+  /* remap cbp-button vars */
+  cbp-button[fill=ghost][color=secondary] {
+    --cbp-button-min-height: 3.5rem; 
+    --cbp-button-color: var(-nav-item-button-color);
+    --cbp-button-border-width: 0 0 var(--cbp-border-size-xl) 0;
+    --cbp-button-border-radius: var(--cbp-border-radius-sharp);
+
```
dgibson666, line 87: These aren't supposed to be all caps, so you need to unset that.

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.scss
```diff
@@ -0,0 +1,68 @@
...
+    font-size: var(--cbp-font-weight-bold);
+
+    &:hover{
+      --cbp-button-color-hover: var(--cbp-nav-item-button-color-hover);
+      --cbp-button-color-bg-hover: var(--cbp-nav-item-button-color-bg-hover);
+      --cbp-button-color-border-hover: var(--cbp-nav-item-button-border-bottom-hover);
```
dgibson666: I don't see a bottom border on hover.

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.scss
```diff
@@ -0,0 +1,68 @@
...
+    font-size: var(--cbp-font-weight-bold);
+
+    &:hover{
+      --cbp-button-color-hover: var(--cbp-nav-item-button-color-hover);
+      --cbp-button-color-bg-hover: var(--cbp-nav-item-button-color-bg-hover);
+      --cbp-button-color-border-hover: var(--cbp-nav-item-button-border-bottom-hover); 
+    }
+
+    &:active{
+      --cbp-button-focus-outline-width: 0;
```
dgibson666: This shouldn't be needed. The button design is the same as the nav item, so I'm guessing the outline is still there because focus is still present. In any case, it's the button's job to fix this if it's not correct.

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { setCSSProps } from '../../utils/utils';
+@Component({
+  tag: 'cbp-nav-item',
+  styleUrl: 'cbp-nav-item.scss',
+})
+
+export class CbpNavItem {
+
+  @Element() host: HTMLElement;
+
```
dgibson666, line 11: You should have a prop to denote the active/current state. Maybe "selected" to match others? This is reflected and used as a style hook for the "current" nav item. You also need to apply aria-current to the controls via the DOM (since it's slotted, not rendered) and manage removing aria-current when one is clicked and made the new current item. You can reference tabs for this.

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
```diff
@@ -23,17 +23,27 @@ const Template = ({ context, sx }) => {
...
-          text='Multi Nav Item'
-          href='/'
-        >
-        </cbp-app-header-nav>
-        <cbp-app-header-nav
-          text='Single Nav Item'
-          href='/'
-        >
-        </cbp-app-header-nav>
+        <cbp-nav-item>
```
dgibson666: I'd make these an array of objects and pass them in as an argument. Use # for the href so that it stays on the same page and can demo updating the selected/current state.

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.tsx
```diff
@@ -0,0 +1,32 @@
...
+    if (typeof this.sx == 'string') {
+      this.sx = JSON.parse(this.sx) || {};
+    }
+    setCSSProps(this.host, {
+      ...this.sx,
+    });
+  }
+
+  render() {
+    <Host>
```
dgibson666: needs an onclick listener to set the item as selected/current.

dgibson666 (commented):
See code comments.

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
bagrub: updated to use the arguments and hooked the up to the storybook controls

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.scss
bagrub: removed the -button- part from the nav-item cssvars names

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.scss
bagrub: added a text-transform: unset mirroring the cbp-button selector

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.scss
bagrub: discussed via call & added tech debt for this. button-dark was overriding the nav-item dark border. Assigned the vars to the --cbp-button-color-border-hover & --cbp-button-color-border-hover-dark explicitly instead of using their nav-item cssvars to resolve and left note describing the tech debt

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.scss
bagrub: removed

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.tsx
bagrub: updated as discussed using the tab(s) as an example

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.tsx
bagrub: updated

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -7,9 +7,46 @@ import { Component, Host, h, Prop } from '@stencil/core';
...
+
+  @Element() host: HTMLElement;
+
   /** Specifies the context of the component as it applies to the visual design and whether it inverts when light/dark mode is toggled. Default behavior is "light-inverts" and does not have to be specified. */
   @Prop({ reflect: true }) context: "light-inverts" | "light-always" | "dark-inverts" | "dark-always";
 
+
+
+  initNavItemset() {
+    // check for a default navItem, otherwise set the first one active
```
dgibson666, line 15: Unlike tabs, I don't think this is a good assumption - you can't just assume the first nav is active/current. It's on the app to make this work and make sense. Maybe make the story default the first item to current to show the styling and behavior.

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.scss
```diff
@@ -1,68 +1,79 @@
...
 }
 
 /* 
- * Dark Mode - display dark design based on mode or context.
- */
- 
+* Dark Mode - display dark design based on mode or context.
+*/
+
 [data-cbp-theme=light] cbp-nav-item[context*=dark]:not(#fakeId),
```
dgibson666: With a parent-child relationship, I think it makes more sense to look at the context of the parent. These children don't even need a context prop really. I think you can just reference the parent component here instead of the nav-item and have it work the same (toggling nav-item specific variables)

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -7,9 +7,46 @@ import { Component, Host, h, Prop } from '@stencil/core';
...
+    let activeNavItem = this.navItems[0];
+    this.navItems.forEach(navItem => {
+      if (navItem.selected === true) activeNavItem = navItem;
+    });
+    this.setActiveNav(activeNavItem);
+  }
+
+  setActiveNav(activatedNav) {
+    this.navItems.forEach((navItem: HTMLCbpNavItemElement) => {
+      let link = navItem.querySelector('a');
```
dgibson666: This function is accepting a node that is marked as selected, so you should handle buttons as well (regardless of other downstream integration concerns). If the dev set it to selected, it's selected. you can qs multiple elements easily enough: 

```
navItem.querySelector('a, button')
```

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.scss
```diff
@@ -1,68 +1,79 @@
...
-    --cbp-nav-item-button-color-bg-hover: var(--cbp-color-interactive-secondary-lighter);
-    --cbp-nav-item-button-color-bg-hover-dark: var(--cbp-color-interactive-secondary-darker);
-    
-    --cbp-nav-item-button-border-bottom-hover: var(--cbp-color-interactive-secondary-darker);
-    --cbp-nav-item-button-border-bottom-hover-dark: var(--cbp-color-interactive-secondary-lighter);
+  --cbp-nav-item-color-bg-hover: var(--cbp-color-interactive-secondary-lighter);
+  --cbp-nav-item-color-bg-hover-dark: var(--cbp-color-interactive-secondary-darker);
+  
+  --cbp-nav-item-border-bottom-hover: var(--cbp-color-interactive-secondary-darker);
+  --cbp-nav-item-border-bottom-hover-dark: var(--cbp-color-interactive-secondary-lighter);
```
dgibson666, line 35: I don't see any styles for the current/selected item.

dgibson666 (commented):
See code comments.

Also, I don't see aria-current being applied to the current element (Home) or any styling on it.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -7,9 +7,46 @@ import { Component, Host, h, Prop } from '@stencil/core';
...
+
   /** Specifies the context of the component as it applies to the visual design and whether it inverts when light/dark mode is toggled. Default behavior is "light-inverts" and does not have to be specified. */
   @Prop({ reflect: true }) context: "light-inverts" | "light-always" | "dark-inverts" | "dark-always";
 
+
+
+  initNavItemset() {
+    // check for a default navItem, otherwise set the first one active
+    let activeNavItem = this.navItems[0];
+    this.navItems.forEach(navItem => {
```
dgibson666: you can query selector this to be more efficient than looping over each.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: update var init to not assume selected state

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.scss
bagrub: updated selector to use the parent for context

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.scss
bagrub: updated styles to have the active state on the [selected] attribute

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
dgibson666: this still applies. this is a 1-liner if you this.host.querySelector('cbp-nav-item[selected]')

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -7,9 +7,41 @@ import { Component, Host, h, Prop } from '@stencil/core';
...
+  initNavItemset() {
+    // check for a default navItem, otherwise set the first one active
+    let activeNavItem;
+    this.navItems.forEach(navItem => {
+      if (navItem.selected === true) activeNavItem = navItem;
+    });
+    this.setActiveNav(activeNavItem);
+  }
+
+  setActiveNav(activatedNav) {
```
dgibson666, line 22: this should probably be called on componentDidLoad as well. I'm not seeing the aria-current on the home link.

dgibson666 (commented):
Couple issues still. 

Nice to see the URL is working for testing. If for some reason this doesn't work when deployed (different folder structure), we may need to write a function to compose the URLs in storybook. See: https://developer.mozilla.org/en-US/docs/Web/API/URL

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: updated to use query selector above

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: updated to call this in componentDidLoad

dgibson666 (approved):
looks good!

PR #230, opened 2024-12-10
Feature/combobox

dgibson666:
* Massive refactor of `cbp-dropdown` to:
  * Add keypress option cycling like a native select when full string search is not enabled.
  * Add combobox functionality with search string filtering (based on the `filter` prop).
  * Update the accessibility model to use `aria-activedescendant` rather than focus (tested in NVDA).
  * Update the dropdown item with a `current` prop to replace focus as a styling hook.
* Created States and Countries dropdown stories, added filter control.
* Created getInvertedContext function for tooltip use.


bagrub (approved):
LGTM

PR #234, opened 2024-12-13
Requirements PR

bagrub:
outline the requirements & documentation for the cbp-toggle component


dgibson666 (approved):
approved, but remember, you don't ever set booleans to true by default, and we want the status shown by default.

PR #235, opened 2024-12-19
Toggle component

bagrub:
Creation of the toggle component

### DIFF
--- a/packages/web-components/src/components/cbp-toggle/cbp-toggle.scss
+++ b/packages/web-components/src/components/cbp-toggle/cbp-toggle.scss
@@ -33,6 +33,8 @@
  * @Prop   --cbp-circle-inset: 0.375rem;
  * @Prop   --cbp-circle-border-width: 0.25rem;
  * @Prop   --cbp-toggle-gap: 1rem;  
+ * @Prop   --cbp-toggle-margin: 0 0 var(--cbp-space-4x) 0;  
+
  */
 
 :root{ 
@@ -73,13 +75,18 @@
   --cbp-toggle-circle-border-color-disabled: var(--cbp-color-interactive-disabled-light);
   --cbp-toggle-circle-border-color-disabled-dark: var(--cbp-color-interactive-disabled-dark);
 
+  --cbp-toggle-text-color: var(--cbp-color-text-darkest);
+  --cbp-toggle-text-color-dark: var(--cbp-color-text-lightest);
+
   --cbp-slider-width: 3.25rem;
   --cbp-slider-height: 1.75rem;
   --cbp-circle-diameter: 1rem;
   --cbp-circle-inset: 0.375rem;
   --cbp-circle-border-width: 0.25rem;
+  --cbp-status-text-width: var(--cbp-space-16x);
 
-  --cbp-toggle-gap: 1rem;  
+  --cbp-toggle-gap: var(--cbp-space-4x);
+  --cbp-toggle-margin: 0 0 var(--cbp-space-4x) 0;  
 }
 
 [data-cbp-theme=light] cbp-toggle[context*=dark]:not([context=light-always]),
@@ -102,16 +109,22 @@
   --cbp-toggle-background-color-disabled: var(--cbp-toggle-background-color-disabled-dark);
   --cbp-toggle-circle-color-disabled: var(--cbp-toggle-circle-color-disabled-dark);
   --cbp-toggle-circle-border-color-disabled: var(--cbp-toggle-circle-border-color-disabled-dark);
+  
+  --cbp-toggle-text-color: var(--cbp-toggle-text-color-dark);
 }
 
 cbp-toggle {
   display: flex;
   align-items: center;
   gap: var(--cbp-toggle-gap);
+  margin: var(--cbp-toggle-margin);
+  color: var(--cbp-toggle-text-color);
 
-  .slider{
+  input[type='checkbox']{
+    appearance: none;
     height: var(--cbp-slider-height);
     min-width: var(--cbp-slider-width);
+    flex-basis: var(--cbp-slider-width);
     border-radius: var(--cbp-border-radius-pill);
     background-color: var(--cbp-toggle-background-color);
     outline: var(--cbp-toggle-outline);
@@ -129,6 +142,13 @@ cbp-toggle {
     }
   }
 
+  span{
+    flex-basis: var(--cbp-status-text-width);
+    text-overflow: ellipsis;
+    overflow: hidden;
+    white-space: nowrap;
+  }
+
   /** 'On' States */
   &[checked] {
     --cbp-toggle-background-color: var(--cbp-color-interactive-selected-dark);
@@ -137,7 +157,7 @@ cbp-toggle {
     --cbp-toggle-circle-color-dark: var(--cbp-color-interactive-secondary-darker);
     --cbp-toggle-circle-border-color-dark: var(--cbp-color-interactive-secondary-darker);
 
-    & .slider:before{
+    & input[type='checkbox']:before{
       margin-left: calc(var(--cbp-slider-width) - var(--cbp-circle-diameter) - var(--cbp-circle-inset));
     }
 
@@ -161,7 +181,7 @@ cbp-toggle {
   /** 'Off' States */
   &:not([checked]) {
 
-    & .slider:before{
+    & input[type='checkbox']:before{
       margin-left: var(--cbp-circle-inset);
     }
 
@@ -196,13 +216,7 @@ cbp-toggle {
     --cbp-toggle-circle-border-color-dark: var(--cbp-toggle-circle-border-color-disabled-dark);
   }
 
-  //hide child checkbox control for visual
-  cbp-checkbox,
-  input[role="checkbox"]{
-    display: none;
-  }
-
-  //tech debt: override a browser default
+  // override a browser default
   &:focus-visible{
     outline: none;
   }
--- a/packages/web-components/src/components/cbp-toggle/cbp-toggle.scss
+++ b/packages/web-components/src/components/cbp-toggle/cbp-toggle.scss
@@ -32,8 +32,8 @@
  * @Prop   --cbp-toggle-color-circle-disabled-dark: var(--cbp-color-interactive-disabled-dark);
  * @Prop   --cbp-toggle-color-circle-border-disabled: var(--cbp-color-interactive-disabled-light);
  * @Prop   --cbp-toggle-color-circle-border-disabled-dark: var(--cbp-color-interactive-disabled-dark);
- * @Prop   --cbp-slider-width: 3.25rem;
- * @Prop   --cbp-slider-height: 1.75rem;
+ * @Prop   --cbp-toggle-control-width: 3.25rem;
+ * @Prop   --cbp-toggle-control-height: 1.75rem;
  * @Prop   --cbp-circle-diameter: 1rem;
  * @Prop   --cbp-circle-inset: 0.375rem;
  * @Prop   --cbp-circle-border-width: 0.25rem;
@@ -85,8 +85,8 @@
   --cbp-toggle-text-color: var(--cbp-color-text-darkest);
   --cbp-toggle-text-color-dark: var(--cbp-color-text-lightest);
 
-  --cbp-slider-width: 3.25rem;
-  --cbp-slider-height: 1.75rem;
+  --cbp-toggle-control-width: 3.25rem;
+  --cbp-toggle-control-height: 1.75rem;
   --cbp-circle-diameter: 1rem;
   --cbp-circle-inset: 0.375rem;
   --cbp-circle-border-width: 0.25rem;
@@ -123,15 +123,24 @@
 cbp-toggle {
   display: flex;
   align-items: center;
-  gap: var(--cbp-toggle-gap);
   margin: var(--cbp-toggle-margin);
   color: var(--cbp-toggle-text-color);
 
+  label{
+    // display: inline-flex;
+    // align-items: center;
+
+    display: grid;
+    grid-template-columns: 1fr var(--cbp-toggle-control-width) 1fr; 
+    gap: var(--cbp-toggle-gap);
+    align-items: center;
+  }
+
   input[type='checkbox']{
     appearance: none;
-    height: var(--cbp-slider-height);
-    min-width: var(--cbp-slider-width);
-    flex-basis: var(--cbp-slider-width);
+    height: var(--cbp-toggle-control-height);
+    min-width: var(--cbp-toggle-control-width);
+    flex-basis: var(--cbp-toggle-control-width);
     border-radius: var(--cbp-border-radius-pill);
     background-color: var(--cbp-toggle-color-bg);
     outline: var(--cbp-toggle-outline-width) var(--cbp-toggle-outline-style) var(--cbp-toggle-outline-color);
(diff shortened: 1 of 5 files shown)


packages/web-components/src/components/cbp-toggle/cbp-toggle.scss
```diff
@@ -1,17 +1,223 @@
 /**
- * @Prop   --cbp-toggle-XXX
+ * @Prop   --cbp-toggle-background-color: var(--cbp-color-interactive-secondary-light);
```
dgibson666: Your naming is not consistent with other components/standards. This should be `--cbp-toggle-color-bg`.

Almost all of these should be `--cbp-toggle-color-` something. I'm probably inconsistent with the "parts" being before color or after but since these are all small parts, not major structural pieces, I'd probably put them all after "color".

Outline and border should be broken up into their parts. Especially because dark mode only affects the color. The style and size do not need to be exposed for customization.

packages/web-components/src/components/cbp-toggle/cbp-toggle.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+    `;
+}
+
+export const MultipleToggle = MultipleTemplate.bind({});
+
+MultipleToggle.args={
+    ToggleItems: [
+        {
+            label: 'Toggle #1:',
+            statusTexton: '#1 On',
```
dgibson666: I thought we said get rid of these status text overrides and just use the main story controls, so they are all the same. It is the most likely use case.

packages/web-components/src/components/cbp-toggle/cbp-toggle.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+            />
+        </cbp-toggle>
+        `;
+    });
+    return html.join('');
+  }
+
+const MultipleTemplate = ({ToggleItems, labelWidth, hideStatus, disabled, context, sx }) => {
+    return `
+    <div>
```
dgibson666: Replace with `cbp-form-field` as discussed

packages/web-components/src/components/cbp-toggle/cbp-toggle.tsx
```diff
@@ -10,19 +10,19 @@ export class CbpToggle {
...
-  @Prop() disabled: boolean;
+  @Prop({ reflect: true }) disabled: boolean;
 
   /** Determines if the status text is visible for the render*/
-  @Prop() hideStatus: boolean = true;
+  @Prop({ reflect: true }) hideStatus: boolean;
 
   /** Determines the status text for the true toggle*/
-  @Prop() statusTextOn: string = 'on';
+  @Prop({ reflect: true }) statusTextOn: string = 'On';
```
dgibson666: this should not need to be reflected

packages/web-components/src/components/cbp-toggle/cbp-toggle.tsx
```diff
@@ -10,19 +10,19 @@ export class CbpToggle {
...
 
   /** Determines the status text for the false toggle*/
-  @Prop() statusTextOff: string = 'off';
+  @Prop({ reflect: true }) statusTextOff: string = 'Off';
```
dgibson666: this should not need to be reflected

packages/web-components/src/components/cbp-toggle/cbp-toggle.tsx
```diff
@@ -10,19 +10,19 @@ export class CbpToggle {
...
-  @Prop() checked: boolean;
+  @Prop({ reflect: true }) checked: boolean;
 
   /** Marks the toggle in a disabled state when specified. */
-  @Prop() disabled: boolean;
+  @Prop({ reflect: true }) disabled: boolean;
 
   /** Determines if the status text is visible for the render*/
-  @Prop() hideStatus: boolean = true;
+  @Prop({ reflect: true }) hideStatus: boolean;
```
dgibson666: this should not need to be reflected

packages/web-components/src/components/cbp-toggle/cbp-toggle.tsx
```diff
@@ -39,18 +39,64 @@ export class CbpToggle {
...
   }
 
   render() {
-    /**boilerplate HTML */
-    return (
-      <Host>
+    let ariaLabel = this.host.querySelector('label').textContent;
+    let ariaCheck = this.checked;
+
+    if(!this.hideStatus){
```
dgibson666: This should be doable as 1 return with a logical and controlling the status text output.

```
{!this.hideStatus && <span>{this.checked ? this.statusTextOn : this.statusTextOff}</span>}
```

packages/web-components/src/components/cbp-toggle/cbp-toggle.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+    </div>
+    `;
+}
+
+export const MultipleToggle = MultipleTemplate.bind({});
+
+MultipleToggle.args={
+    ToggleItems: [
+        {
+            label: 'Toggle #1:',
```
dgibson666: Remove colons from all labels. The mocks don't have them, and neither do any other form patterns.

dgibson666 (commented):
Are you sure you pushed all your changes? A few things we talked about are still missing.

packages/web-components/src/components/cbp-toggle/cbp-toggle.stories.tsx
bagrub: uploaded what you saw in discussion earlier, addresses this issue

packages/web-components/src/components/cbp-toggle/cbp-toggle.stories.tsx
bagrub: uploaded what you saw in discussion earlier, addresses this issue

packages/web-components/src/components/cbp-toggle/cbp-toggle.stories.tsx
bagrub: removed

packages/web-components/src/components/cbp-toggle/cbp-toggle.tsx
bagrub: updated as requested to consolidate

packages/web-components/src/components/cbp-toggle/cbp-toggle.tsx
bagrub: remove the reflected

packages/web-components/src/components/cbp-toggle/cbp-toggle.tsx
bagrub: remove the reflected

packages/web-components/src/components/cbp-toggle/cbp-toggle.scss
```diff
@@ -1,17 +1,230 @@
...
+ * @Prop   --cbp-toggle-color-bg-focus-dark: var(--cbp-color-interactive-focus-light);
+ * @Prop   --cbp-toggle-outline-focus: 2px solid var(--cbp-color-white);
+ * @Prop   --cbp-toggle-outline-focus-dark: 2px solid var(--cbp-color-black);
+ * @Prop   --cbp-toggle-color-bg-disabled: var(--cbp-color-interactive-disabled-dark);
+ * @Prop   --cbp-toggle-color-bg-disabled-dark: var(--cbp-color-interactive-disabled-light);
+ * @Prop   --cbp-toggle-color-circle-disabled: var(--cbp-color-interactive-disabled-light);
+ * @Prop   --cbp-toggle-color-circle-disabled-dark: var(--cbp-color-interactive-disabled-dark);
+ * @Prop   --cbp-toggle-color-circle-border-disabled: var(--cbp-color-interactive-disabled-light);
+ * @Prop   --cbp-toggle-color-circle-border-disabled-dark: var(--cbp-color-interactive-disabled-dark);
+ * @Prop   --cbp-slider-width: 3.25rem;
```
dgibson666: slider is a bad name for this. It needs to be namespaced with cbp-toggle. But since that's the full component, it probably makes sense to clarify exactly what you are applying the width to (not the full host element). I would call it "control", e.g.,  `--cbp-toggle-control`. All the others below need to keep the `--cbp-toggle` before whatever part they are describing.

packages/web-components/src/components/cbp-toggle/cbp-toggle.stories.tsx
```diff
@@ -0,0 +1,140 @@
...
+    return html.join('');
+  }
+
+const MultipleTemplate = ({ToggleItems, labelWidth, hideStatus, statusTexton, statusTextoff,  disabled, context, sx }) => {
+    //TODO: wrap in a form control w/ a label of Settings. use form fields as an example
+    return `
+    <cbp-form-field
+        label="Settings"
+        description="An example of multiple toggles in a form field"
+        group=""
```
dgibson666: you don't need the ="" on boolean props in HTML (even if it shows up like that in the DOM). It's just like "checked".

packages/web-components/src/components/cbp-toggle/cbp-toggle.tsx
```diff
@@ -39,18 +39,46 @@ export class CbpToggle {
...
-    return (
-      <Host>
-        <slot></slot>
-      </Host>
-    );
+    let ariaLabel = this.host.querySelector('label').textContent;
+    let ariaCheck = this.checked;
+
+      return (
+        <Host
```
dgibson666: Now that you're using a real checkbox as the control, this should not have role=checkbox or tabindex (I notice 2 tab stops and 1 had not visible focus indicator). You shouldn't need the click handler because everything should be rendered in a label like a checkbox. See my latest checkbox updates for generating and applying the fieldId (also allows it to be specified on the slotted control instead). Adding a label means you'll need to move the flex context to it instead. You also shouldn't need the named slot for the label. (I've done some testing and flex seems to work on plain text nodes as if it was an element - if it gives you trouble, let's screen share and discuss).

dgibson666 (commented):
See code comments. looks pretty good, but a few issues remain.

packages/web-components/src/components/cbp-toggle/cbp-toggle.scss
bagrub: updated instances of --cbp-slider to --cbp-toggle-control

packages/web-components/src/components/cbp-toggle/cbp-toggle.stories.tsx
bagrub: removed

packages/web-components/src/components/cbp-toggle/cbp-toggle.tsx
bagrub: Updated as discussed

packages/web-components/src/components/cbp-toggle/cbp-toggle.scss
```diff
@@ -85,8 +85,8 @@
   --cbp-toggle-text-color: var(--cbp-color-text-darkest);
   --cbp-toggle-text-color-dark: var(--cbp-color-text-lightest);
 
-  --cbp-slider-width: 3.25rem;
-  --cbp-slider-height: 1.75rem;
+  --cbp-toggle-control-width: 3.25rem;
+  --cbp-toggle-control-height: 1.75rem;
   --cbp-circle-diameter: 1rem;
```
dgibson666: these last 3 should still use the component namespace.

packages/web-components/src/components/cbp-toggle/cbp-toggle.stories.tsx
```diff
@@ -81,17 +79,16 @@ function generateToggles(items, labelWidth, hideStatus, statusTexton, statusText
         ${context && context != 'light-inverts' ? `context=${context}` : ''}
         ${sx ? `sx=${JSON.stringify(sx)}` : ``}
         >
-            <label
-                slot='toggleLabel'
+            <span
                 style='width: ${labelWidth}'
```
dgibson666: I'm not a fan of this implementation. I would make it a prop on the component and use it to set a CSS variable in sx that gets used in the component CSS. The span should not be needed then.

packages/web-components/src/components/cbp-toggle/cbp-toggle.stories.tsx
```diff
@@ -81,17 +79,16 @@ function generateToggles(items, labelWidth, hideStatus, statusTexton, statusText
...
-            </label>
-            <input
-                type="checkbox"
-                name="checkbox"
-                value="1"
-            />
+           </span>
+                <input
+                    type="checkbox"
+                    name="checkbox"
```
dgibson666: you should probably pass the name and value in the object and iterate one of them. A group of checkboxes with the same name and value would not be useful in real life.

packages/web-components/src/components/cbp-toggle/cbp-toggle.stories.tsx
```diff
@@ -0,0 +1,137 @@
...
+    argTypes: {
+        hideStatus: {
+            description: 'Determines if the status text for the `on` and `off` is visible for the toggle control',
+            control: 'boolean'
+        },
+        disabled: {
+            description: 'Sets the disable state for the toggle control',
+            control: 'boolean'
+        },    
+        statusTexton: {
```
dgibson666: Sorry, this is nitpicky, but please make the On/Off control names camelcase to be consistent with the prop names.

packages/web-components/src/components/cbp-toggle/cbp-toggle.tsx
```diff
@@ -39,18 +39,46 @@ export class CbpToggle {
...
+  handleChange() {
+    this.toggleEvent();
+  }
+
+  @Event() toggleClick: EventEmitter;
+
   /** Event: toggles the control true/false & updates DOM accordingly*/
   toggleEvent(){
-    //TODO: logic here
+    let checkbox = this.host.querySelector('input[type="checkbox"]');
```
dgibson666: This should be a component-scoped variable that you get on componentDidLoad. The disabled prop needs to push the down the attribute on the native input. See checkbox for an example. (this will set you up for implementing the ID part later as well).

It looks like I assign the event listener that way as well (rather then using @Listen) and invoke the handler directly. Checkbox also emits a custom event, which toggle should also do to be consistent.

dgibson666 (commented):
See code comments. Mostly story/implementation details.

dgibson666 (approved):
looks good.

packages/web-components/src/components/cbp-toggle/cbp-toggle.scss
bagrub: updated as requested

packages/web-components/src/components/cbp-toggle/cbp-toggle.stories.tsx
bagrub: update as discussed in call

packages/web-components/src/components/cbp-toggle/cbp-toggle.stories.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-toggle/cbp-toggle.stories.tsx
bagrub: update as requested

packages/web-components/src/components/cbp-toggle/cbp-toggle.tsx
bagrub: updated as discussed

PR #238, opened 2025-01-13
Remove Roboto fonts from `cbp-app`

dgibson666:
* Remove Roboto fonts from `cbp-app`
* Include them in Storybook as external dependencies


bagrub (approved):
LGTM

PR #241, opened 2025-01-15
Loader component

bagrub:
Initial creation of the Loader component


packages/web-components/src/components/cbp-loader/cbp-loader.scss
```diff
@@ -1,3 +1,245 @@
...
+[data-cbp-theme=dark] cbp-loader:not([context=dark-inverts]):not([context=light-always]) {
+
+  --cbp-loader-color: var(--cbp-loader-color-dark);
+  --cbp-loader-center-color: var(--cbp-loader-center-color-dark);
+  --cbp-loader-track-color: var(--cbp-loader-track-color-dark);
+  
+  --cbp-loader-status-description-color: var(--cbp-loader-status-description-color-dark);
+  --cbp-loader-status-indicator-color: var(--cbp-loader-status-indicator-color-dark);}
+
+@keyframes prixClipFix {
```
dgibson666: what does "prixClipFix" even mean? Maybe rename it something that makes sense?

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -7,12 +8,115 @@ import { Component, Host, h } from '@stencil/core';
...
+    if (typeof this.sx == 'string') {
+      this.sx = JSON.parse(this.sx) || {};
+    }
+    setCSSProps(this.host, {
+      ...this.sx,
+    });
+  }
+
+  componentDidLoad() {
+    if(this.value < this.min){ //TODO: not changing color, i think the prop update isn't getting to the html for the css?
```
dgibson666: Props are by default immutable internally, so if you intend to change a prop value internally, you need to specify:

`@Prop({mutable: true}) error: boolean;`

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -7,12 +8,115 @@ import { Component, Host, h } from '@stencil/core';
...
+      statusIndicator =  Math.round((this.value / this.max) * 100) + "%"
+    }
+
+
     return (
       <Host>
-        <slot></slot>
+        {this.determinate && this.variant == 'linear' && 
+          
+            <span class='cbpLoaderDesc'>{statusDescription}
```
dgibson666: Don't use caps in CSS. Namespacing should be cbp-loader-desc.

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -7,12 +8,115 @@ import { Component, Host, h } from '@stencil/core';
...
+          </span>  
+        }
+        {this.determinate && this.variant == 'circular' && this.size == 'small' && this.color != 'progress'
+        ?
+          statusIndicator
+        :
+          <progress
+            value={this.value}
+            max={this.max}
+            aria-busy={this.value < this.max ? 'true' : 'false'}
```
dgibson666: Reviewing the docs, I don't think that aria-busy is supposed to be on the progress bar itself. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -7,12 +8,115 @@ import { Component, Host, h } from '@stencil/core';
...
+          <span class='cbpLoaderDesc'>
+            {statusIndicator}
+          </span>  
+        }
+        {this.determinate && this.variant == 'circular' && this.size == 'small' && this.color != 'progress'
+        ?
+          statusIndicator
+        :
+          <progress
+            value={this.value}
```
dgibson666: For indeterminate progress bars, there cannot be a value, so this should be enforced logically here. Based on the docs, a value makes it not indeterminate. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -7,12 +8,115 @@ import { Component, Host, h } from '@stencil/core';
...
+            max={this.max}
+            aria-busy={this.value < this.max ? 'true' : 'false'}
+            aria-valuenow={this.value}
```
dgibson666: You don't need the aria-attributes because you're using a native progress element (not role="progressbar".

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -7,12 +8,115 @@ import { Component, Host, h } from '@stencil/core';
...
+  @Prop({ reflect: true }) determinate: boolean = false;
+
+  /** Used in deternminate mode to display the current value of loaded content*/
+  @Prop() value: number = 0;
+
+  /** Used in deternminate mode to display the max value of loaded content*/
+  @Prop() max: number = 100;
+
+  /** Used in deternminate mode to display the min value of loaded content*/
+  @Prop() min: number = 0;
```
dgibson666: I guess you don't need "min" because it's implicitly 0 in a progress bar: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -7,12 +8,115 @@ import { Component, Host, h } from '@stencil/core';
 
 export class CbpLoader {
 
+  @Element() host: HTMLElement;
+  
+  /** Defines if the loader will be in displayed as a circular or linear variant*/
+  @Prop({ reflect: true }) variant: "circular" | "linear";
+
+  /** Defines the color of the loader render, used to express the state of the loader */
+  @Prop() color: "progress" | "success" | "error" = 'progress';
```
dgibson666: remove colors per discussion - it only changes color based on success or error states (booleans).

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -7,12 +8,115 @@ import { Component, Host, h } from '@stencil/core';
...
+
+  /** Defines the size of the loader render, default value of large */
+  @Prop() size: "large" | "small" = "large";
```
dgibson666: This needs to be reflected. or you can't use it as a CSS hook.

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -7,12 +8,115 @@ import { Component, Host, h } from '@stencil/core';
...
+  @Prop() value: number = 0;
+
+  /** Used in deternminate mode to display the max value of loaded content*/
+  @Prop() max: number = 100;
+
+  /** Used in deternminate mode to display the min value of loaded content*/
+  @Prop() min: number = 0;
+
+  /** Used to set the loader to the 'success' state of the loader */
+  @Prop() success: boolean;
```
dgibson666: needs to be reflected (and possibly mutable?).

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -7,12 +8,115 @@ import { Component, Host, h } from '@stencil/core';
...
+
+  /** Used to set the loader to the 'error' state of the loader */
+  @Prop() error: boolean;
```
dgibson666: needs to be reflected

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -7,12 +8,115 @@ import { Component, Host, h } from '@stencil/core';
...
+    let statusIndicator;
+
+    if (this.color == 'success'){
+      statusDescription = <span>Success</span>;
+      statusIndicator = <cbp-icon class="statusIndicator" name="check-circle" color='var(--cbp-loader-status-indicator-color)'></cbp-icon>
+    }else if(this.color =='error'){
+      statusDescription = <span>Error</span>;
+      statusIndicator = <cbp-icon class="statusIndicator" name="exclamation-circle" color='var(--cbp-loader-status-indicator-color)'></cbp-icon>
+    }else{
+      statusDescription = <span>Uploading...</span>;
```
dgibson666: This needs to be customizable. TIL you can use a label with a progress element, so that probably makes the most sense. You can hide it when it's not visible and it should still provide an accessible label.

packages/web-components/src/components/cbp-loader/cbp-loader.scss
```diff
@@ -1,3 +1,245 @@
...
+      
+      //hide the visuals for the value/max relationship for the cirular loader. 
+      &::-webkit-progress-bar{
+        display: none;
+      }
+
+      &:before{
+        content: "";
+        position: absolute;
+        z-index: 100;
```
dgibson666: should use a token. This is an overlay at the page/canvas level (not like a dialog), so I don't think it needs to be too large.

packages/web-components/src/components/cbp-loader/cbp-loader.scss
```diff
@@ -1,3 +1,245 @@
...
+        border-radius: calc(var(--cbp-border-radius-circle) / 2);
+        margin: auto auto;
+        background: var(--cbp-loader-color);
+
+      }
+
+      &:after{
+        content: "";
+        position: absolute;
+        z-index: 100;
```
dgibson666: should use a token. This is an overlay at the page/canvas level (not like a dialog), so I don't think it needs to be too large.

packages/web-components/src/components/cbp-loader/cbp-loader.scss
```diff
@@ -1,3 +1,245 @@
...
+        text-align: center;
+        z-index: 101;
+        color: var(--cbp-loader-color);
+        font-weight: var(--cbp-font-weight-bold);
+      }
+    }
+  }
+
+/** Color variations for Loader */
+&[success]:not([error]){
```
dgibson666: I would put success after error and you shouldn't have to use the :not(). Both success and error should never be true, but that way success overrides the error.

dgibson666 (commented):
See code comments. I found a few discrepancies with the native progress element after reviewing MDN docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress

packages/web-components/src/components/cbp-loader/cbp-loader.scss
bagrub: prixClipFix was just the name from the example i based the original animation on, renamed this to spin

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: update to be mutable

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: moved the aria-busy to the host tag instead of the progress tag

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: updated class name as requested

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: logically enforced the to only be populated when determinate is true

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: removed aria attributes

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: removed the min prop

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: removed color prop

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: update to reflect prop

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: updated prop to be reflected. with current logic mutable wasn't necessary

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: updated prop to reflected

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: updated statusDescription to be a slot within a label & added logic for the error/ success state text changes as per design

packages/web-components/src/components/cbp-loader/cbp-loader.scss
bagrub: updated to use the defined cssvar tokens for z-index

packages/web-components/src/components/cbp-loader/cbp-loader.scss
bagrub: updated to use the defined cssvar tokens for z-index

packages/web-components/src/components/cbp-loader/cbp-loader.scss
bagrub: Removed the :not() on the selector here. Might have been typo in comment but original order for success then error seems to be working as expected (error overrides success) will update if that is not correct

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -10,14 +10,14 @@ export class CbpLoader {
 
   @Element() host: HTMLElement;
   
+  /** Specifies a unique `ID` for the loader, used to wire up the controls and accessibility features. */
+  @Prop() uid: string = createNamespaceKey('cbp-accordion-item');
```
dgibson666: make this 'cbp-loader' as the prefix and name it "progressId" to be consistent with other IDs. If you just call it uid, it's not clear what it belongs to. (it's on the wrong thing anyway).

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -28,14 +28,11 @@ export class CbpLoader {
   /** Used in deternminate mode to display the max value of loaded content*/
   @Prop() max: number = 100;
 
-  /** Used in deternminate mode to display the min value of loaded content*/
-  @Prop() min: number = 0;
-
   /** Used to set the loader to the 'success' state of the loader */
-  @Prop() success: boolean;
+  @Prop({mutable: true}) success: boolean;
```
dgibson666: needs to be reflected too since you're using it as a CSS hook.

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -28,14 +28,11 @@ export class CbpLoader {
...
 
   /** Used to set the loader to the 'error' state of the loader */
-  @Prop() error: boolean;
+  @Prop({mutable: true}) error: boolean;
```
dgibson666: needs to be reflected too since you're using it as a CSS hook.

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -53,66 +50,63 @@ export class CbpLoader {
...
     
   }
  
   render() {
-    let statusDescription = '';
     let statusIndicator;
 
-    if (this.color == 'success'){
-      statusDescription = <span>Success</span>;
+    if(this.success && !this.error){
```
dgibson666: you don't need a !this.error in an if else... you're making the error take precedence over success then.

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -53,66 +50,63 @@ export class CbpLoader {
...
     }else{
-      statusDescription = <span>Uploading...</span>;
       statusIndicator =  Math.round((this.value / this.max) * 100) + "%"
     }
-
-
+  
     return (
-      <Host>
+      <Host
```
dgibson666: The host is not a semantic element, so the ID doesn't do any good here. The ID should be on the actual progress element. aria-busy can be removed.

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -53,66 +50,63 @@ export class CbpLoader {
...
+          <span class='cbp-loader-desc'>
             {statusIndicator}
           </span>  
         }
-        {this.determinate && this.variant == 'circular' && this.size == 'small' && this.color != 'progress'
+        {this.determinate && this.variant == 'circular' && this.size == 'small' && (this.success || this.error)
         ?
           statusIndicator
         :
           <progress
```
dgibson666: the progress element needs the ID, which is referenced by the label.

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -53,66 +50,63 @@ export class CbpLoader {
...
+        aria-busy={this.determinate ?
+           (this.value < this.max ? 'true' : 'false') 
+          : (!this.success ? 'true' : 'false')
+        }
+      >
         {this.determinate && this.variant == 'linear' && 
           
-            <span class='cbpLoaderDesc'>{statusDescription}
-            
+            <label class='cbp-loader-desc'>
```
dgibson666: You might not need the class now that this is a label tag. The label needs to reference the progress' id with `htmlFor={this.progressId}

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -53,66 +50,63 @@ export class CbpLoader {
...
+              {(this.success && !this.error) ?
+                `Complete`
+              :( this.error ?
+                  `Error`
+                :
+                null
+              ) 
+              }
+              <slot name='cbp-loader-desc' />
```
dgibson666: This might not need to be a named slot if you don't have a default slot anywhere... I feel like the app could provide the success and error text as well but we can hold that for a later discussion.

dgibson666 (commented):
Pretty close. see code comments

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: updated to also reflect

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: updated to also reflect

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: removed the !this.error in the logic & update css to mirror logic as discussed

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: removed the aria-busy & attached the id to the progress tag

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: added ID and updated label to have a labelled-by with same id

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: removed class from this case & update styles to use the label tag in its place

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: removed the slot name, holding on the text update as discussed in call

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -69,26 +68,22 @@ export class CbpLoader {
...
-          : (!this.success ? 'true' : 'false')
-        }
-      >
+      <Host >
         {this.determinate && this.variant == 'linear' && 
           
-            <label class='cbp-loader-desc'>
-              {(this.success && !this.error) ?
+            <label
+              aria-labelledby={this.progressid}
```
dgibson666: this should be `htmlFor`. The label is labelling the progress bar. Not the other way around. (and capital "I").

packages/web-components/src/components/cbp-loader/cbp-loader.stories.tsx
```diff
@@ -64,7 +64,7 @@ const Template = ({uid, variant, label, size, determinate, value, max, success,
             ${sx ? `sx=${JSON.stringify(sx)}` : ``}
             >
             ${variant == 'linear' && label && !(success || error)? 
-            `<label slot='cbp-loader-desc'> ${label}</label>`
+            `<label> ${label}</label>`
```
dgibson666: You're rendering double/nested label tags now. Remove it from the story (leave the text).

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -11,7 +11,7 @@ export class CbpLoader {
   @Element() host: HTMLElement;
   
   /** Specifies a unique `ID` for the loader, used to wire up the controls and accessibility features. */
-  @Prop() uid: string = createNamespaceKey('cbp-accordion-item');
+  @Prop() progressid: string = createNamespaceKey('cbp-loader');
```
dgibson666: this should be capital "I" to be consistent.

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -104,7 +99,8 @@ export class CbpLoader {
         ?
           statusIndicator
         :
-          <progress
+          <progress      
+            id={this.progressid}
```
dgibson666: this should be capital "I" as well.

dgibson666 (commented):
Couple remaining comments. Also, please add API docs.

packages/web-components/src/components/cbp-loader/cbp-loader.stories.tsx
bagrub: removed label tag from in storybook

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: updated to progressId

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: updated to progressId

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: updated to htmlFor

dgibson666 (approved):
looks good for first cut. May need additional iterations, as discussed.

PR #244, opened 2025-02-03
Loader update

bagrub:
update loader circular determinate variant to include slot for descriptive text

### DIFF
--- a/packages/web-components/src/components/cbp-loader/cbp-loader.stories.tsx
+++ b/packages/web-components/src/components/cbp-loader/cbp-loader.stories.tsx
@@ -29,6 +29,11 @@ export default {
             control: 'text',
             if: { arg: 'determinate', eq: true}
         },
+        orientation: {
+            control: 'select',
+            options: ['right', 'bottom'],
+            if: {arg: 'determinate', eq: true}
+        },
         success : {
             control: 'boolean'
         },
@@ -49,7 +54,7 @@ export default {
     }
   };
 
-const Template = ({progressid, variant, label, size, determinate, value, max, success, error, context, sx }) => {
+const Template = ({progressid, variant, label, size, determinate, value, max, orientation, success, error, context, sx }) => {
     return ` 
         <cbp-loader
             ${progressid ? `progressid=${progressid}` : ''}
@@ -58,6 +63,7 @@ const Template = ({progressid, variant, label, size, determinate, value, max, su
             ${determinate ? 'determinate' : ''}
             ${value ? `value=${value}` : ``}
             ${max ? `max=${max}` : ``}
+            ${orientation ? `orientation=${orientation}` : ``}
             ${success ? 'success' : ''}
             ${error ? 'error' : ''}
             ${context && context != 'light-inverts' ? `context=${context}` : ''}
@@ -67,10 +73,11 @@ const Template = ({progressid, variant, label, size, determinate, value, max, su
             `${label}`
             : ``
             }
+            ${variant == 'circular' && size == 'large' && label && !(success || error)? 
+            `${label}` : `` }
         </cbp-loader>
     `;
 };
-  
 export const Loader = Template.bind({});
   
 Loader.args = {
--- a/packages/web-components/src/components/cbp-loader/cbp-loader.stories.tsx
+++ b/packages/web-components/src/components/cbp-loader/cbp-loader.stories.tsx
@@ -26,8 +26,14 @@ export default {
             if: { arg: 'determinate', eq: true },
         },
         label : {
+            name: 'label (slotted)',
             control: 'text',
             if: { arg: 'determinate', eq: true}
+        },
+        orientation: {
+            control: 'select',
+            options: ['horizontal', 'vertical'],
+            if: {arg: 'variant', eq: 'circular'}
         },
         success : {
             control: 'boolean'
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/cbp-loader/cbp-loader.stories.tsx
```diff
@@ -29,6 +29,11 @@ export default {
             control: 'text',
             if: { arg: 'determinate', eq: true}
         },
+        orientation: {
+            control: 'select',
+            options: ['right', 'bottom'],
+            if: {arg: 'determinate', eq: true}
```
dgibson666: This doesn't seem to have any effect on linear progress bars, so shouldn't the if statement target that?

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -28,6 +28,9 @@ export class CbpLoader {
   /** Used in deternminate mode to display the max value of loaded content*/
   @Prop() max: number = 100;
 
+  /** Used to set the text orientation for the circular determinate loader's description*/
+  @Prop() orientation: "right" | "bottom" = 'right';
```
dgibson666: I don't feel like these values make sense in the context of the name. I would use "horizontal" and "vertical" instead of right and bottom (those are more positions, but you couldn't rename the prop without hyphenating it to "labelPosition" or something. This prop should be reflected as well to use as a CSS selector.

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -53,7 +56,15 @@ export class CbpLoader {
     if(this.determinate && this.variant == 'circular'){
       this.host.style.setProperty("--cbp-loader-circular-determinate", `conic-gradient(var(--cbp-loader-color) ${((this.value / this.max) * 100) * 3.6}deg, var(--cbp-loader-track-color) 0deg)`)
     }
-    
+
+    console.log(this.orientation);
```
dgibson666: remove the console.log

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -53,7 +56,15 @@ export class CbpLoader {
...
+    if(this.orientation == 'bottom'){
```
dgibson666: I'm not a fan of this implementation. These should be set in CSS based off the attribute selectors.

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -66,9 +77,9 @@ export class CbpLoader {
     }else{
       statusIndicator =  Math.round((this.value / this.max) * 100) + "%"
     }
-  
+
     return (
-      <Host >
+      <Host>
```
dgibson666, line 73: I feel like this render logic could be significantly simplified. It's looking pretty convoluted right now. In which cases is an actual progress bar rendered? In which cases is a label rendered? (I still think indeterminate loaders should have the option for label text, but it seems to be tied to determinate only). I'd like to talk through this at least before deciding whether to defer it.

packages/web-components/src/components/cbp-loader/cbp-loader.stories.tsx
```diff
@@ -29,6 +29,11 @@ export default {
             control: 'text',
```
dgibson666, line 30: If a control represents slotted text, use a name key to mark it "label (slotted)" for clarity. That's how I've done it in my stories. Otherwise devs get lazy and may assume they are all props.

dgibson666 (commented):
I have a few code comments and would like to review your render logic to see if refactoring it is simple enough to do now or should be deferred.

packages/web-components/src/components/cbp-loader/cbp-loader.stories.tsx
bagrub: update to logic to be based on circular variant as discussed. known issue with storybook if{} keeps us from doing more complex logic here. issue thread: https://github.com/storybookjs/storybook/discussions/18542

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: Updated as requested

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: removed

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: removed this logic & used the attribute selectors from orientation to do the styling

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: re did logic of render & discussed changes on call. Also added discussed changes for visibility of loader on the determinate small circular error/success loader instead of not rendering it as discussed

packages/web-components/src/components/cbp-loader/cbp-loader.stories.tsx
bagrub: added

dgibson666 (approved):
Approving, but I think we need to discuss some things further and iterate again. Things to discuss:

* Having a label (optional) for all variants
* Whether the complete/error text should be hardcoded or app-provided
* Whether the value needs to be set to 100 for indeterminate on completion (needs screen reader testing)
* What happens with a long label in vertical orientation (should it be centered and wrap?)

PR #246, opened 2025-02-07
Feature/subnav

bagrub:

### DIFF
--- a/packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
+++ b/packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
@@ -1,0 +1,3 @@
+:host {
+  display: block;
+}
--- a/packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
+++ b/packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
@@ -15,7 +15,7 @@
   --cbp-subnav-item-color: var(--cbp-color-interactive-secondary-darker);
   // --cbp-subnav-item-color-dark: var(--cbp-color-interactive-secondary-darker);
   --cbp-subnav-item-color-bg: transparent;
-  // --cbp-subnav-item-color-bg-dark: var();
+  // --cbp-subnav-item-color-bg-dark: var(--cbp-color-gray-cool-70);
 
   --cbp-subnav-item-current-color:var(--cbp-color-interactive-active-dark);
   // --cbp-subnav-item-current-color-dark: var(--cbp-color-interactive-secondary-light);
@@ -82,17 +82,29 @@ cbp-subnav-item {
     --cbp-button-border-width: var(--cbp-border-size-sm);
     
     --cbp-button-color: var(--cbp-subnav-item-color);
+    --cbp-button-color-dark: var(--cbp-subnav-item-color);
     --cbp-button-color-hover: var(--cbp-color-interactive-secondary-darker);
+    --cbp-button-color-hover-dark: var(--cbp-color-interactive-secondary-darker);
     --cbp-button-color-focus: var(--cbp-color-text-lightest);
+    --cbp-button-color-focus-dark: var(--cbp-color-text-lightest);
     --cbp-button-color-active: var(--cbp-color-text-lightest);
+    --cbp-button-color-active-dark: var(--cbp-color-text-lightest);
     --cbp-button-color-bg: var(--cbp-subnav-item-color-bg);
+    --cbp-button-color-bg-dark: var(--cbp-color-white);
     --cbp-button-color-bg-hover: var(--cbp-color-interactive-secondary-lighter);
+    --cbp-button-color-bg-hover-dark: var(--cbp-color-interactive-secondary-lighter);
     --cbp-button-color-bg-focus: var(--cbp-color-interactive-focus-dark);
+    --cbp-button-color-bg-focus-dark: var(--cbp-color-interactive-focus-dark);
     --cbp-button-color-bg-active: var(--cbp-color-interactive-active-dark);
+    --cbp-button-color-bg-active-dark: var(--cbp-color-interactive-active-dark);
     --cbp-button-color-border: var(--cbp-color-gray-cool-40);
+    --cbp-button-color-border-dark: var(--cbp-color-gray-cool-40);
     --cbp-button-color-border-hover: var(--cbp-color-gray-cool-40);
+    --cbp-button-color-border-hover-dark: var(--cbp-color-gray-cool-40);
     --cbp-button-color-border-focus: var(--cbp-color-gray-cool-40);
+    --cbp-button-color-border-focus-dark: var(--cbp-color-gray-cool-40);
     --cbp-button-color-border-active: var(--cbp-color-gray-cool-40);
+    --cbp-button-color-border-active-dark: var(--cbp-color-gray-cool-40);
 
     /**Tech Debt: dark mode not outlined in spec so currently mapped to primary outline button values, associated vars to override below w/ original cbp-button values
     --cbp-button-color-dark: var(--cbp-color-text-darkest);
--- a/packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
+++ b/packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
@@ -1,3 +1,108 @@
-:host {
-  display: block;
+/** 
+ * @prop  --cbp-subnav-item-active-border-color: var(--cbp-color-interactive-active-dark);
+ * @prop  --cbp-subnav-item-active-color:var(--cbp-color-interactive-active-dark);
+ * @prop  --cbp-subnav-item-active-color-bg: var(--cbp-color-interactive-active-light);
+ */
+
+:root {
+  /* 
+    Component spec does not have dark mode, commented out vars & code are guess values incase they are needed at later date
+  */
+  --cbp-subnav-item-active-border-color: var(--cbp-color-interactive-active-dark);
+  --cbp-subnav-item-active-color:var(--cbp-color-interactive-active-dark);
+  // --cbp-subnav-item-active-color-dark: var(--cbp-color-interactive-secondary-light);
+  --cbp-subnav-item-active-color-bg: var(--cbp-color-interactive-active-light);
+  // --cbp-subnav-item-active-color-bg-dark: var(--cbp-color-interactive-active-dark);
+  // --cbp-subnav-item-active-border-color-dark: var(--cbp-color-interactive-active-light); 
+
+  --cbp-subnav-item-indent: var(--cbp-space-4x);
 }
+
+// [data-cbp-theme=light] cbp-subnav-item[context*=dark]:not([context=light-always]),
+// [data-cbp-theme=dark] cbp-subnav-item:not([context=dark-inverts]):not([context=light-always]) {
+
+  // --cbp-subnav-item-active-border-color: var(--cbp-subnav-item-active-border-color-dark);
+// }
+
+cbp-subnav-item {
+  display: flex;
+
+  --cbp-subnav-item-rotate-icon: rotate(90deg);
+  &[open]{
+    --cbp-subnav-item-rotate-icon: rotate(270deg);
+  }
+
+  &[current] cbp-button:first-of-type,
+  &:has( + * cbp-subnav-item[current]) cbp-button:first-of-type{
+    border-left: var(--cbp-border-size-xl) solid var(--cbp-subnav-item-active-border-color);
+  }
+
+  &[current] cbp-button[fill=outline][color=primary]:first-of-type{
+    --cbp-button-color: var(--cbp-subnav-item-active-color);
+    background-color: var(--cbp-subnav-item-active-color-bg);
+    font-style: italic;
+  }
+
+  //Overrides for button variables
+  cbp-button[fill=outline][color=primary]{
+    --cbp-button-min-height: 3.5rem;
+    --cbp-button-width: 100%;
+    --cbp-button-border-radius: var(--cbp-border-radius-sharp);
+    --cbp-button-border-width: var(--cbp-border-size-sm);
+    
+    --cbp-button-color: var(--cbp-color-interactive-secondary-darker);
+    --cbp-button-color-hover: var(--cbp-color-interactive-secondary-darker);
+    --cbp-button-color-focus: var(--cbp-color-text-lightest);
+    --cbp-button-color-active: var(--cbp-color-text-lightest);
+    --cbp-button-color-bg: transparent;
+    --cbp-button-color-bg-hover: var(--cbp-color-interactive-secondary-lighter);
+    --cbp-button-color-bg-focus: var(--cbp-color-interactive-focus-dark);
+    --cbp-button-color-bg-active: var(--cbp-color-interactive-active-dark);
+    --cbp-button-color-border: var(--cbp-color-gray-cool-40);
+    --cbp-button-color-border-hover: var(--cbp-color-gray-cool-40);
+    --cbp-button-color-border-focus: var(--cbp-color-gray-cool-40);
+    --cbp-button-color-border-active: var(--cbp-color-gray-cool-40);
+
+    /**Tech Debt: dark mode not outlined in spec so currently mapped to primary outline button values, associated vars to override below w/ original cbp-button values
+    --cbp-button-color-dark: var(--cbp-color-text-darkest);
+    --cbp-button-color-hover-dark: var(--cbp-color-text-darkest);
+    --cbp-button-color-focus-dark: var(--cbp-color-text-darkest);
+    --cbp-button-color-active-dark: var(--cbp-color-text-darkest);
+    --cbp-button-color-bg-dark: var(--cbp-color-interactive-primary-base);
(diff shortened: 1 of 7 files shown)


packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
```diff
@@ -0,0 +1,141 @@
...
+ * @prop  --cbp-subnav-item-current-border-color: var(--cbp-color-interactive-active-dark);
+ * @prop  --cbp-subnav-item-current-color:var(--cbp-color-interactive-active-dark);
+ * @prop  --cbp-subnav-item-current-color-bg: var(--cbp-color-interactive-active-light);
+
+ * @prop  --cbp-subnav-item-indent: var(--cbp-space-4x);
+ * @prop  --cbp-subnav-item-font-size: var(--cbp-font-size-heading-md);
+ */
+
+:root {
+  /*
```
dgibson666, line 11: We need to define the text and bg colors. Right now, the Storybook demo is in-page and looks awful on my PC which is in dark mode. I know it's supposed to be in a drawer, but we can't have the storybook demo looking broken.

packages/web-components/src/components/cbp-subnav/cbp-subnav.stories.tsx
```diff
@@ -0,0 +1,135 @@
...
+        <cbp-subnav
+            accessibilitytext="${accessibilitytext}"
+        >
+            ${generateContent(items)}
+        </cbp-subnav>
+    `;
+};
+  
+export const Subnav = SubnavTemplate.bind({});
+
```
dgibson666, line 39: set the Subnav.storyName = "Sub-Navigation"

packages/web-components/src/components/cbp-subnav/cbp-subnav.stories.tsx
```diff
@@ -0,0 +1,135 @@
+export default {
+    title: 'Components/Subnav',
```
dgibson666: Use "Sub-Navigation" and update else where to match. (API docs, requirements)

packages/web-components/src/components/cbp-subnav/cbp-subnav.tsx
```diff
@@ -0,0 +1,56 @@
...
+  styleUrl: 'cbp-subnav.scss'
+})
+export class CbpSubNav {
+
+  private subNavItems: HTMLCbpSubnavItemElement[] = [];
+
+  @Element() host: HTMLElement;
+
+  /**  Sets the aria-label for the navigation element of the subnav*/
+  @Prop() accessibilitytext: string;
```
dgibson666: This is basically required because it's a second nav element. Maybe we should default it to "Sub-Navigation"? They can still override it if needed.

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
```diff
@@ -0,0 +1,141 @@
...
+      transform: var(--cbp-subnav-item-rotate-icon);
+
+    }
+  }
+
+  &[open]{
+    --cbp-subnav-item-rotate-icon: rotate(270deg);
+  }
+
+  &  cbp-subnav-item {
```
dgibson666: you don't need to use & for nested descendant selectors. Nesting equates to descendant selector by default.

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
```diff
@@ -0,0 +1,141 @@
...
+
+  
+  &[current] cbp-button:first-of-type,
+  &:has(cbp-subnav-item[current]) > div cbp-button:first-of-type{
+    border-left: var(--cbp-border-size-xl) solid var(--cbp-subnav-item-current-border-color);
+  }
+
+  &[current] cbp-button[fill=outline][color=primary]:first-of-type{
+    --cbp-button-color: var(--cbp-subnav-item-current-color);
+    background-color: var(--cbp-subnav-item-current-color-bg);
```
dgibson666: The current item does not have a blue background for me. You probably need to use the button CSS API to set --cbp-button-color-bg instead.

The Current item is also supposed to be italics (and bold? it looks bolder in the mockups).

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
```diff
@@ -0,0 +1,141 @@
...
+
+// }
+
+cbp-subnav-item {
+
+  --cbp-subnav-item-rotate-icon: rotate(90deg);
+  
+
+  
+  &[current] cbp-button:first-of-type,
```
dgibson666: Is there a reason not to set this left border on the div rather than the cbp-button? I think it would simplify and remove the need for the :first-of-type. And use border-inline-start logical property to be consistent.

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.tsx
```diff
@@ -0,0 +1,104 @@
+import { Component, Element, Prop, Host, h } from '@stencil/core';
+import { setCSSProps } from '../../utils/utils';
+
```
dgibson666, line 3: document your slots above the component decorator with JSDocs. See almost any of my components for examples. button has a named slot, so that would work.

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.tsx
```diff
@@ -0,0 +1,104 @@
...
+          aria-current={this.current}
+        >
+          <div>
+          <cbp-button
+            tag="a"
+            fill="outline"
+            color="primary"
+            href={this.href}
+          >
+            <slot name='icon'> </slot>
```
dgibson666: This needs to be a named slot for a full label. Not just an icon. cbp-subnav-item-label. There's a technique I use for optional slots something like:

```
{ !this.host.querySelector('[slot=cbp-subnav-item-label]') && this.label}
<slot name="cbp-subnav-item-label" />
```

This might be tricky in this case since subnav items are nested and it could pick up a child selector... might need a :scope > div [slot=cbp-subnav-item-label] to force it to look at the immediate child. Needs testing.

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.tsx
```diff
@@ -0,0 +1,104 @@
...
+      this.open= false;
+    }
+    else{
+      this.host.lastElementChild.removeAttribute('hidden');
+      this.open= true;
+    }
+  }
+
+  render() {
+    if(this.parent){
```
dgibson666: I don't think you need 2 separate returns. You can add the expand button inline with a logical and (&&). Same for the section tag. Is section needed? Why not just a div?

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.tsx
```diff
@@ -0,0 +1,104 @@
...
+      return (
+        <Host
+          aria-current={this.current}
```
dgibson666: aria-current only works for interactive elements, which the host tag is not. cbp-button is a special case where I'm testing out code to push aria-tags from the host to the rendered element. So if you move this to the cbp-button, it should be rendered on the actual anchor that's rendered.

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.tsx
```diff
@@ -0,0 +1,104 @@
...
+          </div>
+          <section hidden>
+            <slot />
+          </section>
+        </Host>
+      );
+    }else {
+      return (
+        <Host
+          aria-current={this.current}
```
dgibson666: Same as above: aria-current only works for interactive elements, which this host tag is not.

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  /** Specifies the context of the component as it applies to the visual design and whether it inverts when light/dark mode is toggled. Default behavior is "light-inverts" and does not have to be specified. */
+  @Prop({ reflect: true }) context: "light-inverts" | "light-always" | "dark-inverts" | "dark-always";
+
+  /** Supports adding inline styles as an object */
+  @Prop() sx: any = {};
+
+  componentWillLoad() {
+    this.icon = this.host.closest('cbp-subnav-item cbp-subnav-item cbp-subnav-item') ? 'caret-down' : 'chevron-right';    
+    this.host.querySelector('cbp-subnav-item') ? this.parent = true : this.parent = false;
```
dgibson666: This can be set more directly as:

```
this.parent =  !!this.host.querySelector('cbp-subnav-item'); // should force it to a boolean value
```

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+    if (typeof this.sx == 'string') {
+      this.sx = JSON.parse(this.sx) || {};
+    }
+    setCSSProps(this.host, {
+      ...this.sx,
+    });
+  }
+
+  toggleChildVisibility(){
```
dgibson666: I think this would be simple enough as "toggleSubnav" It should probably also have a custom event emitter so that apps can piggyback off these sorts of UI events.

This also needs to toggle the expanded prop on the control.

I also think you need to set a @watch on "current" that triggers this same function only when current=true (because you're going to toggle it false at the higher level). This allows for reactive changes to the prop outside of just user interaction. Might need to discuss this a little - it also needs to call the parent function that turns off current for all the others.

packages/web-components/src/components/cbp-subnav/cbp-subnav.stories.tsx
```diff
@@ -0,0 +1,135 @@
...
+        >
+            ${generateContent(items)}
+        </cbp-subnav>
+    `;
+};
+  
+export const Subnav = SubnavTemplate.bind({});
+
+Subnav.args = {
+        items: [
```
dgibson666, line 42: Each of these keys should have a "current" and "open" boolean so that you can test what it's like to set an item current at different levels.

packages/web-components/src/components/cbp-subnav/cbp-subnav.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  /** Specifies the context of the component as it applies to the visual design and whether it inverts when light/dark mode is toggled. Default behavior is "light-inverts" and does not have to be specified. */
+  @Prop({ reflect: true }) context: "light-inverts" | "light-always" | "dark-inverts" | "dark-always";
+
+  /** Supports adding inline styles as an object */
+  @Prop() sx: any = {};
+
+  setCurrentSubNav(activatedSubNavItem) {
```
dgibson666: I don't see this being called anywhere because of page reloads. Since the code isn't currently reactive (per previous comments), it's not possible to test, but it's not firing when I set a new item current in the DOM.

dgibson666 (commented):
See code comments - still quite a few things to consider.

It should probably also auto-expand the parents of the current item, but I'm willing to defer that for now as part of the integration with the nav.

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
bagrub: updated to add css vars --cbp-subnav-item-color, --cbp-subnav-item-color-bg for the defaults, dark mode still isn't great but it it is legible. will add correct values when Jeremy hands them off as discussed

packages/web-components/src/components/cbp-subnav/cbp-subnav.stories.tsx
bagrub: updated to use the 'Sub-Navigation' as requested

packages/web-components/src/components/cbp-subnav/cbp-subnav.stories.tsx
bagrub: updated to use the 'Sub-Navigation' as requested

packages/web-components/src/components/cbp-subnav/cbp-subnav.tsx
bagrub: updated to default value of 'Sub-Navigation' as requested

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
bagrub: removed unnessecary & from scss structure

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
bagrub: setting the --cbp-button-color-bg instead of the background value. Local testing looks correct to me currently, please note the parent of a current item doesn't have the background color (i think that might be why it looks inconsistent, by default i have item 2-1-1-3 set to true). Also added the bold on current item

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
bagrub: updated border to be set on div

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.tsx
bagrub: added slot documentation as requested

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.tsx
bagrub: updated as discussed

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.tsx
bagrub: update render logic to be more condensed

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.tsx
bagrub: updated as requested, confirmed aria-current is being set on anchor

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.tsx
bagrub: Added event emitter, refined logic/ styling to be based on the open prop, added @watch for current (confirmed it via DOM)

packages/web-components/src/components/cbp-subnav/cbp-subnav.tsx
bagrub: removed this function since it isn't being consumed

packages/web-components/src/components/cbp-subnav/cbp-subnav.stories.tsx
```diff
@@ -19,51 +19,60 @@ export default {
...
 export const Subnav = SubnavTemplate.bind({});
 
+Subnav.storyName='Sub-Navigation';
 Subnav.args = {
         items: [
             {
-                icon: `<cbp-icon slot='icon' name="home"></cbp-icon> `,
+                icon: `<cbp-icon name="home"></cbp-icon> `,
                 label: `Subnav Item 1`,
                 href: './?path=/story/components-subnav--subnav',
```
dgibson666: You need to update all these paths since you renamed the story. Currently broken URL.

packages/web-components/src/components/cbp-subnav/cbp-subnav.stories.tsx
```diff
@@ -19,51 +19,60 @@ export default {
...
-            //   current: true,
+                current: false,
             },
             {
                 label: "Subnav Item 2",
                 href: './?path=/story/components-subnav--subnav',
+                open: undefined,
```
dgibson666: why are you using undefined instead of false? False should work fine.

dgibson666 (commented):
Couple of code comments. Still need to apply a background color to the subnav items to make the story not completely broken in dark mode.

There are a couple more issues related to reactivity I'm willing to defer to the next iteration.

packages/web-components/src/components/cbp-subnav/cbp-subnav.stories.tsx
bagrub: updated to correct URL

packages/web-components/src/components/cbp-subnav/cbp-subnav.stories.tsx
bagrub: set these to false. was set to undefined before i updated the toggleSubnav function logic so i think this just got oversite of update

dgibson666 (approved):
Looks good for first cut.

PR #259, opened 2025-02-27
Feature/code snippet

bagrub:


packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.scss
```diff
@@ -0,0 +1,72 @@
+/**
+ * @Prop --cbp-chip-color-text: var(--cbp-color-text-lightest);
```
dgibson666: delete?

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.scss
```diff
@@ -0,0 +1,72 @@
...
+  --cbp-code-snippet-color: var(--cbp-code-snippet-color-dark);
+  --cbp-code-snippet-color-bg: var(--cbp-code-snippet-color-bg-dark);
+
+}
+
+cbp-code-snippet {
+  color: var(--cbp-code-snippet-color);
+  background-color: var(--cbp-code-snippet-color-bg);
+
+  &[variant="inline"]{
```
dgibson666: Why inline block? You cannot place this in the flow of text as such. I think this whole selector can go away and inherit the default inline display of the custom element.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.scss
```diff
@@ -0,0 +1,72 @@
...
+    display: inline-block;
+  }
+  
+  &[variant="block"]{
+    display: block;
+    pre{
```
dgibson666, line 52: I don't know if if needs to go on pre or code, but you're not specifying our mono font family design token, so it's using the wrong/default monospaced fonts.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.specs.mdx
```diff
@@ -0,0 +1,41 @@
+import { Meta } from '@storybook/addon-docs';
+
+<Meta title="Components/Code Snippet/Specifications" />
+
+# cbp-code-snippet
+
+## Purpose
+
```
dgibson666, line 8: You've got like zero punctuation in this whole file :-)

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.scss
```diff
@@ -0,0 +1,72 @@
...
+      display: flex;
+      --cbp-code-snippet-pre-padding: var(--cbp-space-2x);
+      overflow: auto;
+      height: var(--cbp-code-snippet-height);
```
dgibson666: You're setting this as a height rather than max-height, so in the storybook example when I set any sizable max-height value (e.g., 100 or 200px), it actually shrinks the size when you click the button to "show more" (which is not at all needed because the block it larger than the code snippet) and expands when you click "show less". In reality, the button should not show up at all.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.specs.mdx
```diff
@@ -0,0 +1,41 @@
...
+## Technical Specifications
+
+* Variants: 2 variants of code snippet are supported 
+    * Inline: displays simple code inline with rest of text
+    * Block: displays blocks of code in a scrollable container w/ internal button to copy code block
+* min-height prop for Block variant, when used will populate the 'show more' button
+* expanded is used w/ the 'show more' button to pass to cbp-button expanded prop
+
+### User Interactions
+
```
dgibson666, line 26: needs docs

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.specs.mdx
```diff
@@ -0,0 +1,41 @@
...
+* 
+
+### Responsiveness
+
+*
```
dgibson666: needs docs. I'm not sure using flex will give good results on a small screen. even larger screens with longer code could be problematic.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.specs.mdx
```diff
@@ -0,0 +1,41 @@
...
+* 
+
+### Responsiveness
+
+* 
+
+### Accessibility
+
+* tab focus to internal buttons 'copy'/'Show more'
+
```
dgibson666, line 40: + the aria-expanded on the show more button
+ buttons having accessible names.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+    title: 'Components/Code Snippet',
+    //tags: ['autodocs'],
+    argTypes: {
+      codeSnippet: {
+        name: 'Code (slotted)',
+        description: 'Code to slot into code snippet',
+        control: 'text',
+      },
+      variant: {
+        name: 'Code Snippet Variant',
```
dgibson666: you don't need these more verbose/redundant names. Using the prop name when they are props results in less confusion.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+        description: 'determines the display type of the code snippet',
+        control: 'select',
+        options: [ "inline", "block"]
+      },
+      height: {
+        name: 'Code Snippet max-height',
```
dgibson666: same - use prop name, not longer, verbose name.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+      },
+      sx: {
+        description: 'Supports adding inline styles as an object of key-value pairs comprised of CSS properties and values. Values should reference design tokens when possible.',
+        control: 'object',
+      },
+    },
+    args: {     
+      /** code snippet passed to the <code> block so any indent/spacing will be in final render*/
+      codeSnippet: `<a href='#'> test code snippet </a>`,
+      variant: 'inline'
```
dgibson666: remove this so it doesn't force the prop to be set for what is the default value.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+    },
+  };
+  
+  const Template = ({codeSnippet, variant, height, context, sx}) => {
+    return ` 
+    <cbp-code-snippet
+      variant= ${variant}
```
dgibson666: then this should be a ternary to only output if a value is defined.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
```diff
@@ -0,0 +1,113 @@
...
+      this.toggleButtonText = 'Show Less';
+    }else{
+      this.host.style.setProperty('--cbp-code-snippet-height', this.height)
+      this.showAllToggle = false;
+      this.toggleButtonRotate = 90
+      this.toggleButtonText = 'Show More';
+    }
+    this.expanded = !this.expanded
+    this.toggleShowAllClick.emit({
+      host: this.host
```
dgibson666: add the expanded state as well

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
```diff
@@ -0,0 +1,113 @@
...
+    if (typeof this.sx == 'string') {
+      this.sx = JSON.parse(this.sx) || {};
+    }
+    setCSSProps(this.host, {
+      ...this.sx,
+    });
+  }
+
+  componentDidLoad(){
+    if(this.height != null && !this.expanded){
```
dgibson666, line 85: this doesn't seem to be working to me. If I set the height to 200px, it's still showing the button when the code is shorter. You need to check against the actual  content height.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
```diff
@@ -0,0 +1,113 @@
...
+            fill="ghost"
+            color="secondary"
+            onClick={() => this.copyText()}
+          >
+            Copy
+          </cbp-button>
+        }
+        </pre>
+        
+        {this.height &&
```
dgibson666: this should check against a state that is set based on the componentDidLoad calculations of actual content.

dgibson666 (commented):
See code comments.

Also, the buttons do not respond to context changes. Since they are rendered by the component, they should be managed internally by the component.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.scss
bagrub: deleted

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.scss
bagrub: updated to inline. was originally inline-block to have 1 statement for bg/color statement, moved bg/color statement inside variant blocks to ensure we are still getting correct var mapping with this update

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.scss
bagrub: added the token to the code tag

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.specs.mdx
bagrub: updated to add punctuation

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.scss
bagrub: updated this to be max-height instead of height based on discussion

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.specs.mdx
bagrub: updated & added white-space: pre-wrap to the pre tag to address issues w/ smaller screensizes

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.specs.mdx
bagrub: updated to add interactions w/ buttons

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.specs.mdx
bagrub: updated as requested

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.stories.tsx
bagrub: removed

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.stories.tsx
bagrub: removed

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.stories.tsx
bagrub: removed

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.stories.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
bagrub: updated

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
bagrub: updated to add prop to pull current dom value of container & code as discussed

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
bagrub: updated to match with above comment on didLoad()

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
```diff
@@ -84,23 +99,27 @@ export class CbpCodeSnippet {
     
         {this.variant == 'block' &&
           <cbp-button
+            name='copy'
```
dgibson666: remove. that's not what accessible name mean. The text on the button is the accessible name.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
```diff
@@ -84,23 +99,27 @@ export class CbpCodeSnippet {
...
           >
             Copy
           </cbp-button>
         }
         </pre>
         
-        {this.height && 
+        {this.maxheight && (this.codeContainerHeight < this.codeBlockHeight) &&
           <cbp-button
+            name= {this.toggleButtonText}
```
dgibson666: remove. that's not what accessible name mean. The text on the button is the accessible name.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
```diff
@@ -65,10 +74,16 @@ export class CbpCodeSnippet {
     });
   }
 
+
   componentDidLoad(){
-    if(this.height != null && !this.expanded){
-      this.host.style.setProperty('--cbp-code-snippet-height', this.height);
-      console.log('check for height var placement');
+    setTimeout(() => { //** timeout need for DOM to render and correct values to populate for below */
+      this.codeContainerHeight = this.host.closest('cbp-code-snippet').offsetHeight;
```
dgibson666: you don't need to do closest because it's getting the host itself.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
```diff
@@ -65,10 +74,16 @@ export class CbpCodeSnippet {
...
+      this.codeBlockHeight = this.host.closest('cbp-code-snippet').querySelector('pre').scrollHeight;
```
dgibson666: same. this.host.querySelector()... no closest needed.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
```diff
@@ -13,13 +13,21 @@ export class CbpCodeSnippet {
...
   private toggleButtonRotate = 90;
+  // private codeContainerHeight;
+  // private codeBlockHeight;
 
   /**  Specifies the visual variant of the code snippet*/
   @Prop({ reflect: true }) variant: "inline" | "block" = 'inline';
   
-  /** Specifies the min height for a muliple line block variant(in CSS units). If set will populate the 'show more' button on the render to display full height */
-  @Prop() height: string;
+  /** Specifies the max height for a muliple line block variant(in CSS units). */
```
dgibson666, line 22: I'm on the fence if this should be changed. "height" is easier to understand even if it's applied as max-height in CSS.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.specs.mdx
```diff
@@ -6,36 +6,40 @@ import { Meta } from '@storybook/addon-docs';
...
 
-* Variants: 2 variants of code snippet are supported 
-    * Inline: displays simple code inline with rest of text
-    * Block: displays blocks of code in a scrollable container w/ internal button to copy code block
-* min-height prop for Block variant, when used will populate the 'show more' button
-* expanded is used w/ the 'show more' button to pass to cbp-button expanded prop
+* Variants: 2 variants of code snippet are supported, see below for breakdown of variants. 
+    * Inline: displays simple code inline with rest of text.
+    * Block: displays blocks of code in a scrollable container w/ internal button to copy code block.
+* min-height prop for Block variant.
```
dgibson666: You should be describing the technical requirements without prop names (which is tied to implementation but not a requirement).

Check this whole file for caps and punctuation though. Still spotty and inconsistent.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.scss
```diff
@@ -31,20 +30,26 @@
...
+
     pre{
       display: flex;
       --cbp-code-snippet-pre-padding: var(--cbp-space-2x);
       overflow: auto;
-      height: var(--cbp-code-snippet-height);
+      max-height: var(--cbp-code-snippet-max-height);
     
       code{
         padding-block-start: var(--cbp-space-2x);
```
dgibson666: This is applying extra padding to the top, which makes the whole code block look vertically uncentered (because it is now). Is this intentional? It doesn't look right to me and was the first thing I noticed and dug into.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.scss
```diff
@@ -31,20 +30,26 @@
 }
 
 cbp-code-snippet {
-  color: var(--cbp-code-snippet-color);
-  background-color: var(--cbp-code-snippet-color-bg);
 
   &[variant="inline"]{
-    display: inline-block;
+    display: inline;
```
dgibson666: This isn't needed because inline is the default display of custom elements.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.scss
```diff
@@ -31,20 +30,26 @@
...
+  
+    code{
+      color: var(--cbp-code-snippet-color);
+      background-color: var(--cbp-code-snippet-color-bg);  
+    }
   }
   
   &[variant="block"]{
     display: block;
+    color: var(--cbp-code-snippet-color);
```
dgibson666: Colors that are the same in both variants shouldn't be repeated.

dgibson666 (commented):
See code comments.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
```diff
@@ -0,0 +1,132 @@
...
+        {this.variant == 'block' &&
+          <cbp-button
+            name='copy'
+            type="button"
+            fill="ghost"
+            color="secondary"
+            onClick={() => this.copyText()}
+            context={this.context}
+          >
+            Copy
```
dgibson666, line 109: The copy button is missing its icon and also rendering in a non-roboto monospaced font. It should not be mono.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
bagrub: removed

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
bagrub: removed

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
bagrub: removed

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
bagrub: removed

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
bagrub: updated to 'height' as discussed

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.specs.mdx
bagrub: update as requested

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.scss
bagrub: removed. Was initially added to bring first line of code to be more centered with copy button for the 'single' variant when that was still part of this

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.scss
bagrub: removed

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.scss
bagrub: discussed in call. keeping in due to weirdness with display: inline & background color

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
bagrub: updated

dgibson666 (approved):
looks good!

PR #260, opened 2025-03-10
Bugfix/pagination a11y

dgibson666:
* Refactor/cleanup on code snippet
* Fixed missing CSS API docs across the whole project
* Updates to dropdown to normalize ID assignment and fix the label/field ID mismatch in stories
* Make Pagination next/prev buttons match input border color
* Apply outline highlight to menu when the control is focused (which it should be while open except for special cases using a screen reader)


bagrub (approved):
LGTM

PR #261, opened 2025-03-18
Passenger list structured list

bagrub:
Creation of the passenger list archetype focus on the structured list. Filter & manifest to be refined in another story

### DIFF
--- a/packages/web-components/src/stories/Archetype_Passenger.stories.tsx
+++ b/packages/web-components/src/stories/Archetype_Passenger.stories.tsx
@@ -1,0 +1,879 @@
+export default {
+    title: 'Archetypes/Passenger List',
+    parameters: {
+      layout: 'fullscreen',
+      html: {
+        root: '#storybook-root',
+      },
+    },
+    argTypes: {
+    },
+    args: {
+      username: 'UI Test',
+      isLoggedIn: true,
+    },
+  };
+
+    // const passengerArchetypeList = HTMLCbpStructuredListElement;
+    //TODO: needs default values (currently hard set to component defaults) below is 'ideal' but DOM not rendered at this point
+      let page = 1; //document.getElementsByTagName('cbp-pagination')[0].page;
+      let pageSize: number | "all" = 10; //document.getElementsByTagName('cbp-pagination')[0].pageSize;
+
+    function getTimeDiff(date1, date2){
+      const timeDiff = Math.abs(date1 - date2);
+      const seconds = Math.floor(timeDiff / 1000);
+      const minutes = Math.floor(seconds / 60);
+      const hours = Math.floor(minutes / 60);
+      const days = Math.floor(hours / 24);
+
+      return {days, hours: hours % 24, minutes: minutes % 60, seconds: seconds % 60}
+    }
+
+    function generatePassengers(passengerArgs, page, pageSize) {
+      //TODO: update api call for img to have some variation see doc: https://www.dicebear.com/styles/personas/
+      //TODO: math on the time to arrival format isn't great but not sure how to setup obj to not have diff for days
+      
+      const html = passengerArgs.map(({name, bio, arrival, error }, index) => {
+      const timeDiff = getTimeDiff(Date.now(), arrival) 
+      const formatedTime = arrival.toLocaleString('en-US', {
+        timeZone: 'America/New_York',
+        timeZoneName: "short"
+      });
+
+      if(index >= ((page-1) * pageSize) && index < (page * pageSize)){
+        return `
+          <cbp-structured-list-item
+            ${error ? `color=danger`: ``} 
+          >
+            <cbp-grid
+              grid-template-columns="14rem 1fr 10rem"
+              gap="5rem"
+            >
+              <cbp-flex>
+                  <img
+                    src="https://api.dicebear.com/9.x/personas/svg"
+                    alt="avatar"
+                    style="width: 100px"
+                  />
+                <cbp-flex direction="column" gap="var(--cbp-space-1x)">
+                  <cbp-typography tag="h3">${name}</cbp-typography>
+                  <cbp-tag> Arriving In: ${timeDiff.hours}:${timeDiff.minutes}: ${timeDiff.hours} </cbp-tag>
+                  ${error ?
+                     `<cbp-tag color="danger"> T-list</cbp-tag>`
+                    : ``}
+                </cbp-flex>
+              </cbp-flex>
+              <cbp-flex direction="column" gap="var(--cbp-space-1x)">
+                <span>
+                  <b>BIO:</b> ${bio}
+                </span>
+                <span>
+                  <b>Flight Inbound</b> A471 from LHR to JFK (A99/SM)
+                </span>
+                <span>
+                  <b>Scheduled Arrival Time:</b> ${formatedTime}
+                </span>
+              </cbp-flex>
+              <cbp-grid-item align-self="center">
+                <cbp-button
+                  type="button"
+                  fill="ghost"
+                  color="secondary"
+                >
+                  <cbp-icon name="book"></cbp-icon> Vet Passenger
+                </cbp-button>
+              </cbp-grid-item>
+            </cbp-grid>
+          </cbp-structured-list-item>`
+        }
+    });
+    return html.join('');
+    }
+
+    function generateManifest(manifestArgs){
+      //TODO: tag icon for gate is not correct 
+
+      const html = manifestArgs.map(({flight, arrivalTerminal, arrivalLocation, arrivalGate, arrivalTime, departureTerminal, departureLocation, departureTime, people, hotlist}) => {
+      const timeDiff = getTimeDiff(Date.now(), arrivalTime);
+      
+        return `
+          <cbp-card
+            variant="decision"
+            class="hydrated"
+          >
+            <div>
+              <cbp-flex 
+                direction="row"
+                gap='33%'
+              >  
+                <cbp-typography
+                  tag="h3"
+                 
+                  id="card-heading-1"
+                  class="hydrated"
+                >
+                  ${flight}
+                </cbp-typography>
+                <cbp-tag color="danger"> Arriving In: ${timeDiff.hours}:${timeDiff.minutes}: ${timeDiff.hours} </cbp-tag>
+              </cbp-flex>
+
+              <cbp-flex 
+                direction="row"
+                gap='1rem'
+              >
+                <cbp-flex direction="column">
+                  <cbp-typography tag="h4">${departureTerminal}</cbp-typography>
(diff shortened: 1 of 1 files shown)


packages/web-components/src/stories/Archetype_Passenger.stories.tsx
```diff
@@ -0,0 +1,879 @@
...
+        const structuredListComponent = document.querySelector("#passengerList > div[role='list']");
+        page = paginationComponent.page;
+        pageSize = paginationComponent.pageSize;
+  
+        structuredListComponent.innerHTML = generatePassengers(passengerArgs, page, pageSize);
+      });
+  
+      return `
+      <cbp-flex-item flex-grow="1">
+            <cbp-grid
```
dgibson666: This sort of inline set of controls makes more sense to use flex than grid. The buttons should wrap if there's not enough room to show them. You can use a flex basis on the dropdown to enforce a minimum width of sorts.

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
```diff
@@ -0,0 +1,879 @@
...
+                </cbp-dropdown>
+              </cbp-form-field>
+              <cbp-grid-item align-self="center">
+                <cbp-button
+                  type="button"
+                  fill="outline"
+                  color="secondary"
+                  target-prop="open"
+                  controls="manifestDrawer"
+                  accessibility-text="Open Manifest Drawer"
```
dgibson666: shouldn't need accessibility text when there's visible text as the label.

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
```diff
@@ -0,0 +1,879 @@
...
+            >
+              <cbp-typography
+                slot="cbp-panel-header"
+                tag="h3"
+                variant="heading-lg"
+                id="panelheader"
+              >
+                Sidebar Header
+              </cbp-typography>
+                <cbp-form-field
```
dgibson666, line 327: you have everything nested inside of this form field and it should not be. The search is the first form field (and its bottom margin should space it out from the next filter).

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
```diff
@@ -0,0 +1,879 @@
...
+      </cbp-app-header>
+    </header>
+
+    <cbp-container sx='{"padding":"1rem var(--cbp-responsive-spacing-outer)"}'>
+      <main id="main" tabindex="-1">
+        <cbp-typography tag="h1" divider="underline" sx='{"margin-bottom":"var(--cbp-space-5x)"}'>
+          Passenger Vetting
+        </cbp-typography>
+
+        <cbp-flex
```
dgibson666: structural layout of a page is better done using grid, as copied from the 2 column template. Also, if you look at that, you'll see that the `main` tag only wraps the content area. The filter panel can be marked up like the sidebar in that storym which functions as an `aside` with its role and label.

Since the manifest is a drawer, I would place it after everything - maybe between the container and footer.

dgibson666 (commented):
Overall, I would prefer to see the main structure as one large chunk of HTML like viewing a static page. You can still use functions for  the dynamic data like the structured list contents/items, but I don't think breaking the different structural parts of the site into separate functions helps - it actually made it harder for me to follow. You can still mark structure in the template string with HTML comments to make it readable.

See my code comments as well.

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
```diff
@@ -0,0 +1,879 @@
...
+                <cbp-form-field group
+                  label='Passenger Sex'
+                  description='Required'
+                >
+                  <cbp-flex
+                    gap="var(--cbp-space-5x)"
+                    breakpoint="28rem"
+                    sx='{"width":"max-content"}'
+                  >
+                    <cbp-radio>
```
dgibson666, line 362: These radios aren't marked up properly and keyboard nav doesn't work as a result. All radios in a group must share the same name.

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
```diff
@@ -0,0 +1,879 @@
...
+                  <cbp-form-field-wrapper>
+                    <input
+                      type="time"
+                    />
+                    <span slot="cbp-form-field-overlay-start">
+                    <cbp-icon name="book"></cbp-icon>
+                    </span>
+   
+                    <span slot="cbp-form-field-attached-button">
+                      <cbp-button
```
dgibson666, line 467: These are supposed to act like segmented buttons, so wrap the individual buttons in a segmented button group and make sure the markup is correct. I have not tested and made segmented buttons work in this context (straightening the leftmost corners), but it should be easy to do (later) and look no worse than this does.

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
bagrub: update to be flex-item

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
bagrub: removed

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
bagrub: was missing the closing tag for this, updated to be separate form-fields

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
bagrub: update to be grid as per the template pages

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
bagrub: update to be correct markup

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
bagrub: wrapped in segmented buttons instead of span

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
```diff
@@ -179,51 +183,73 @@ export default {
...
                   color="secondary"
                   target-prop="open"
                   controls="manifestDrawer"
-                  accessibility-text="Open Manifest Drawer"
                 >
-                  <cbp-icon name="book"></cbp-icon> Manifests
+                  <cbp-icon 
+                    name="book"
+                    sx='{"margin-right":"var(--cbp-space-2x)"}'
+                    ></cbp-icon> Manifests
```
dgibson666: you need to be careful putting a space after the icon because it will make the spacing larger than expected with the margin.

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
```diff
@@ -179,51 +183,73 @@ export default {
...
-            <cbp-grid
-              grid-template-columns=" 1fr 8.5rem 8.5rem"
+            <cbp-flex
               gap="1rem"
+              breakpoint="70rem"
             >
-              <cbp-form-field
-                label="Sort By"
-                field-id="filterResults"
+              <cbp-flex-item
```
dgibson666, line 192: the dropdown is taking up way too much space. I would put a flex-basis on it to give it a default width that works and also flex-shrink=1.

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
```diff
@@ -341,23 +367,23 @@ export default {
                     <cbp-radio>
                       <input 
                         type="radio" 
-                        name="Male"
+                        name="Passenger Sex"
```
dgibson666: why not just use "sex"? You would never want a space in something that becomes a DOM and data reference - this could cause a wide array of issues.

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
```diff
@@ -442,7 +468,7 @@ export default {
                     <cbp-icon name="book"></cbp-icon>
                     </span>
    
-                    <span slot="cbp-form-field-attached-button">
+                    <cbp-segmented-button-group slot="cbp-form-field-attached-button">
```
dgibson666, line 466: These turned out great. Doesn't look like they need any more work.

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
```diff
@@ -621,9 +647,26 @@ export default {
      * Filter panel should have a fixed max width probably
      * Footer missing InfoSec section (right side of footer)
      * 
-    */
-    
-    return ` 
+    */       
+
+/**
```
dgibson666: What was this big block of sx for? Delete if not needed.

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
```diff
@@ -0,0 +1,922 @@
...
+    parameters: {
+      layout: 'fullscreen',
+      html: {
+        root: '#storybook-root',
+      },
+    },
+    argTypes: {
+    },
+    args: {
+      username: 'UI Test',
```
dgibson666: change this back to "HASHID" to match the template exactly. Jeremy intends for this to be a hash id over user name, so it's instructive placeholder text.

dgibson666 (commented):
see code comments

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
bagrub: removed space char as discussed on call

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
bagrub: updated for flex-basis & flex-shrink

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
bagrub: updated as requested

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
bagrub: removed

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
bagrub: updated back to HASHIDX

dgibson666 (approved):
looks good overall. some minor things, but they can be iterated on later.

PR #263, opened 2025-03-25
Loader updates

bagrub:
#### updates to the loader based on visual design review:
* update for slotted label to render on indeterminate variants
* update font size & weight for slotted text
* update Icon sizes
* update variable color mappings

### DIFF
--- a/packages/web-components/src/components/cbp-loader/cbp-loader.scss
+++ b/packages/web-components/src/components/cbp-loader/cbp-loader.scss
@@ -4,7 +4,7 @@
  * @prop  --cbp-loader-center-color: var(--cbp-color-branding-cbp-light);
  * @prop  --cbp-loader-center-color-dark: var(--cbp-color-branding-cbp-dark);
  * @prop  --cbp-loader-track-color: var(--cbp-color-gray-cool-20);
- * @prop  --cbp-loader-track-color-dark: var(--cbp-color-gray-cool-60);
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/cbp-loader/cbp-loader.scss
```diff
@@ -80,6 +84,12 @@
 
 cbp-loader {
```
dgibson666: I'm seeing some odd side effects in different cases like linear small (inline) variant causing horizontal scroll bars and scrolling past the edge of the screen. Adding these properties to the host element fixes it by making it a container:

```
    max-width: 100%;
    overflow: hidden;
    position: relative;
```

dgibson666 (commented):
See 1 code comment. Also, is there a reason the label is hidden for small circular? The control is still there is Storybook and it seems like a valid case to still say "Uploading" or whatever. The same space is used for the success or error states.

packages/web-components/src/components/cbp-loader/cbp-loader.scss
bagrub: updated as discussed

dgibson666 (approved):
Looks good overall. Even though we prevented the small linear variant from overflowing, the animation calculation could probably be tweaked for the appropriate width. It's obvious from the easing that it's ending "off screen" in smaller viewports. But that's fairly minor and can be a future iteration.

PR #264, opened 2025-03-28
toast update

bagrub:


dgibson666 (commented):
The color of the icon for warning in dark mode appears to be incorrect. Everything else looks good. I would probably use the appropriate icons for each color variant rather than the person. They should be fairly standard with the option to customize (per the implementation).

dgibson666 (approved):
looks good. Make a note to update the icons to the proper ones sometime in a later bugfix.

PR #267, opened 2025-04-04
code snippet update

bagrub:
code snippet updates based on visual design review


packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
```diff
@@ -109,7 +109,6 @@ export class CbpCodeSnippet {
           {this.variant == 'block' && 
             <cbp-button type="button" fill="ghost" color="secondary" onClick={() => this.copyText()} context={this.context}>
```
dgibson666: If this button has no accessible label, it needs the prop `accessibilityText="Copy"`. It also needs `variant="square"` since it's now an icon button.

dgibson666 (commented):
See code comments about copy button. Everything else looks good.

packages/web-components/src/components/cbp-code-snippet/cbp-code-snippet.tsx
bagrub: updated to add the accessibilityText & variant

dgibson666 (approved):

PR #268, opened 2025-04-09
sub nav dark mode update

bagrub:
update sub nav variable mapping for darkmode, update indentation spacing, & add prop boolean for indentation

### DIFF
--- a/packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
+++ b/packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
@@ -1,4 +1,4 @@
-/** TODO: update this with dark mode values
+/** 
  * @prop  --cbp-subnav-item-color: var(--cbp-color-interactive-secondary-darker);
  * @prop  --cbp-subnav-item-color-dark: var(--cbp-color-interactive-secondary-lighter);
  * @prop  --cbp-subnav-item-color-bg: transparent;
@@ -29,10 +29,9 @@
   --cbp-subnav-item-font-size: var(--cbp-font-size-heading-md);
 }
 
-[data-cbp-theme=light] cbp-subnav-item[context*=dark]:not([context=light-always]),
-[data-cbp-theme=dark] cbp-subnav-item:not([context=dark-inverts]):not([context=light-always]) {
+[data-cbp-theme=light] cbp-subnav[context*=dark]:not([context=light-always]),
+[data-cbp-theme=dark] cbp-subnav:not([context=dark-inverts]):not([context=light-always]) {
 
-  
   --cbp-subnav-item-color: var(--cbp-subnav-item-color-dark);
   --cbp-subnav-item-color-bg: var(--cbp-subnav-item-color-bg-dark);
   --cbp-subnav-item-current-border-color: var(--cbp-subnav-item-current-border-color-dark); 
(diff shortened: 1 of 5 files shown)


packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
```diff
@@ -1,44 +1,43 @@
...
-  // --cbp-subnav-item-current-border-color-dark: var(--cbp-color-interactive-active-light); 
+  --cbp-subnav-item-current-border-color-dark: var(--cbp-color-interactive-active-light); 
 
   --cbp-subnav-item-indent: var(--cbp-space-4x);
   --cbp-subnav-item-font-size: var(--cbp-font-size-heading-md);
 }
 
-// [data-cbp-theme=light] cbp-subnav-item[context*=dark]:not([context=light-always]),
-// [data-cbp-theme=dark] cbp-subnav-item:not([context=dark-inverts]):not([context=light-always]) {
+[data-cbp-theme=light] cbp-subnav-item[context*=dark]:not([context=light-always]),
```
dgibson666: Each item should not need or have context. This should be handled at the parent subnav component level. You can still reference that parent component at this level for the CSS vars toggle. For example, I've done this in dropdown.

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
```diff
@@ -1,44 +1,43 @@
-/** 
- * @prop  --cbp-subnav-item-current-border-color: var(--cbp-color-interactive-active-dark);
+/** TODO: update this with dark mode values
```
dgibson666: remove todo

packages/web-components/src/components/cbp-subnav/cbp-subnav.tsx
```diff
@@ -11,6 +11,9 @@ export class CbpSubNav {
   /**  Sets the aria-label for the navigation element of the subnav*/
   @Prop() accessibilitytext: string = 'Sub-Navigation';
 
+  /** used to toggle child indenation */
+  @Prop({ reflect: true }) indentation: boolean = true;
```
dgibson666: you cannot set a boolean prop to true by default because the attribute equivalent of false is no attribute (in which case it's still true to the component). This needs to be rethought. Do we want to call it "flat"? Open for discussion. but I also avoid nagatives, so "noIndent" is off the table.

dgibson666 (commented):
Couple of major things in code and I'm still seeing the right border on the buttons. I would like to discuss the button implementation again. I know we talked about why you didn't put the borders on the subnav-items, but based on the styling, I'm still seeing a gray line (border) even when the item link is focused.

We also need to put a techdebt/todo note about routing support if it's not in there already.

Also, still see a selector for `[slot="icon"]` in your CSS, but I don't think that exists in the component anymore.

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
bagrub: updated to use the cbp-subnav selector, also passed context from subnav-item to buttons in render so they update as appropriate

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
bagrub: removed

packages/web-components/src/components/cbp-subnav/cbp-subnav.tsx
bagrub: update prop name to 'flat' as suggested & updated logic

dgibson666 (approved):
looks good.

PR #270, opened 2025-04-14
Feature/dropdown async

dgibson666:
* Update dropdown to accept items as JSON (rather than slotted).
* Updates to dropdown for asynchronous updates via API call.
* Close dropdown when tabbing out of it.
* Fix for faux-button gap on select/dropdown at some resolutions/zoom.
* Story for demoing multiple dropdowns, but also compound input patterns


bagrub (approved):
LGTM

PR #272, opened 2025-04-15
Bugfix/table slider design review

dgibson666:
* Fix slider bug when min and max are not 0 and 100.
* Fix slider design review issues.
* Update slider docs.
* Create volume slider story with slotted icons.
* Clean up form field group story issues.
* Fix table design review issues.


bagrub (approved):
LGTM

PR #283, opened 2025-04-24
Feature/inline drawer

dgibson666:
* Updated the drawer to be able to persist in the flow of the document at a certain breakpoint.
* Updated the Passenger list story to use a persistent drawer.
* Fixed drawer and dialog controls to be square using `variant=square`.

I'm deferring the dark mode update to the backdrop, because I think that should be tokenized rather than 1-offs in 2 different places. That can be a separate PR with any dark mode updates Jeremy has planned.


bagrub (approved):
LGTM

PR #286, opened 2025-05-01
app header integration with subnav

bagrub:


packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
```diff
@@ -5,67 +5,181 @@ export default {
     layout: 'fullscreen',
   },
   argTypes: {
+    open: {
```
dgibson666: I don't think it makes sense to have an open control since you'd never have a drawer open by default.

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
```diff
@@ -5,67 +5,181 @@ export default {
...
+      description: 'Specifies whether the drawer is open or closed.',
+      control: 'boolean',
+    },
+    uid: {
```
dgibson666: I'd probably name this "drawerId" for clarity. And as such, it doesn't make much sense to give it a value of "applicationHeader" - this id is placed on the drawer component itself.

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
```diff
@@ -5,67 +5,181 @@ export default {
...
   },
 };
 
-function generateNavItems(navItems){
-  const html =  navItems.map(({html, selected}) => {
-      return `<cbp-nav-item ${selected ? 'selected' : ''}> ${html}</cbp-nav-item>`;
+function generateNavItems(items, uid){
+  const html =  items.map(({ label, name, href, selected, children}) => {
+    if(!children){
+      return `<cbp-nav-item name="${name}" ${selected ? 'selected' : ''}>  <cbp-button tag='a' onclick=${function(e){e.preventDefault()}} href=${href} fill="ghost" color="secondary">${label}</cbp-button></cbp-nav-item>`;
```
dgibson666: I don't think this prevent default is working like you think it is. First, it's throwing a console error (which is why it might appear to not navigate). But it's not on the native button, so I don't know if you can cancel it here. Rather than some inline event, you need to use addEventListener() on the rendered button like I have elsewhere in stories. (I don't recall where, but I demoed it previously).  Furthermore, putting it inline in the story code, means it gets copied from the story markup, which we don't want.

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
```diff
@@ -5,67 +5,181 @@ export default {
...
+    } else {
+      return `<cbp-nav-item name="${name}" ${selected ? 'selected' : ''}>  <cbp-button tag='button' fill="ghost" color="secondary" target-prop="open" controls=${uid}>${label} <cbp-icon name="chevron-right" sx='{"transform":"rotate(90deg)", "margin-left":"0.25rem"}'></cbp-icon></cbp-button></cbp-nav-item>`;
```
dgibson666: Icons have a rotate prop - you don't need to do that in `sx`.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -10,6 +11,11 @@ export class CbpAppHeader {
   private navItems: HTMLCbpNavItemElement[] = [];
 
   @Element() host: HTMLElement;
+  @Listen('drawerClose')
+  handleNavDrawerClose() {
+    var current = this.host.querySelector('[name=\"' + state.activeItemName + '\"] cbp-button').firstElementChild as HTMLAnchorElement;
```
dgibson666: 1. using var.
2. var is named current, but it appears to be active, not current.
3. use template literals in your query selector rather than string concatenation - it's cleaner and more modern. e.g., 
```
`[name="${state.activeItemName}"] cbp-button`
```
4. Select the thing you want. no need for firstElementChild in addition.
5. This is listening for any drawerClose event. What happens if it's not a subnav drawer? Needs to be scoped somehow so that you're only doing these things when applicable.

packages/web-components/src/components/cbp-app-header/store.ts
```diff
@@ -0,0 +1,31 @@
...
+
+onChange('activeItemName', value => {
+  state.activeItemName = value;
+  updateSelected();
+});
+
+function updateSelected(){ //TODO: move this to not the store file
+  resetSubNav();
+  let attr = '[name=\"' + state.activeItemName + '\"]' 
+  let subNavItem = document.querySelectorAll(attr)[1] as HTMLCbpSubnavItemElement;
```
dgibson666: use a template literal and this is fine as a 1-liner.

packages/web-components/src/components/cbp-app-header/store.ts
```diff
@@ -0,0 +1,31 @@
...
+  subNavItem.setAttribute("open", "true");
```
dgibson666: if you've typed it as the proper type, you can set the prop via dot notation. No need for setAttribute. `subNavItem.open=true;`

packages/web-components/src/components/cbp-app-header/store.ts
```diff
@@ -0,0 +1,31 @@
...
+  setTimeout(() => {subNavItem.querySelector('a').focus()}, 101) //Note: Time 101 is set due to cbp-drawer setting @ 100
+}
+
+function resetSubNav(){//TODO: move this to not the store file
+  let subnav = document.querySelectorAll('cbp-subnav-item');
+
+  subnav.forEach(e => { e.hasAttribute("open") ? e.removeAttribute("open") : '' });
```
dgibson666: This should use props rather than setAttribute. It also might make more sense to make a method on the `cbp-subnav` component to do that. Make a TechDebt note if you don't want to do that right now.

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.tsx
```diff
@@ -1,22 +1,30 @@
...
+  @Prop({ reflect: true }) name: string;
+
   /** Supports adding inline styles as an object */
   @Prop() sx: any = {};
   
+  //techdebt: event logic only works for links
   @Event() navClicked: EventEmitter;
   handleNavClick() {
+    state.activeItemName = this.name;
     this.selected=true;
```
dgibson666: You need some logic to set selected. This is only applicable for a link. If it's a button controlling a drawer, it doesn't set selected. This is why I'm still seeing the current state change even if nothing in the subnav was selected - it's not really leveraging the state store properly for that.

packages/web-components/src/components/cbp-subnav/cbp-subnav.tsx
```diff
@@ -11,9 +12,15 @@ export class CbpSubNav {
...
   @Prop() accessibilitytext: string = 'Sub-Navigation';
 
+  /** Specifies a name used to associated nav items with subnav items*/
+  @Prop({ reflect: true }) name: string;
+
   /** used to toggle child indenation */
   @Prop({ reflect: true }) flat: boolean;
 
+  /** used to set the current item in the subnav  */
+  @Prop({ reflect: true}) current: boolean;
```
dgibson666: This makes no sense on the parent. Current should only be on a specific subnav-item.

packages/web-components/src/components/cbp-subnav/cbp-subnav.tsx
```diff
@@ -30,6 +37,10 @@ export class CbpSubNav {
     });
   }
   
+  componentDidLoad(){
+    this.current ? state.currentPage = this.name : null;
```
dgibson666: I don't understand why this is behind this.current. Logic doesn't make sense to me.

dgibson666 (commented):
Still doesn't seem to be working quite right. See code comments. Current item is updating when it should only be active. Please console.log your states on any change so we can see what and when things are changing better until it's verified working.

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
bagrub: removed control for this in the app header story

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
bagrub: updated to drawerId & updated default to navDrawer for better semantics

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
bagrub: updated preventDefault() as per discussion this morning using the form field wrapper stories as an example

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
bagrub: updated to use rotate prop, sx still setting the margin-left for icon

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: updated as per above. updated event logic to check if drawer being closed has parent of app header

packages/web-components/src/components/cbp-app-header/store.ts
bagrub: updated to use template literal & condensed to 1 line

packages/web-components/src/components/cbp-app-header/store.ts
bagrub: updated as per above

packages/web-components/src/components/cbp-app-header/store.ts
bagrub: updated as per above. note: added techdebt to move this function to subnav at later date

packages/web-components/src/components/cbp-nav-item/cbp-nav-item.tsx
bagrub: updated to only set the selected state if the navItem has an anchor child

packages/web-components/src/components/cbp-subnav/cbp-subnav.tsx
bagrub: removed

packages/web-components/src/components/cbp-subnav/cbp-subnav.tsx
bagrub: removed

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
```diff
@@ -5,67 +5,180 @@ export default {
...
+  const html = items.map(({ icon, label, name, href, open, children, current }) => {
+      return `<cbp-subnav-item label="${label}" name="${name}" href=${href} ${current? `current=${current}` : ``} ${open? `open=${open}` : ``} >${icon ? `<span slot="cbp-subnav-item-label">${icon} ${label}</span>` : ``} ${children? generateSubnav(children) : ``}</cbp-subnav-item>`;
+  });
+      return html.join('');
+}
+
+
+const Template = ({ open, drawerid, items, sx }) => {  
+  
+  document.addEventListener('click', function(e) { e.preventDefault(); });
```
dgibson666: This need to be more specific because of storybook. I confirmed that once loading this story, all clicks on any subsequently loaded story are canceled as well. You need to queryselect anchors specifically inside of this app header and subnav (by ID ideally) and only preventDefault on those.

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
```diff
@@ -5,67 +5,180 @@ export default {
...
   return ` 
       <cbp-app-header
         ${sx ? `sx=${JSON.stringify(sx)}` : ''}
       >
-        ${generateNavItems(navItems)}
+       ${generateNavItems(items, drawerid)}
+      
+
+        <cbp-drawer
+        ${open ? `open=${open}` : ''}
```
dgibson666: where is `open` coming from? I thought it was removed as a control, so I'm not sure how this isn't complaining.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -33,13 +44,34 @@ export class CbpAppHeader {
...
   }
 
+  updateCurrent(current){
+    state.currentPage = current.getAttribute('name');
+    this.findParent(current);
+  }
+
+  findParent(element){
+    let parent;
+    if( element.parentNode.parentNode.tagName == 'CBP-SUBNAV-ITEM'){
```
dgibson666: this looks kind of hacky. I would use closest(), but remember closest() will also select the current item, so you break out of the self reference by making sure closest() is not the same as this.host (might need to make sure this.host is properly typed then too).

But wait - I'm really confused why you are looking at subnav items in the app-header component. Doesn't seem right or ideal to be tightly coupled this way.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -33,13 +44,34 @@ export class CbpAppHeader {
...
+
   componentWillLoad() {
     this.navItems = Array.from(this.host.querySelectorAll('cbp-nav-item'));
 
     // Attach event listeners to the child navItem
     this.navItems.forEach(navItem => {
       navItem.addEventListener('navClicked', e => this.setActiveNav(e.detail.host));
     }); 
+
+    this.subnavItems = Array.from(this.host.querySelectorAll('cbp-subnav-item')).filter(subnav => subnav.closest('cbp-subnav'))
```
dgibson666: More subnav logic - I'm not sure how this is working or why it's here, but I don't think it belongs in the app header component. It should be in the subnav component.

dgibson666 (commented):
See code comments. Seems like you have some duplicate/misplaced code.

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
bagrub: updated the preventDefault() to only affect the anchors within the appHeader

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
bagrub: removed the Open prop from this file

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: updated to use the closest() & moved to the subnav files

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: Moved to the subnav component

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -8,15 +9,26 @@ export class CbpAppHeader {
...
+  @Listen('drawerClose', { target: 'body'})
+  handleNavDrawerClose(e) {
+    const Subnav = e.target.querySelector('cbp-subnav');
+    // Only update focus and current states if the drawer holds a subnav using state store.
+    if(Subnav?.store == true) {
+      let active = this.host.querySelector(`[name="${state.activeItemName}"] cbp-button > button `) as HTMLButtonElement;
+      active?.focus(); // TechDebt: this needs to be revisited for navigation events that may auto-close the drawer.
+
+      /*
+      state.currentParent
```
dgibson666: remove this comment. I shortened it because the parent should always be present now.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -28,22 +40,23 @@ export class CbpAppHeader {
         link.setAttribute('aria-current', 'true');
       } else {
         navItem.selected = false;
+        //link.hasAttribute('aria-current') ? link.removeAttribute('aria-current') : '';
```
dgibson666: remove this comment - I simplified because removeAttribute doesn't care if it exists or not.

dgibson666 (commented):
I flaged a couple more comments to remove and then it's good to go.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: removed

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: removed

dgibson666 (approved):
good to go

PR #288, opened 2025-05-08
Bugfix/button link gaps

dgibson666:
* Implemented gap for spacing of buttons (2x) and links and tags (1x).
* Updated the following components in relation to these changes:
  * Banner
  * Code Snippet
  * File Input
  * Subnav Item (slotted label with icon)
  * Tabs (icons and badges use gap spacing)
  * USA Banner
* Updated the following Stories in relation to these changes:
  * Card - buttons and title
  * Table (sortable headers)
  * Universal Header (and this content in all templates and archetyopes)
* Updated Changelog with Breaking changes notes.

Anyone using code from these impacted stories would need to update that code on their end. Some of those component updates could impact the downloaded code too like tabs and subnav.

### DIFF
--- a/packages/web-components/CHANGELOG.md
+++ b/packages/web-components/CHANGELOG.md
@@ -6,6 +6,8 @@ The React components are wrappers generated from this package and will share the
 
 ## [unpublished] TBD
 
+* First cut of the `cbp-menu` component.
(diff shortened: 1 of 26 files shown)


bagrub (approved):
LGTM

PR #289, opened 2025-05-12
Bugfix/drawer dark mode updates

dgibson666:
* Tokenize backdrop (light is .3 alpha, dark is .5 alpha).
* Implemented dark mode for drawer.
* Updated User Preferences story and CSS API to support it.
* Bump up close button to 1.25 rem (5x).
* Fixed close button disappearing on resize when persistent option is in use. (test with Passenger Archetype story)


bagrub (approved):
LGTM

PR #290, opened 2025-05-14
Bugfix/pagination and toggle

dgibson666:
* Fix toggle story inheriting from cbp-form-field for multiple toggle story
* Increase tooltip close button size
* Increase the indent of menu items x2 (but remove from default story - this will be used in breadcrumbs)
* Fix dropdown CSS specificity affecting pagination
* Make typography render a span by default (required prop was not enforced and render `undefined` tags in storybook)
* Ran some tests on Common login buttons (commented out in story for easy access)
* Update size of volume slider icon to make it larger.
* Change SubNav accessibilitytext prop to accessibilityText for consistency with other components
* Remove ID on app header and update event listener in story.
* Adjust footer link padding, fix error in story markup, add aria-label to footer nav element.
* Misc cleanup and documentation.


bagrub (approved):
LGTM

PR #291, opened 2025-05-15
Feature/selection card

bagrub:

### DIFF
--- a/packages/web-components/src/components/cbp-card/cbp-card.scss
+++ b/packages/web-components/src/components/cbp-card/cbp-card.scss
@@ -20,10 +20,10 @@
  * @prop  --cbp-flag-card-color-background: var(--cbp-color-gray-cool-60);
  * @prop  --cbp-flag-card-color-background-dark: var(--cbp-color-gray-cool-30);
 
-  * @prop  --cbp-interactive-card-outline-focus: var(--cbp-color-white);
-  * @prop  --cbp-interactive-card-outline-focus-dark: var(--cbp-color-black);
-  * @prop  --cbp-interactive-card-danger-outline-focus: var(--cbp-color-white);
-  * @prop  --cbp-interactive-card-danger-outline-focus-dark: var(--cbp-color-black);
+  * @prop  --cbp-card-outline-focus: var(--cbp-color-white);
+  * @prop  --cbp-card-outline-focus-dark: var(--cbp-color-black);
+  * @prop  --cbp-card-danger-outline-focus: var(--cbp-color-white);
+  * @prop  --cbp-card-danger-outline-focus-dark: var(--cbp-color-black);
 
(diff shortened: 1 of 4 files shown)


packages/web-components/src/components/cbp-card/cbp-card.tsx
```diff
@@ -47,6 +47,20 @@ export class CbpCard {
   }
 
   render() {
+
```
dgibson666: This does not belong in the render. You could potentially be setting a new event listener every render. Probably in componentDidLoad().

Additionally, you are in the card component. You should only be doing it for the individual card - you should almost never do document.querySelector from a component.

packages/web-components/src/components/cbp-card/cbp-card.scss
```diff
@@ -58,6 +67,16 @@
   --cbp-interactive-card-outline-color-focus-dark: var(--cbp-color-black);
   --cbp-interactive-card-danger-outline-color-focus: var(--cbp-color-white);
   --cbp-interactive-card-danger-outline-color-focus-dark: var(--cbp-color-black);
+
```
dgibson666, line 70: can we get rid of the "multi" in the variable name? It should be namespaced to the actual component, which is cbp-card. I think the selected state gives enough context. Even the "interactive" name below could be removed. You wouldn't have a focus if it wasn't interactive...

packages/web-components/src/components/cbp-card/cbp-card.tsx
```diff
@@ -46,6 +46,19 @@ export class CbpCard {
     });
   }
 
+  componentDidLoad(){
+    if(this.interactive == 'selectable'){
+      this.host.addEventListener("click", (e) => {
+        let parent = this.host.querySelector('cbp-card *[slot="cbp-card-title"]');
+        let target = e.target as Element;
```
dgibson666: Is this necessary? Can't you just say `if (!parent.contains(e.target))`?

dgibson666 (commented):
Looks good overall. I think your CSS variables need some cleaning up - they are not sticking to conventions and are more verbose than necessary I think. One other small code comment as well.

packages/web-components/src/components/cbp-card/cbp-card.tsx
bagrub: updated as discussed on call

packages/web-components/src/components/cbp-card/cbp-card.scss
bagrub: update to remove the -multi- css variable name

packages/web-components/src/components/cbp-card/cbp-card.tsx
bagrub: update as requested

dgibson666 (approved):
looks good

PR #295, opened 2025-05-29
Update dialog story

bagrub:
update Dialog story to include args for displaying icon in dialog title & various techdebt resolutions.

*update link to include definition variant including styling, prop, & story args
*update tooltip to an invert context prop to set button state based on parent context or app theme
*update nav-item update variable names of --cbp-nav-item-border-bottom-selected to include color
*update accordion to remove color: inherit !important and use component color var to map to typography tag
*consolidate the structured list item styles to the structured list style sheet


packages/web-components/src/components/cbp-dialog/cbp-dialog.stories.tsx
```diff
@@ -98,6 +102,14 @@ const Template = ({ title, content, color, open, uid, accessibilityText, actions
         variant="heading-dialog"
         divider="underline"
       >
+        ${withIcon ? 
+          `<cbp-icon
```
dgibson666, line 106: Storybook sometimes shows fully rendered code and you have to refresh to get the "source". if you see class="hydrated" you know you got the rendered DOM code and this is wrong to use. All you should need here are the name and size. Size should be the 5x token, per the ticket. If you were matching the text, 1em would do it, but Jeremy's designs rarely work that way.

packages/web-components/src/components/cbp-link/cbp-link.tsx
```diff
@@ -28,7 +28,9 @@ export class CbpLink {
   
   /** Specifies the `download` boolean attribute of the rendered anchor. */
   @Prop() download: boolean;
-  //@Prop({ reflect: true }) variant: 'definition';
+
+  /** Specifies if the 'definition' styling is used in the rendered anchor.  */
```
dgibson666: I think we should have discussed this and should still revisit it. If the definition link brings up a tooltip, then it doesn't belong to link... And if it links out to somewhere else, then it should look like a normal link...

packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
```diff
@@ -1,38 +0,0 @@
-/** todo: update doc
```
dgibson666, line 1: I'm not sure deleting this was necessary. Were you having trouble with the dark mode toggle?

dgibson666 (commented):
Couple things to fix. We should discuss some of the other points though before merging.

packages/web-components/src/components/cbp-dialog/cbp-dialog.stories.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-link/cbp-link.tsx
bagrub: removed changes, as discussed in call definition link is covered as part of the tooltip story so this is not needed

packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
bagrub: discussed in call, confirmed change is good but might be worth looking back over structured list at later time

dgibson666 (approved):
looks good

PR #298, opened 2025-06-03
bugfix/ghost-button

dgibson666:
* Updates to ghost button hover state to be more like outline (to work better on various backgrounds).
* Updates to drawer and code snippet, which were affected.


bagrub (approved):
LGTM

PR #299, opened 2025-06-03
Bugfix/loader success and error icon update

bagrub:
#### update Toast icon color to reassign css var in style sheet


dgibson666 (approved):
looks good

PR #300, opened 2025-06-05
Bugfix/app header stories nav items

dgibson666:
* Refactored nav-item CSS to style any button or link in support of component-based routers.
* Created new default story to be simple nav, moving the complex one with subnav as the second story.
* Some misc cleanup.


bagrub (approved):
LGTM

PR #303, opened 2025-06-10
update archetype and templates to use the cbp-home slot

bagrub:


dgibson666 (approved):
I made a small edit and pushed as well. Approved.

PR #304, opened 2025-06-11
add icon based on color variant for danger state

bagrub:
#### conditionally add icon based on the color prop as per the visual spec

### DIFF
--- a/packages/web-components/src/components/cbp-accordion-item/cbp-accordion-item.scss
+++ b/packages/web-components/src/components/cbp-accordion-item/cbp-accordion-item.scss
@@ -110,10 +110,6 @@ cbp-accordion-item {
       flex-grow: 1;
       justify-content: space-between;
 
-      // & > *{//todo update to use slot="cbp-accordion-item-label"
-      //   gap: var(--cbp-accordion-item-heading-gap);
-      // }
-
       :first-child {
         flex-grow: 1;
       }
--- a/packages/web-components/src/components/cbp-accordion-item/cbp-accordion-item.scss
+++ b/packages/web-components/src/components/cbp-accordion-item/cbp-accordion-item.scss
@@ -110,10 +110,10 @@ cbp-accordion-item {
       flex-grow: 1;
       justify-content: space-between;
 
-      & > *{
-        gap: var(--cbp-accordion-item-heading-gap);
-      }
-      
+      // & > *{//todo update to use slot="cbp-accordion-item-label"
+      //   gap: var(--cbp-accordion-item-heading-gap);
+      // }
+
       :first-child {
         flex-grow: 1;
       }
@@ -121,6 +121,7 @@ cbp-accordion-item {
       [slot="cbp-accordion-item-label"] {
         display: flex;
         align-items: center;
+        gap: var(--cbp-accordion-item-heading-gap);
       }
 
     }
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/cbp-accordion-item/cbp-accordion-item.tsx
```diff
@@ -87,6 +87,7 @@ export class CbpAccordionItem {
           <div id={this.headingId} class="cbp-accordion-item--heading">
             <slot name="cbp-accordion-item-label" />
             {this.label && <cbp-typography tag={this.headingLevel} variant="heading-sm">{this.label}</cbp-typography>}
+            {this.color == 'danger' && <cbp-icon name="triangle-exclamation"></cbp-icon> }
```
dgibson666: This does not belong in the component logic - it should be slotted by the story. Additionally, looking at the "Accordion with slotted label" story, I think we need a default gap of --cbp-space-2x on the div.cbp-accordion-item--heading with a CSS var to override.

bagrub: updated to be in stories.mdx not component & added gap w/ CSS var as requested

packages/web-components/src/components/cbp-accordion-item/cbp-accordion-item.scss
```diff
@@ -108,6 +110,10 @@ cbp-accordion-item {
       flex-grow: 1;
       justify-content: space-between;
 
+      & > *{
```
dgibson666: I'm confused why this is here. the flex context is on .cbp-accordion-item--heading , which I pointed out in my comment. I would test if it's working here, but the icon is no longer showing up alongside the badge in the second story...

packages/web-components/src/components/cbp-accordion/cbp-accordion.stories.tsx
```diff
@@ -57,12 +57,15 @@ export default {
 function generateChildren(items) {
   const html = items.map(({ label, content, open, color, headingLevel }) => {
     return `
-      <cbp-accordion-item
```
dgibson666, line 60: This change has broken the label prop and made them all slotted labels. The second story to show slotted labels no longer shows the icon at all.

There is a separate function for generateSlottedChildren, so maybe there should be some logic to pass it off to that function when color=danger.

dgibson666 (commented):
Need to rethink your implementation and retest.

packages/web-components/src/components/cbp-accordion-item/cbp-accordion-item.scss
bagrub: updated to target the cbp-accordion-item-label slot

packages/web-components/src/components/cbp-accordion/cbp-accordion.stories.tsx
bagrub: updated to use showBadge and slotted generatedSlottedChildren for the warning icon based on paired programming discussion

packages/web-components/src/components/cbp-accordion-item/cbp-accordion-item.scss
```diff
@@ -108,13 +110,18 @@ cbp-accordion-item {
       flex-grow: 1;
       justify-content: space-between;
 
+      // & > *{//todo update to use slot="cbp-accordion-item-label"
```
dgibson666: Remove this commented CSS and you're all good.

dgibson666 (commented):
just remove that commented CSS and you're all good.

dgibson666 (approved):

PR #306, opened 2025-06-17
update ghost primary  and subnav item border left

bagrub:
updated ghost primary hover background color for light mode and update subnav item update hover state

### DIFF
--- a/packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
+++ b/packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
@@ -26,8 +26,13 @@
   --cbp-subnav-item-current-border-color: var(--cbp-color-interactive-active-dark);
   --cbp-subnav-item-current-border-color-dark: var(--cbp-color-interactive-active-light); 
 
+  --cbp-subnav-item-border-color: transparent;
+  --cbp-subnav-item-hover-border-color: var(--cbp-color-interactive-secondary-darker);
+  --cbp-subnav-item-hover-border-color-dark: var(--cbp-color-interactive-secondary-lighter);
+  
   --cbp-subnav-item-indent: var(--cbp-space-4x);
   --cbp-subnav-item-font-size: var(--cbp-font-size-heading-md);
+  --cbp-subnav-item-current-border-width: var(--cbp-border-size-xl);
 }
 
 [data-cbp-theme=light] cbp-subnav[context*=dark]:not([context=light-always]),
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
```diff
@@ -26,8 +26,13 @@
   --cbp-subnav-item-current-border-color: var(--cbp-color-interactive-active-dark);
   --cbp-subnav-item-current-border-color-dark: var(--cbp-color-interactive-active-light); 
 
+  --cbp-subnav-item-border-color: transparent;
+  --cbp-subnav-item-hover-border-color: var(--cbp-color-interactive-secondary-darker);
+  --cbp-subnav-item-hover-border-color-dark: var(--cbp-color-interactive-secondary-lighter);
+  
   --cbp-subnav-item-indent: var(--cbp-space-4x);
   --cbp-subnav-item-font-size: var(--cbp-font-size-heading-md);
+  --cbp-subnav-item-current-border-width: var(--cbp-border-size-xl);
```
dgibson666: I don't think this needs to be exposed, personally.

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
```diff
@@ -118,7 +130,7 @@ cbp-subnav-item {
     &:first-of-type a{
       justify-content: start;
       font-weight: var(--cbp-font-weight-medium);
-      padding-left: var(--cbp-subnav-item-indent);
+      padding-left: calc(var(--cbp-subnav-item-indent) - var(--cbp-subnav-item-current-border-width));
```
dgibson666: I would give them all the same border and just toggle the color. That's how we typically do it to prevent jumpiness or this sort of workaround. It also looks visually weird currently. I can demo.

dgibson666 (commented):
probably need to discuss the subnav real quick.

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
bagrub: removed this var

packages/web-components/src/components/cbp-subnav-item/cbp-subnav-item.scss
bagrub: updated to use the button background color to resolve visual diff seen in visual review

dgibson666 (approved):
Looks good. I'm not 100% sure on the focus state, but we can revisit if Jeremy doesn't like it.

PR #307, opened 2025-06-23
Feature/responsive breadcrumbs

dgibson666:
* Updated resize observer for actual use and added docs.
* Added responsive behavior to breadcrumbs using resize observer and menu.
* Fixed debounce function.
* Added vscode output target to Stencil config.
* Cleanup and tweaks to USA Banner.
* Updated other package and component docs.

I wanted to get debounce working at the resize observer component level (exposed to the end user), but had trouble. I left some comments in related to that so I can revisit it. It works for this case and I wanted to get this out the door.


bagrub (approved):
LGTM

PR #317, opened 2025-07-11
Bugfix/a11y fixes 20250630

dgibson666:
* The main issue is changing the `pressed` and `expanded` properties of `cbp-button` to a string union of `"true"|"false"` from a Boolean. This is due to the fact that JSX does not render the `aria-expanded` or `aria-pressed` at all with a false value. When `aria-expanded="false"` is not rendered, the statefulness of the control is not conveyed to screen readers and updates to the state are only spoken when the value is true.
* Resulting component updates impacted by this change:
  * Accordion - removed aria-label (verbose), updated aria-describedby to aria-labelledby, outputting expanded=false improves accessibility and feedback on state changes/user interaction.
  * Breadcrumb (responsive menu) - added expanded="false” to rendered menu button.
  * Chip - accepts a boolean, but converts to a string - fixed by defaulting to false, which improves accessibility and feedback on state changes/user interaction.
  * Expand - removed aria-label (verbose), updated aria-describedby to aria-labelledby, outputting expanded=false improves accessibility and feedback on state changes/user interaction.
  * Code Snippet - updated to output expanded as a string, defaulting to “false,” which improves accessibility and feedback on state changes/user interaction.
  * Menu - set button expanded="false” on load, toggle it in the @watch on the open prop - improves accessibility and feedback on state changes/user interaction. However, focus was not moving properly to the first menu item when opened, and wasn’t read by the screen reader until the delay was increased to 500ms.
  * Segmented Button Group - updated to work with updated Button pressed prop (“true”|”false” string union). Updated so single select groups toggle pressed=“true” on user interaction but cannot toggle them off (like a radio button). Groups with multiple=true can toggle pressed=true/false on each user interaction.
  * Subnav Item - set expanded as a string on the expand button. Also generate a UID to apply as an `id` on the main subnav item control (cbp-button), which associates a label for the expand button (when present) via aria-labelledby.
  * Table - pressed updated internally. Set pressed (both true and false) on initial load as part of makeSortable() function.
  * USA Banner - updated to set expanded as a string, open prop defaulted to false.

Dialog control, Drawer control, and Nav Item (Drawer Control) do not need `aria-expanded` and removing the auto-toggling of expanded (when there’s an aria-controls on the Button component) fixes the issue where these controls are left in aria-expanded=“true” state when a dialog or drawer is closed.

### DIFF
--- a/packages/web-components/CHANGELOG.md
+++ b/packages/web-components/CHANGELOG.md
@@ -4,6 +4,17 @@ This CHANGELOG.md tracks the updates to the web components package of the CBP de
 
 The React components are wrappers generated from this package and will share the same changes. Projects using React 19 may use the native web components without React wrappers.
 
+## [unreleased] TBD
+
+* BREAKING: Updated the `cbp-button` properties of `expanded` and `pressed` to a string union of `"true" | "false"` rather than a true Boolean value.
+  * This is due to the fact that JSX does not render the `aria-expanded` or `aria-pressed` at all with a false value.
+  * When `aria-expanded="false"` is not rendered, the statefulness of the control is not conveyed to screen readers and updates to the state are only spoken when the value is true.
+  * These properties are primarily used internally and this change is unlikely to cause significant problems.
+  * Components updated to work with this change include: Accordion Item, Expand, Menu, Code Snippet, Subnav Item, and Segmented Button Group.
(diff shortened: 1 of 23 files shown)


bagrub (approved):
LGTM

PR #318, opened 2025-07-21
Bugfix/a11y fixes 20250718

dgibson666:
* Gave the universal header an explicit role.
* Removed the `header` tags from the templates/archetype - App header is now sticky as intended.
* Added `aria-invalid` to form field groups (e.g., checklist, radiolists, etc.).
* Fixed aria attributes and added `aria-described` by to dropdown control.
* Other small docs and code tweaks with no functionality changes.


bagrub (approved):
LGTM

PR #319, opened 2025-07-22
Time input story for form field wrapper

bagrub:
Add time input story to the form field to match spec


dgibson666 (approved):
I see some minor things, but go ahead and merge for discussion and we can cleanup in the normal iteration process.

PR #320, opened 2025-07-22
Bugfix/a11y fixes 20250721

dgibson666:
* Update skipnav to set focus rather than navigate.
* Fix HTML heading line-height and add scroll-margin to account for sticky app header.
* Added explicit aria-labelledby to form field groups.
* Added accessible text to the lock icon in USA Banner  for better reading.
* Updated changelog with latest fixes.


bagrub (approved):
LGTM

PR #325, opened 2025-07-25
Adding name pairing to stories to set state for app header and subnav…

bagrub:
Update story props to include matching name values on navItems to integrate with the store state between subnav and appHeader

### DIFF
--- a/packages/web-components/src/stories/templates.stories.tsx
+++ b/packages/web-components/src/stories/templates.stories.tsx
@@ -24,19 +24,23 @@ export default {
     navItems: [
       {
         label: 'Application Name',
+        name: 'Application Name',
         href: './?path=/story/components-application-header--application-header#',
         current: true
       },
       {
         label: 'Nav Item 1',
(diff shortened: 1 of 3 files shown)


packages/web-components/src/stories/templates.stories.tsx
```diff
@@ -209,7 +213,7 @@ const Internal2ColumnTemplate = ({ isLoggedIn, username, navItems, contentGridSi
         ${generateNavItems(navItems)}
       </cbp-app-header>
       
-      ${renderDrawer(navItems, 'appHeaderDrawer', false)}
+      ${renderDrawer(navItems, 'appHeaderDrawer', true)}
```
dgibson666, line 216: I only see this change twice, but there are 3 template stories. Add it to the cards template as well.

bagrub: updated to include card template layout

dgibson666 (approved):

PR #327, opened 2025-08-01
Accessibility update bundle

bagrub:
Various updates found in the recent Accessibility Audit:
*Update Dialog story to have button text "Open Dialog" instead of a hamburger icon
*Update Panel:
  *remove role="complimentary " from stories
  *update header text to "Drawer header" instead of "Panel Header"
*Update Action Menu story to add button label of "Action Menu"

### DIFF
--- a/packages/web-components/src/components/cbp-menu/cbp-menu.stories.tsx
+++ b/packages/web-components/src/components/cbp-menu/cbp-menu.stories.tsx
@@ -75,6 +75,7 @@ const ActionsMenuTemplate = ({ position, context, sx }) => {
         accessibilityText="Actions"
         target-prop="open"
         controls="menuId"
+        label="Action Menu"
       >
         <cbp-icon name="ellipsis-vertical"></cbp-icon>
       </cbp-button>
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/cbp-menu/cbp-menu.stories.tsx
```diff
@@ -75,6 +75,7 @@ const ActionsMenuTemplate = ({ position, context, sx }) => {
         accessibilityText="Actions"
         target-prop="open"
         controls="menuId"
+        label="Action Menu"
```
dgibson666: label is not a prop on cbp-button. It looks like the accessible label wasn't working because accessibilityText="Actions" should be dash case in the story markup. Definitely don't need the word "menu" in the label because it's announced as a menu already.

dgibson666 (commented):
Just 1 comment in code.

packages/web-components/src/components/cbp-menu/cbp-menu.stories.tsx
bagrub: removed label & updated to 'accessibility-text'

dgibson666 (approved):

PR #329, opened 2025-08-13
update to address various accessibility issues

bagrub:
update to address below issues found in accessibility testing:

*Tooltip: add accessibility label to close button
*Loader: when indeterminate loader state is changed to success the value is set to be equal to the max 
*Templates: add label to the footers
*Archetype: remove 'role= complimentary' from the filter panel
*Code Snippet: add story for block variant

### DIFF
--- a/packages/web-components/src/components/cbp-loader/cbp-loader.tsx
+++ b/packages/web-components/src/components/cbp-loader/cbp-loader.tsx
@@ -76,7 +76,7 @@ export class CbpLoader {
       statusIndicator = Math.round((this.value / this.max) * 100) + "%"
     }
 
-    if(this.success && !this.determinate){
+    if(this.success){
       this.value = this.max;
     }
 
--- a/packages/web-components/src/components/cbp-loader/cbp-loader.tsx
+++ b/packages/web-components/src/components/cbp-loader/cbp-loader.tsx
@@ -76,6 +76,10 @@ export class CbpLoader {
       statusIndicator = Math.round((this.value / this.max) * 100) + "%"
     }
 
+    if(this.success && !this.determinate){
+      this.value = this.max;
+    }
+
     return (
       <Host>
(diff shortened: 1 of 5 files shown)


packages/web-components/src/components/cbp-loader/cbp-loader.tsx
```diff
@@ -76,6 +76,10 @@ export class CbpLoader {
       statusIndicator = Math.round((this.value / this.max) * 100) + "%"
     }
 
+    if(this.success && !this.determinate){
```
dgibson666: shouldn't this be the same whether determinate or indeterminate?

packages/web-components/src/stories/templates.stories.tsx
```diff
@@ -118,7 +118,9 @@ const InternalTemplate = ({ isLoggedIn, username, hashid, navItems }) => {
           </main>
       </cbp-container>
 
-      <cbp-footer>
+      <cbp-footer
+        label='Footer'
```
dgibson666: Check the footer story where it is correct. I'm pretty sure this was meant to be an aria-label on the nav element. `label` is not a valid attribute.

dgibson666 (commented):
See code comments

packages/web-components/src/components/cbp-loader/cbp-loader.tsx
bagrub: removing the !determinate. based on initial comment from the accessibility test i assumed we only wanted this applied to the indeterminate variant

packages/web-components/src/stories/templates.stories.tsx
bagrub: update to apply aria-label on the nav element instead of the cbp-footer as per the example in the footer story

dgibson666 (approved):

dgibson666 (approved):

PR #331, opened 2025-08-15
Feature/theme switcher

dgibson666:
* Add missing API docs for Toggle.
* Add a watch on checked to make the property reactive and synced with native input.
* Hook up dark mode/theme toggle in templates and archetype.


bagrub (approved):
LGTM

PR #332, opened 2025-08-18
update to address techdebt on toggle component

bagrub:
Address tech debt on toggle component mostly to rename variables for sake of consistency

### DIFF
--- a/packages/web-components/src/components/cbp-toggle/cbp-toggle.scss
+++ b/packages/web-components/src/components/cbp-toggle/cbp-toggle.scss
@@ -178,21 +178,17 @@ cbp-toggle {
   }
 
   span:last-child {
-    flex-basis: var(--cbp-toggle-label-text-width);
+    flex-basis: var(--cbp-toggle-status-text-width);
     text-overflow: ellipsis;
     overflow: hidden;
     white-space: nowrap;
   }
 
   /** 'On' States */
-  &[checked] input[type=checkbox] {
+  &[checked]:not([disabled]) input[type=checkbox] {
     --cbp-toggle-color-bg: var(--cbp-toggle-color-bg-selected);
     --cbp-toggle-circle-color: var(--cbp-toggle-circle-selected-color);
     --cbp-toggle-circle-color-border: var(--cbp-toggle-circle-selected-border-color);
-
-    &:before {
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/cbp-toggle/cbp-toggle.scss
```diff
@@ -180,13 +172,13 @@ cbp-toggle {
...
-      background-color: var(--cbp-toggle-color-circle);
-      border: var(--cbp-toggle-circle-border-width) solid var(--cbp-toggle-color-circle-border);
+      background-color: var(--cbp-toggle-circle-color);
+      border: var(--cbp-toggle-circle-border-width) solid var(--cbp-toggle-circle-color-border);
     }
   }
 
   span:last-child {
-    flex-basis: var(--cbp-status-text-width);
+    flex-basis: var(--cbp-toggle-label-text-width);
```
dgibson666: This isn't using the right variable. It should be --cbp-toggle-status-text-width, which is properly named (the label would be the first child).

dgibson666 (commented):
The renaming looks good other than 1 code comment using the wrong variable name.

Also, while testing I noticed that a disabled control that is checked is not styled correctly - the checked color is overriding the disabled styles. Can you fix this as well as add a "checked" key to the story object for easier toggling.

packages/web-components/src/components/cbp-toggle/cbp-toggle.scss
bagrub: updated to correct variable name

packages/web-components/src/components/cbp-toggle/cbp-toggle.scss
```diff
@@ -239,19 +231,19 @@ cbp-toggle {
...
 
-      --cbp-toggle-color-circle: var(--cbp-toggle-color-circle-focus);
-      --cbp-toggle-color-circle-border: var(--cbp-toggle-color-circle-border-focus);
+      --cbp-toggle-circle-color: var(--cbp-toggle-circle-color-focus);
+      --cbp-toggle-circle-color-border: var(--cbp-toggle-circle-color-border-focus);
     }
   }
 
   /** Disabled */
   &[disabled] {
```
dgibson666: Sorry, I'm just catching this. All of these instances should be `:disabled` not `[disabled]`. You can disable the entire group (in the second story) without placing disabled on the individual toggle, which means its style isn't showing the disabled state.

dgibson666 (approved):
looks like everything's working now.

PR #333, opened 2025-08-20
Feature/segmented button value

dgibson666:
* Update Segmented Button Group to render a hidden input when a name is specified, so that it can pass a form value.
* Minor button fix to reflect pressed and expanded props to attributes.
* Update the Time pattern segmented button group with a name and proper button values.
* Updated changelog.


bagrub (approved):
LGTM

PR #334, opened 2025-08-21
Bugfix/card updates

bagrub:
*Addition of a story so to show Banner with Decision Card to support App directory usage of card
*Add Radio button example to Interactive card story
*Add Interactive Radio Card story to show the usage of a radio list of interactive cards
*Update Decision Card story to accept an arg 'actionsFill' to support the button fill variations (solid, outline, ghost)

### DIFF
--- a/packages/web-components/src/components/cbp-card/cbp-card.scss
+++ b/packages/web-components/src/components/cbp-card/cbp-card.scss
@@ -218,21 +218,21 @@ cbp-card {
         --cbp-button-min-height: var(--cbp-space-11x);
         width: 100%;
 
-        &[fill=ghost][color=primary] button:not(:hover):not(:focus),
-        &[fill=ghost][color=secondary] button:not(:hover):not(:focus),
-        &[fill=ghost][color=tertiary] button:not(:hover):not(:focus)
+        // &[fill=ghost][color=primary] button:not(:hover):not(:focus),
+        // &[fill=ghost][color=secondary] button:not(:hover):not(:focus),
+        // &[fill=ghost][color=tertiary] button:not(:hover):not(:focus)
+        
+        &[fill=ghost][color] button:not(:hover):not(:focus)
         {
           --cbp-button-color-border: var(--cbp-banner-card-color-border);
-          --cbp-button-border-width: var(--cbp-border-size-md) 0 var(--cbp-border-size-md) 0;//TODO: suppress border width for inbetween exlude first & last
+          --cbp-button-border-width: var(--cbp-border-size-md) 0 var(--cbp-border-size-md) 0;
         }
 
 
         &:first-child {
           --cbp-button-border-radius: 0 0 0 var(--cbp-border-radius-softer);
 
-          &[fill=ghost][color=primary] button:not(:hover):not(:focus),
-          &[fill=ghost][color=secondary] button:not(:hover):not(:focus),
-          &[fill=ghost][color=tertiary] button:not(:hover):not(:focus){
+          &[fill=ghost][color] button:not(:hover):not(:focus){
             --cbp-button-border-width: var(--cbp-border-size-md) 0 var(--cbp-border-size-md) var(--cbp-border-size-md);
           }
         }
@@ -240,15 +240,17 @@ cbp-card {
         &:last-child {
           --cbp-button-border-radius: 0 0 var(--cbp-border-radius-softer) 0;
 
-          &[fill=ghost][color=primary] button:not(:hover):not(:focus),
-          &[fill=ghost][color=secondary] button:not(:hover):not(:focus),
-          &[fill=ghost][color=tertiary] button:not(:hover):not(:focus){
+          &[fill=ghost][color] button:not(:hover):not(:focus){
             --cbp-button-border-width: var(--cbp-border-size-md) var(--cbp-border-size-md) var(--cbp-border-size-md) 0;
           }
         }
-
+        
         &:first-child:last-child {
           --cbp-button-border-radius: 0 0 var(--cbp-border-radius-softer) var(--cbp-border-radius-softer);
+        
+          &[fill=ghost][color] button:not(:hover):not(:focus){
+            --cbp-button-border-width: var(--cbp-border-size-md);
+          }
         }
   
         button, a {
@@ -415,6 +417,7 @@ cbp-card {
       &:not(disabled) .cbp-card-body{
         outline: 2px solid var(--cbp-card-outline-color-focus); 
         outline-offset: -4px;
+        border-radius: var(--cbp-border-radius-softer);
       }
     }
 
--- a/packages/web-components/src/components/cbp-card/cbp-card.scss
+++ b/packages/web-components/src/components/cbp-card/cbp-card.scss
@@ -209,7 +209,7 @@ cbp-card {
       border-top-right-radius: var(--cbp-border-radius-softer);
       border-bottom: 0;
     }
-
+  }
      [slot=cbp-card-actions] {
       display: flex;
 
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/cbp-card/cbp-card.scss
```diff
@@ -218,12 +218,33 @@ cbp-card {
         --cbp-button-min-height: var(--cbp-space-11x);
         width: 100%;
 
+        &[fill=ghost][color=primary] button:not(:hover):not(:focus),
```
dgibson666: I don't think you need the value of the color, which requires 3 selectors. If you can do without the button selector, that would be more ideal as well.

packages/web-components/src/components/cbp-card/cbp-card.scss
```diff
@@ -218,12 +218,33 @@ cbp-card {
...
+        {
+          --cbp-button-color-border: var(--cbp-banner-card-color-border);
+          --cbp-button-border-width: var(--cbp-border-size-md) 0 var(--cbp-border-size-md) 0;//TODO: suppress border width for inbetween exlude first & last
+        }
+
+
         &:first-child {
           --cbp-button-border-radius: 0 0 0 var(--cbp-border-radius-softer);
+
+          &[fill=ghost][color=primary] button:not(:hover):not(:focus),
```
dgibson666: I don't think you need the value of the color, which requires 3 selectors.

packages/web-components/src/components/cbp-card/cbp-card.scss
```diff
@@ -218,12 +218,33 @@ cbp-card {
...
+          &[fill=ghost][color=secondary] button:not(:hover):not(:focus),
+          &[fill=ghost][color=tertiary] button:not(:hover):not(:focus){
+            --cbp-button-border-width: var(--cbp-border-size-md) 0 var(--cbp-border-size-md) var(--cbp-border-size-md);
+          }
         }
   
         &:last-child {
           --cbp-button-border-radius: 0 0 var(--cbp-border-radius-softer) 0;
+
+          &[fill=ghost][color=primary] button:not(:hover):not(:focus),
```
dgibson666: I don't think you need the value of the color, which requires 3 selectors.

dgibson666 (commented):
A few code comments, but some other concerns are:
* Radio cards - add margin-block-end: var(--cbp-space-1x) via sx. (verify with Jeremy - maybe 2x, since these are large colored blocks)
* The visibility of the selected radio card is really poor and should be verified with Jeremy or an official spec sheet.
* We should not have 3 layers of borders on selected, focused cards. I'm seeing the inset outline, the background color, then a gray border. Usually the border changes to the bg color on these interactive states. Then you also need to verify the inset amount - typically the inset is the same size as the border because of the border color change.
* The border radius of the inset outline should match the actual border. It seems these are being applied at different levels, so you just need to make sure it's applied where the outline is as well.
* Decision card story with a single action ghost button is cutting of the left border. The other variants appear fine.

packages/web-components/src/components/cbp-card/cbp-card.scss
bagrub: reduced selector to &[fill=ghost][color]... as discussed on call

packages/web-components/src/components/cbp-card/cbp-card.scss
bagrub: reduced selector to &[fill=ghost][color]... as discussed on call

packages/web-components/src/components/cbp-card/cbp-card.scss
bagrub: reduced selector to &[fill=ghost][color]... as discussed on call

packages/web-components/src/components/cbp-card/cbp-card.scss
```diff
@@ -415,6 +417,7 @@ cbp-card {
       &:not(disabled) .cbp-card-body{
         outline: 2px solid var(--cbp-card-outline-color-focus);
```
dgibson666: 2px should use a border size token

packages/web-components/src/components/cbp-card/cbp-card.scss
```diff
@@ -218,37 +218,39 @@ cbp-card {
...
         width: 100%;
 
-        &[fill=ghost][color=primary] button:not(:hover):not(:focus),
-        &[fill=ghost][color=secondary] button:not(:hover):not(:focus),
-        &[fill=ghost][color=tertiary] button:not(:hover):not(:focus)
+        // &[fill=ghost][color=primary] button:not(:hover):not(:focus),
+        // &[fill=ghost][color=secondary] button:not(:hover):not(:focus),
+        // &[fill=ghost][color=tertiary] button:not(:hover):not(:focus)
+        
+        &[fill=ghost][color] button:not(:hover):not(:focus)
```
dgibson666: did you try these without the button descendant selector?

packages/web-components/src/components/cbp-card/cbp-card.scss
```diff
@@ -415,6 +417,7 @@ cbp-card {
       &:not(disabled) .cbp-card-body{
         outline: 2px solid var(--cbp-card-outline-color-focus); 
         outline-offset: -4px;
```
dgibson666: I always do this with tokens. calc(var(--cbp-space-1x) * -1)

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
```diff
@@ -476,7 +476,7 @@ const InteractiveRadioListTemplate = ({ title, color, disabled, bodyText, withIc
         ${disabled ? `disabled=${disabled}` : ''}
         ${color ? `color=${color}` : ''}
         ${context && context != 'light-inverts' ? `context=${context}` : ''}
-        ${sx ? `sx='${JSON.stringify(sx)}'` : ''}
+        sx={"margin-bottom":"var(--cbp-space-1x)"}
```
dgibson666: if you're already passing this as a control, set it as a value in the story rather than hardcoding it. Now you've got a dead control in your story.

dgibson666 (commented):
See code comments. Looks good overall.

packages/web-components/src/components/cbp-card/cbp-card.scss
bagrub: updated to use border token

packages/web-components/src/components/cbp-card/cbp-card.scss
bagrub: refactored so that the border width doesn't use the descendant selector, the border color needs the descendant to override dark mode from the button

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
bagrub: update to use story args instead of the hardcoding

packages/web-components/src/components/cbp-card/cbp-card.scss
bagrub: updated to use border token

packages/web-components/src/components/cbp-card/cbp-card.scss
```diff
@@ -218,16 +218,37 @@ cbp-card {
         --cbp-button-min-height: var(--cbp-space-11x);
         width: 100%;
 
+        &[fill=ghost][color]{
+          --cbp-button-border-width: var(--cbp-border-size-md) 0 var(--cbp-border-size-md) 0;
+        
+          &  button:not(:hover):not(:focus){
```
dgibson666: you don't need the & to do a descendant selector. You just use nesting.

packages/web-components/src/components/cbp-card/cbp-card.scss
```diff
@@ -240,7 +261,7 @@ cbp-card {
     &[color=danger] .cbp-card-body { 
       border-color: var(--cbp-color-danger-base);
     }
-  }
+  // }
```
dgibson666: remove?

dgibson666 (commented):
Looks good visually, but see code comments. Also, on like 513 of the CSS, you have a selector that is wrong:     &:not(disabled) .cbp-card-body - the disabled should be in square brackets.

Finally, your story is outputting the disabled prop wrong - it should not have a value. Just be present when true.

packages/web-components/src/components/cbp-card/cbp-card.scss
bagrub: removed

packages/web-components/src/components/cbp-card/cbp-card.scss
bagrub: removed

packages/web-components/src/components/cbp-card/cbp-card.tsx
```diff
@@ -50,6 +50,15 @@ export class CbpCard {
 
   componentDidLoad(){
     if(this.interactive == 'selectable' || this.interactive == 'radio'){
+
+      if(this.disabled){
+        let input = this.host.querySelector('cbp-checkbox input[type="checkbox"], cbp-radio input[type="radio"]') as HTMLInputElement;
+        input.disabled = true;
+        console.log('Card disabled input');
```
dgibson666: delete console.logs

packages/web-components/src/components/cbp-card/cbp-card.tsx
```diff
@@ -50,6 +50,15 @@ export class CbpCard {
 
   componentDidLoad(){
     if(this.interactive == 'selectable' || this.interactive == 'radio'){
+
+      if(this.disabled){
+        let input = this.host.querySelector('cbp-checkbox input[type="checkbox"], cbp-radio input[type="radio"]') as HTMLInputElement;
```
dgibson666, line 56: I'll let this pass for now, but doing it in componentDidLoad means it's not reactive to changes.

dgibson666 (commented):
see code comments

packages/web-components/src/components/cbp-card/cbp-card.tsx
bagrub: removed

packages/web-components/src/components/cbp-card/cbp-card.tsx
bagrub: adding techdebt comment to flag this

dgibson666 (approved):
looks good

PR #335, opened 2025-08-26
Global Search example in app header

bagrub:
Update app header with a story including the global search according to mockup

### DIFF
--- a/packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
+++ b/packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
@@ -40,17 +40,14 @@ cbp-app-header{
   top: calc(-1 * var(--cbp-space-half-x));
 
   display: flex;
+  align-items: center;
+  gap: var(--cbp-space-4x);
 
   nav{
     display: flex;
     align-items: center;
     align-content: stretch;
     width: max-content;
-  }
-
-  search{
-    align-self: center;
-    margin-left: var(--cbp-space-4x);
   }
 
   cbp-resize-observer{
@@ -63,7 +60,7 @@ cbp-app-header{
     width: 100%
   }
 
-  #global-search-field{
+  #cbp-app-header-search{
     position: absolute;
     top: 0;
     left: 0;
@@ -79,7 +76,7 @@ cbp-app-header{
       outline-offset: calc( -2 * var(--cbp-border-size-md));
     }
 
-    form{
+    > [slot="cbp-global-search"]{
       display: inline-flex;
       width: 100%;
 
@@ -99,7 +96,8 @@ cbp-app-header{
       }
     }
     
-    form > *:not(input),
+    //*Spacing for buttons associated to search form (submit, close, etc)
+    > [slot="cbp-global-search"] > *:not(input),
     > cbp-button{
       align-content: center;
       margin-left: var(--cbp-space-4x);
--- a/packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
+++ b/packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
@@ -98,11 +98,8 @@ cbp-app-header{
       div{
         display: flex;  
         flex-direction: row-reverse;
-      
-        cbp-button{
-          align-content: center;
-          margin-left: var(--cbp-space-4x);
-        }  
+        gap: var(--cbp-space-4x);
+        align-items: center;
       }
     }
   }
--- a/packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
+++ b/packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
@@ -76,7 +76,7 @@ cbp-app-header{
       outline-offset: calc( -2 * var(--cbp-border-size-md));
     }
 
-    > [slot="cbp-global-search"]{
+      > form{
       display: inline-flex;
       width: 100%;
 
@@ -94,14 +94,16 @@ cbp-app-header{
           outline: none;
         }
       }
+
+      div{
+        display: flex;  
+        flex-direction: row-reverse;
+      
+        cbp-button{
+          align-content: center;
+          margin-left: var(--cbp-space-4x);
+        }  
+      }
     }
-    
-    //*Spacing for buttons associated to search form (submit, close, etc)
-    > [slot="cbp-global-search"] > *:not(input),
-    > cbp-button{
-      align-content: center;
-      margin-left: var(--cbp-space-4x);
-    }
-  
   }
 }
--- a/packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
+++ b/packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
@@ -79,23 +79,28 @@ cbp-app-header{
       outline-offset: calc( -2 * var(--cbp-border-size-md));
     }
 
-    input{
-      flex-grow: 1;
-      border: none;
-      background-color: var(--cbp-app-header-color-background);
-    
-      &::placeholder{
-        color: var(--cbp-app-header-search-color-placeholder);
-      }
+    form{
+      display: inline-flex;
+      width: 100%;
 
-      &:focus-visible{
-        outline: none;
+      input{
+        flex-grow: 1;
+        border: none;
+        background-color: var(--cbp-app-header-color-background);
+      
+        &::placeholder{
+          color: var(--cbp-app-header-search-color-placeholder);
+        }
+
+        &:focus-visible{
+          outline: none;
+        }
       }
-    }
-     
-    & > *{
-      align-content: center;
-      margin-left: var(--cbp-space-4x);
+      
+      & > *{
+        align-content: center;
+        margin-left: var(--cbp-space-4x);
+      }
     }
   }
 }
--- a/packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
+++ b/packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
@@ -68,6 +68,7 @@ cbp-app-header{
     right: 0;
 
     display: flex;  
+    width: 100%;
     padding: 0 var(--cbp-responsive-spacing-outer);
     background-color: var(--cbp-app-header-color-background);
 
@@ -76,31 +77,30 @@ cbp-app-header{
       outline-offset: calc( -2 * var(--cbp-border-size-md));
     }
 
-      > form{
-      display: inline-flex;
-      width: 100%;
+      // > form{
+      // display: inline-flex;
 
-      input{
-        flex-grow: 1;
-        border: none;
-        background-color: var(--cbp-app-header-color-background);
-      
-        &::placeholder{
-          font-style: italic;
-          color: var(--cbp-app-header-search-color-placeholder);
-        }
-
-        &:focus-visible{
-          outline: none;
-        }
+    input{
+      flex-grow: 1;
+      border: none;
+      background-color: var(--cbp-app-header-color-background);
+    
+      &::placeholder{
+        font-style: italic;
+        color: var(--cbp-app-header-search-color-placeholder);
       }
 
-      div{
-        display: flex;  
-        flex-direction: row-reverse;
-        gap: var(--cbp-space-4x);
-        align-items: center;
+      &:focus-visible{
+        outline: none;
       }
     }
+
+    div{
+      display: flex;  
+      flex-direction: row-reverse;
+      gap: var(--cbp-space-4x);
+      align-items: center;
+    }
+  
   }
 }
--- a/packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
+++ b/packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
@@ -89,6 +89,7 @@ cbp-app-header{
         background-color: var(--cbp-app-header-color-background);
       
         &::placeholder{
+          font-style: italic;
           color: var(--cbp-app-header-search-color-placeholder);
         }
 
@@ -96,11 +97,13 @@ cbp-app-header{
           outline: none;
         }
       }
-      
-      & > *{
-        align-content: center;
-        margin-left: var(--cbp-space-4x);
-      }
     }
+    
+    form > *:not(input),
+    > cbp-button{
+      align-content: center;
+      margin-left: var(--cbp-space-4x);
+    }
+  
   }
 }
--- a/packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
+++ b/packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
@@ -6,6 +6,12 @@
 :root {
   --cbp-app-header-color-background: var(--cbp-color-white);
   --cbp-app-header-color-background-dark: var(--cbp-color-base-neutral-dark);
+
+  --cbp-app-header-search-focus-outline: var(--cbp-color-interactive-focus-dark);
+  --cbp-app-header-search-focus-outline-dark: var(--cbp-color-interactive-focus-light);
+
+  --cbp-app-header-search-color-placeholder: var(--cbp-color-text-dark);
+  --cbp-app-header-search-color-placeholder-dark: var(--cbp-color-text-light);
 }
 
 /* 
@@ -15,6 +21,8 @@
  */
 [data-cbp-theme=dark] cbp-app-header {
   --cbp-app-header-color-background: var(--cbp-app-header-color-background-dark);
+  --cbp-app-header-search-focus-outline: var(--cbp-app-header-search-focus-outline-dark);
+  --cbp-app-header-search-color-placeholder: var(--cbp-app-header-search-color-placeholder-dark);
 }
 
 cbp-app-header{
@@ -27,7 +35,7 @@ cbp-app-header{
   position: sticky;
   top: calc(-1 * var(--cbp-space-half-x));
 
-  display: block;
+  display: flex;
 
   nav{
     display: flex;
@@ -36,9 +44,54 @@ cbp-app-header{
     width: max-content;
   }
 
+  #global-search-toggle{
+    align-self: center;
+    margin-left: var(--cbp-space-4x);
+  }
+
+  cbp-resize-observer{
+    flex-grow: 1;
+    width: calc(100% - (var(--cbp-button-width) + ( 2 * var(--cbp-button-border-width))));
+  }
+
   .cbp-app-header-responsive{
     align-items: center;
     justify-content: space-between;
     width: 100%
+  }
+
+  search{
+    position: absolute;
+    top: 0;
+    left: 0;
+    bottom: 0;
+    right: 0;
+
+    display: flex;  
+    padding: 0 var(--cbp-responsive-spacing-outer);
+    background-color: var(--cbp-app-header-color-background);
+
+    &:focus-within{
+      outline:  var(--cbp-border-size-md) solid var(--cbp-app-header-search-focus-outline);
+      outline-offset: calc( -2 * var(--cbp-border-size-md));
+    }
+
+    input{
+      flex-grow: 1;
+      border: none;
+      background-color: var(--cbp-app-header-color-background);
+    
+      &::placeholder{
+        color: var(--cbp-app-header-search-color-placeholder);
+      }
+
+      &:focus-visible{
+        outline: none;
+      }
+    }
+     
+    & > *{
+      align-content: center;
+    }
   }
 }
--- a/packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
+++ b/packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
@@ -1,14 +1,18 @@
 /** 
  * @prop --cbp-app-header-color-background: var(--cbp-color-white) 
  * @prop --cbp-app-header-color-background-dark: var(--cbp-color-base-neutral-dark);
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
```diff
@@ -36,9 +44,54 @@ cbp-app-header{
...
   }
 
+  #global-search-toggle{
+    align-self: center;
+    margin-left: var(--cbp-space-4x);
+  }
+
+  cbp-resize-observer{
+    flex-grow: 1;
+    width: calc(100% - (var(--cbp-button-width) + ( 2 * var(--cbp-button-border-width))));
```
dgibson666: You shouldn't need to do this at all. If I uncheck it in the browser, it still works fine.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -149,6 +181,36 @@ export class CbpAppHeader {
...
+                color='secondary'
+                variant='square'
+                onClick={this.toggleSearch}
+              >
+                <cbp-icon name='magnifying-glass'></cbp-icon>
+              </cbp-button>
+            }
+
+        {this.globalSearch &&
+          <div>
```
dgibson666: why is this nesting div needed

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -149,6 +181,36 @@ export class CbpAppHeader {
           </nav>
         </cbp-resize-observer>
         
+        {this.globalSearch &&
```
dgibson666: So my intent was to have the `search` tag around this button. If the landmark is hidden, then it's inaccessible. No one will know there's a site search landmark until it's open. You can do this and target a nested wrapper with the absolute positioning.

Since this button is a control that handles progressive disclosure, it needs an expanded="false" on it.

Also, please use double quotes for HTML attributes.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -149,6 +181,36 @@ export class CbpAppHeader {
...
+                <cbp-icon name='magnifying-glass'></cbp-icon>
+              </cbp-button>
+            }
+
+        {this.globalSearch &&
+          <div>
+            <search hidden>
+              <slot name='cbp-global-search' />
+              
+              <cbp-button
```
dgibson666: I am not sure if this button should be rendered by the component or slotted. But it needs to be type="submit" anyway, in case there is a proper form slotted. I think our story should use a form tag with a submit button. Any ajaxy search is entirely on the app to implement.

dgibson666 (commented):
See code comments. 

Please use double quotes on HTML attributes. 

Also, when you open the search, focus should be placed in the search bar automatically (since the control is covered).

packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
```diff
@@ -6,6 +6,12 @@
 :root {
   --cbp-app-header-color-background: var(--cbp-color-white);
   --cbp-app-header-color-background-dark: var(--cbp-color-base-neutral-dark);
+
+  --cbp-app-header-search-focus-outline: var(--cbp-color-interactive-focus-dark);
```
dgibson666: these are colors and should have color in the name (after search)

packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
bagrub: confirmed this wasn't needed & removed

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: updated to have search tag around this button & using the id global-search-field to replace what was originally the search tag. also update button & associated toggle logic to include/account for the expanded prop on the button & updated HTML attributes to double quotes

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: updated to move the button to a slot cbp-global-search-submit and added the type= "submit" to the slotted content

packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
bagrub: updated to include the color in variable name after search

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: Initially this was added because I was getting an JSX error ( JSX expressions must ahve one parent element. ts(2657)) but this has been updated as part of the below to have the id global-search-field and functionally replace the search tag here

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -182,33 +185,25 @@ export class CbpAppHeader {
...
-                fill='solid'
-                color='primary'
-                variant='square'
-                // onClick={this.toggleSearch}
-              >
-              <cbp-icon name='magnifying-glass'></cbp-icon>
-            </cbp-button>
-            </search>
+          <div id= "global-search-field" hidden>
+              <slot name= "cbp-global-search" />
```
dgibson666: You should only have 1 named slot for search. Right now you don't have a wrapping form tag and no way to include one. Please add a form tag to your story.

bagrub: updated to have the form slotted instead of the input and submit button

packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
```diff
@@ -36,9 +48,62 @@ cbp-app-header{
     width: max-content;
   }
 
+  search{
+    align-self: center;
```
dgibson666: I think both of these can be pushed up to the main flex context using align-items and gap.

packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
```diff
@@ -36,9 +48,62 @@ cbp-app-header{
...
+    flex-grow: 1;
+  }
+
   .cbp-app-header-responsive{
     align-items: center;
     justify-content: space-between;
     width: 100%
   }
+
+  #global-search-field{
```
dgibson666: I would namespace any IDs still. So #cbp-app-header-search (to keep it short-ish)

packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
```diff
@@ -36,9 +48,62 @@ cbp-app-header{
...
+    &:focus-within{
+      outline:  var(--cbp-border-size-md) solid var(--cbp-app-header-search-color-focus-outline);
+      outline-offset: calc( -2 * var(--cbp-border-size-md));
+    }
+
+    form{
+      display: inline-flex;
+      width: 100%;
+
+      input{
```
dgibson666: This nesting is fragile. Now if the developer doesn't specify a form tag, you won't get the border or background colors on the input and all the rest. Form tag is technically optional based on their implementation.

packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
```diff
@@ -36,9 +48,62 @@ cbp-app-header{
...
+          color: var(--cbp-app-header-search-color-placeholder);
+        }
+
+        &:focus-visible{
+          outline: none;
+        }
+      }
+    }
+    
+    form > *:not(input),
```
dgibson666: also fragile and not clear what it's targeting. I would add a comment about it.

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
```diff
@@ -302,4 +302,73 @@ AppHeaderWithSubnav.args = {
...
+      <form
+        slot="cbp-global-search"
+      >
+        <input 
+          type="text" 
+          name="global search" 
+          placeholder="Start Typing - Press ESC to Close" 
+        />
+
+        <cbp-button
```
dgibson666: This button has no accessible label. Add accessibility-text="Search" to the button.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -36,6 +40,33 @@ export class CbpAppHeader {
...
+      this.toggleSearch();
+    }
+  }
+
+  handleTabFocusOut({key, shiftKey}) {
+    if(key == 'Tab' && !shiftKey) this.toggleSearch();
+  }
+
+  handleShiftTabFocusOut({key, shiftKey}) {
+    console.log('handleShiftTabFocusOut triggered', key, shiftKey);
```
dgibson666: remove console.log

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -94,6 +124,22 @@ export class CbpAppHeader {
     this.drawerButton ? this.drawerButton.setAttribute('hidden', ''): '';
   }
 
+
+  toggleSearch() {
+    let search = document.getElementById('global-search-field') as HTMLElement;
```
dgibson666: You can make all of these const since you're just selecting them as references and they aren't changing them.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -149,6 +203,36 @@ export class CbpAppHeader {
           </nav>
         </cbp-resize-observer>
         
+        {this.globalSearch &&
+          <search>
+            <cbp-button
+              id= "global-search-toggle"
```
dgibson666, line 226: This button has no accessible label. Add accessibilityText="Global Search" (needs to be different from the search action button)

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -149,6 +203,36 @@ export class CbpAppHeader {
...
+              type= "button"
+              fill= "outline"
+              color= "secondary"
+              variant= "square"
+              onClick= {this.toggleSearch}
```
dgibson666: it's not currently a factor, but I always use arrow function calls here. If you ever needed to reference a this.scoped variable of the component, it would be a problem.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -149,6 +203,36 @@ export class CbpAppHeader {
...
+              expanded= "false"
+            >
+              <cbp-icon name="magnifying-glass"></cbp-icon>
+            </cbp-button>
+
+            <div id= "global-search-field" hidden>
```
dgibson666: namespace the ID as noted above.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -149,6 +203,36 @@ export class CbpAppHeader {
...
+              <slot name= "cbp-global-search" />
+              <cbp-button
+                type= "button"
+                fill= "ghost"
+                color= "secondary"
+                variant= "square"
+                onClick= {this.toggleSearch}
```
dgibson666: as above (and right below), use arrow function.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -149,6 +203,36 @@ export class CbpAppHeader {
...
+              variant= "square"
+              onClick= {this.toggleSearch}
+              expanded= "false"
+            >
+              <cbp-icon name="magnifying-glass"></cbp-icon>
+            </cbp-button>
+
+            <div id= "global-search-field" hidden>
+              <slot name= "cbp-global-search" />
+              <cbp-button
```
dgibson666: button has no accessible label. Use accessibilityText="Close Search" ? Cancel Search? Maybe as the team. Also, is this supposed to be a ghost button? It looks terrible after a primary solid button. I would think it's an outline button.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -149,6 +203,36 @@ export class CbpAppHeader {
...
+                type= "button"
+                fill= "ghost"
+                color= "secondary"
+                variant= "square"
+                onClick= {this.toggleSearch}
+                onKeyDown={(e) => this.handleTabFocusOut(e)}
+                >
+                <cbp-icon name="times"></cbp-icon>
```
dgibson666: you can minimize/self-close these tags in JSX for shorter code. Just not in the stories or real HTML.

dgibson666 (commented):
See code comments

packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
bagrub: updated to remove & relocate to the main flex content

packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
bagrub: update ID to #cbp-app-header-search

packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
bagrub: updated from form tag to the the slot ([slot="cbp-global-search"]) to make this more robust

packages/web-components/src/components/cbp-app-header/cbp-app-header.scss
bagrub: updated with comment (these are affecting the slotted submit button & the close button) & update selector to not use form but instead the [slot="cbp-global-search"]

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: removed

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: updated from let to const

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: update to use arrow functions as requested

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: update to ID to match above's update

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: updated to use arrow function

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: update to use a self-closing tag

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -149,6 +202,38 @@ export class CbpAppHeader {
...
+          <search>
+            <cbp-button
+              id= "global-search-toggle"
+              type= "button"
+              fill= "outline"
+              color= "secondary"
+              variant= "square"
+              onClick= {() => {this.toggleSearch()}}
+              expanded= "false"
+              accessibility-text="Global Search"
```
dgibson666: use property name, not attribute, per my comment.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -149,6 +202,38 @@ export class CbpAppHeader {
...
+
+            <div id= "cbp-app-header-search" hidden>
+              <slot name= "cbp-global-search" />
+              <cbp-button
+                type= "button"
+                fill= "ghost"
+                color= "secondary"
+                variant= "square"
+                accessibility-text="Close Search"
+                onClick= {() => {this.toggleSearch()}}
```
dgibson666: you don't need the extra {} around the single function call.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
```diff
@@ -149,6 +202,38 @@ export class CbpAppHeader {
...
         
+        {this.globalSearch &&
+          <search>
+            <cbp-button
+              id= "global-search-toggle"
+              type= "button"
+              fill= "outline"
+              color= "secondary"
+              variant= "square"
+              onClick= {() => {this.toggleSearch()}}
```
dgibson666: you don't need the extra {} around the single function call.

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
```diff
@@ -302,4 +302,74 @@ AppHeaderWithSubnav.args = {
...
+  setTimeout(() => {
+    let anchors = document.querySelectorAll('cbp-app-header a');
+    anchors.forEach(anchor => {
+      anchor.addEventListener('click', function(e) { e.preventDefault(); })
+    });
+  }, 500);
+
+  return ` 
+    <cbp-app-header    
+      global-search=true
```
dgibson666: If this is a boolean prop, you don't use =true on it.

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
```diff
@@ -302,4 +302,74 @@ AppHeaderWithSubnav.args = {
...
+      ${drawerid ? `subnav-drawer-id=${drawerid}`: ``}
+      ${sx ? `sx=${JSON.stringify(sx)}` : ''}
+    >
+      ${generateNavItems(items)}
+      <form
+        slot="cbp-global-search"
+      >
+        <input 
+          type="text" 
+          name="global search"
```
dgibson666: oops, I saw this earlier but missed commenting on it - you should not have spaces in the name property. This can cause issues referencing those keys.

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
```diff
@@ -302,4 +302,74 @@ AppHeaderWithSubnav.args = {
...
+  }, 500);
+
+  return ` 
+    <cbp-app-header    
+      global-search=true
+      ${drawerid ? `subnav-drawer-id=${drawerid}`: ``}
+      ${sx ? `sx=${JSON.stringify(sx)}` : ''}
+    >
+      ${generateNavItems(items)}
+      <form
```
dgibson666: you should really have an action attribute on the form tag at a minimum. Probably a method="post" as well since "get" is the default, which puts the keys in the URL, which might not work in an SPA.

dgibson666 (commented):
See code comments. We should confirm the close button's icon and fill as well.

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: updated to property name

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: removed as requested

packages/web-components/src/components/cbp-app-header/cbp-app-header.tsx
bagrub: removed as requested

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
bagrub: removed as requested

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
bagrub: renamed to be globalSearch

packages/web-components/src/components/cbp-app-header/cbp-app-header.stories.tsx
bagrub: added a method="post" as requested

dgibson666 (approved):

PR #336, opened 2025-09-03
Feature/dark mode custom toggle

dgibson666:
* Heavily refactored toggle CSS while adding support for custom icons for dark mode toggle.
* Move dark mode toggle variables to sx, fix initial state value in template/archetype stories.
* Hide the dark mode toggle label in stories.
* Add JSDocs to form field CSS file.


bagrub (approved):
LGTM

PR #338, opened 2025-09-09
Bugfix/template console errors

dgibson666:
* Update app directory link in templates to anchor tag.
* Fix console errors in app header, add search submit event, format code.
* Update app header stories to cancel navigation on search form post
* Add named slot for optional right-aligned button(s).
* Update docs with typescript note.

Tested the optional slot but did not leave it live in the story because it's not a best-practice. Replaced the old commented code with it for easy testing.


bagrub (approved):
LGTM

PR #342, opened 2025-09-16
Feature/update all events

dgibson666:
* App Header accessibility improvements.
* Tweaked code snippet block story.
* Added the media query to cbp-hide custom events when state changes. (should be useful for debugging)
* Add nativeEvent key to most custom event emitters.
* Add nativeEvent key to custom events and prevent bubbling in subnav-items since they can be nested.
* Verified every event emitter changed is firing and made in-code comments and external notes about specific behaviors and/or TechDebt.


bagrub (approved):
LGTM

PR #344, opened 2025-09-17
create Search Results page

bagrub:

### DIFF
--- a/packages/web-components/src/stories/Search_Result.stories.tsx
+++ b/packages/web-components/src/stories/Search_Result.stories.tsx
@@ -1,0 +1,709 @@
+export default {
+  title: 'Archetypes/Search Results',
+  parameters: {
+    layout: 'fullscreen',
+    html: {
+      root: '#storybook-root',
+    },
+  },
+    argTypes: {
+        username: {
+        name: 'User Name',
+        description: 'Name of user to be displayed in the Universal Header',
+        type: { name: 'string', required: true },
+        },
+        isLoggedIn: {
+        name: 'User Log In',
+        description: 'Display Universal Header controls for user log in/out',
+        type: { name: 'boolean' },
+        },
+        search: {
+        description: 'determines if the search field is rendered',
+        control: 'boolean'
+        },
+        searchMethod: {
+        description: 'set the method attribute on the form for search',
+        control: 'text',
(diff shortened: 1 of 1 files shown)


packages/web-components/src/stories/Search_Result.stories.tsx
```diff
@@ -0,0 +1,709 @@
...
+                        <cbp-icon
+                        name="magnifying-glass"
+                        ></cbp-icon>
+                    </cbp-button>
+                    </span>
+                </cbp-form-field-wrapper>
+            </cbp-form-field>
+        </cbp-flex-item>
+        <cbp-flex-item
+            sx='{
```
dgibson666: This is hacky. Rather than pushing them down by .5rem, use the flex properties to align them to the bottom instead. (At the parent level, most likely)

packages/web-components/src/stories/Search_Result.stories.tsx
```diff
@@ -0,0 +1,709 @@
...
+                Filter
+            </cbp-button>
+        </cbp-flex-item>
+        <cbp-flex-item
+            sx='{"margin-top":"0.5rem"}'
+        >
+            <cbp-typography tag="span"> <b><i>Alpha Sort</i></b> </cbp-typography>
+        </cbp-flex-item>
+        <cbp-flex-item
+            sx='{"margin-top":"0.5rem"}'
```
dgibson666: This is hacky. Rather than pushing them down by .5rem, use the flex properties to align them to the bottom instead. (At the parent level, most likely)

dgibson666 (commented):
Just some minor code comments. Looks good overall.

packages/web-components/src/stories/Search_Result.stories.tsx
bagrub: removed the sx on these 3 flex items. Alignment via the parent flex item wasn't bringing these inline so added a sx to the formfield (search bar) to increase the margin bottom(--cbp-form-field-margin-bottom) to bring these visually inline

packages/web-components/src/stories/Search_Result.stories.tsx
bagrub: see above

dgibson666 (approved):
looks good

PR #349, opened 2025-09-24
update search height and accordion a11y errors in sb

bagrub:
Address issues:
*App header search height bug found during Cargo Activity Port Info code review
*Accordion dark Accessibility concerns found in storybook


dgibson666 (approved):
Approved.

However, Accordion Item CSS is pretty messy and I feel like needs a refactor at some point. I don't think it's following our standard patterns for dark mode overrides.

PR #350, opened 2025-09-25
bugfix/dropdown-spacebar-multiselect-bug

dgibson666:
* Fix dropdown bug with spacebar when using combobox + multiselect by tracking a typingMode.
* Don't use typingMode for single key "jump to" behavior (space isn't valid here anyway)

To test, set filter=true and multiselect=true and test typing with space and navigating and selecting with space. It should do the appropriate behavior now.


bagrub (approved):
Changes look good, approved

PR #352, opened 2025-09-25
Bugfix/segmented button group borders

dgibson666:
* Fixed the segmented button group borders when a name is specified, rendering a hidden input (and breaking the :last-child selector)
* Fixed the bug where a segmented button group that allows multiple selections would never update the value to empty string if all buttons were unselected. (both in the event emitter and the hidden input)


bagrub (approved):
lgtm

PR #355, opened 2025-09-30
feature/highlight-token

dgibson666:
* Created cbp-color-highlight token
* Build new dist with Style Dictionary and copy styles to cbp-app's and css-variables.scss
* Add style selector for highlights in core.scss


bagrub (approved):
LGTM, approved

PR #356, opened 2025-10-03
Bugfix/size calculations hidden

dgibson666:
* Create test story for the size calculation issues when hidden initially.
* Remove overflow hidden on accordion items - I think this was meant for animation purposes, which isn't working, but it's showing issues on my test page where a dropdown is truncated.
* Add resize observer to dropdown and form field wrapper to handle case where sizing is wrong when initially hidden.
* Updated docs/code comments.

It turns out that the resize observer automatically handles the hidden case because it reports a size change when revealed. So resize observer, tabs, and breadcrumbs worked properly when hidden. Only dropdown and form field wrapper did not.

### DIFF
--- a/packages/design-tokens/README.md
+++ b/packages/design-tokens/README.md
@@ -1,4 +1,4 @@
-# @cbpds/vdesign-tokens
+# @cbpds/design-tokens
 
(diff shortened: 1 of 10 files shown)


bagrub (approved):
lgtm

PR #357, opened 2025-10-08
Various bug fixes: Accordion, action bar, Search Results

bagrub:
Various bug fixes:
- refactor accordion dark mode
- action bar story accessibility text removal
- update Search results to use mark tag


dgibson666 (commented):
Looks good, but since you are refactoring Accordion, can you update all "fill" variables to "background" to be consistent with naming? I think fill may only be used in button, but that's because there's a fill property.

dgibson666 (approved):
lgtm

PR #358, opened 2025-10-16
update to universal header and brand dhs blue token

bagrub:
Universal banner:
- update logo and seal height
- update height of overall bar

Tokens:
- update the cbp-brand token to #005288


dgibson666 (approved):
looks good

PR #359, opened 2025-10-21
Feature/combobox create item

dgibson666:
* Update dropdowns:
  * Dropdown items are self-slotting (like menu items)
  * Added Create option via prop. Needs lots of testing.
  * Added disabled item styling.
  * Updated "no item found" verbiage contextually:
    * "No items found" for general empty dropdown.
    * "No matches found" when filtering a combobox
    * "Enter X characters to search" for async dropdowns with a minimumInputLength specified. (it now defaults to 1 as well)
* Fixed drawer z-index and make a note in specs about placement issues.
* Updated docs.

Testing dropdowns:
* New messaging should be pretty obvious from above description.
  * "no items" messaging should not be shown when the create option is turned on.
* Create option needs testing:
  * Test every iteration of the dropdown story, as they all have considerations.
  * Turn on "filter" and "create" controls (create only shows when filter = true, which is what enables typing).
  * Test single and multi-select for each story.
  * "Create" option only shows up when search string >= minimumInputLength and no exact match in the dropdown (you don't want to create a duplicate item)


bagrub (approved):
LGTM

bagrub (approved):
lgtm

PR #361, opened 2025-10-24
Bugfix/universal header logo sizing

dgibson666:
* Fix dropdown CSS that broke pagination.
* Update img reset css to not affect image height when specified, fixing universal header image heights.
* Updated some test stories.
* Update dropdown docs.
* Updated universal header requirements doc.
* WIP test case for using resize observer in universal header.

Main things to test:
* Pagination looks right (not broken)
* Universal header images are sized properly at both breakpoints.


bagrub (approved):
lgtm

PR #363, opened 2025-10-27
Initial commit for requirements and documentation

bagrub:
Creation of cbp-carousel, cbp-carousel-item, and cbp-dot-indicator to review requirements and documentation before beginning coding


packages/web-components/src/components/cbp-carousel/cbp-carousel.spec.mdx
```diff
@@ -0,0 +1,49 @@
...
+## Purpose
+
+Carousels are a common UI pattern to show a gallery of associated content to users and iterate through show content in the same area
+
+## Functional Requirements
+
+* A Carousel is composed of a Carousel and Carousel-items in a parent-child relationship
+    * Carousel will manage the buttons and control
+    * Carousel-item will be the individual slides that are being iterated through
+* Only one Carousel-item can be 'active' at a time, this is the carousel-item that is currently visible. Non-'active' Carousel-items will not be visible
```
dgibson666: I think it's overstepping to say non-active items will not be visible. There are cases where they could be visible still, as we discussed related to showing multiple items at once (just not navigating them by "page").

packages/web-components/src/components/cbp-carousel/cbp-carousel.spec.mdx
```diff
@@ -0,0 +1,49 @@
...
+* Carousel-items will have a slot for users to define content
```
dgibson666: this isn't really worth stating. the carousel item really is just a wrapper for the contents.

packages/web-components/src/components/cbp-carousel/cbp-carousel.spec.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+## Technical Specifications
+
+* Carousel-items will have prop to dertermine its place in the order (TODO: reword this and add prop name)
```
dgibson666: unnecessary - they would go by source order, just like tabs. You don't need to detail props anyway. Just what it needs to do.

This is pretty lacking for tech specs. It should cover:
* the parent tag containing the child items
* interfacing with slotted controls
* navigation via arrows, gestures, and keyboard
* mention of animation/scroll snap details

I think we need more discussion around options like how do the items get displayed? Do they stretch to fit the carousel viewport or are they sized at a certain size and fit as many as possible in the viewport.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.spec.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+# cbp-dot-indicator
+
+## Purpose
+
+The dot indicator is intended for use to navigate across small sections of related content, such as in a Carousel.
+
+## Functional Requirements
+
+* Comprised of a series of anchor tags with the associated index of items they represent
```
dgibson666: controls are typically buttons. It's not doing navigation in a way that will change the URL, so it's a button.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.spec.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+* Clicking on an individual dot indicator will set that indicator and its associated indexed item as 'active' 
+
+### Responsiveness
+
+* TODO: no current responsive approach here, intended to keep total size of indicator ~4-8 to avoid issues of wrapping/overflow
+
+### Accessibility
+
+* if no index is explicitly set to 'active', the first index will default to 'active'
```
dgibson666, line 32: This should specify what active means. It's a visible state, but what does active map to in accessibility? It should be aria-selected? Plenty of room to copy more from tabs requirements here.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.spec.mdx
```diff
@@ -0,0 +1,38 @@
...
+* for keyboard navigation, the dot indicator uses the 'roving tabindex' paradigm
+    * Only the active indicator is reachable by tab/shift+tab
+    * Once focus is on the active indicator, using the arrow keys will move the focus to previous/next indicator
+    * Indicators must still be activated by clicking or pressing the 'space' or 'enter' key while focused.
+
+### Additional Notes and Considerations
+
+* Intended to be slotted into the index control slot of a Carousel
+* Advance implementation might include a variant of a 'film strip'
```
dgibson666: Film strip will be a completely different component, so don't mention it.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.spec.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+* Comprised of a series of anchor tags with the associated index of items they represent
+
+## Technical Specifications
+
+* will be set to 'active' when the associated index is active
+
+### User Interactions
+
+* Clicking on an individual dot indicator will set that indicator and its associated indexed item as 'active'
```
dgibson666: This would also emit an event so that the carousel can advance to the selected item. 

Keyboard nav should be detailed here as well.

dgibson666 (commented):
see comments

packages/web-components/src/components/cbp-carousel/cbp-carousel.spec.mdx
bagrub: removed reference to non-active visibility

packages/web-components/src/components/cbp-carousel/cbp-carousel.spec.mdx
bagrub: removed

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.spec.mdx
bagrub: updated to be buttons instead of anchor

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.spec.mdx
bagrub: clarifying this is active in visual spec and clarifying that it will have the aria-selected

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.spec.mdx
bagrub: removed reference to filmstrip

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.spec.mdx
bagrub: added note on event emitter & keyboard nav

packages/web-components/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ This CHANGELOG.md tracks the updates to the web components package of the CBP de
 
 The React components are wrappers generated from this package and will share the same changes. Projects using React 19 may use the native web components without React wrappers.
 
+## [Unpublished] TBD
```
dgibson666, line 7: This will probably give a merge conflict, so you should update from develop and deal with it in your branch (maybe after I merge the grid/flex branch, because I updated it again in there).

packages/web-components/src/components/cbp-carousel/cbp-carousel.scss
```diff
@@ -0,0 +1,36 @@
...
+    width: 100%;
+  
+    & > *{
+      height: inherit;
+      width: inherit;
+    }
+  } 
+
+  
+  .cbp-carousel-viewer[aria-animation="carouselForwards"] cbp-carousel-item > *{
```
dgibson666: Your animations are backward, as noted in my testing comments.

packages/web-components/src/components/cbp-carousel-item/cbp-carousel-item.scss
```diff
@@ -0,0 +1,3 @@
+cbp-carousel-item {
+  display: block;
```
dgibson666: There is no reason these should be in a stacked context. They are supposed to be lined up horizontally, so use inline-block.

packages/web-components/src/components/cbp-carousel/cbp-carousel.scss
```diff
@@ -0,0 +1,36 @@
+cbp-carousel {
+  display: block;
+
+  .cbp-carousel-viewer{
+    height: 100%;
+    width: 100%;
+    overflow: hidden;
+  }
+
+  cbp-carousel-item{
```
dgibson666: most of the carousel-item CSS should live in that file, not here.

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,77 @@
...
+})
+
+export class CbpCarousel {
+
+  private slideIndex; //array of slides
+  
+  @Element() host: HTMLElement; 
+  
+  @Prop({reflect: true}) height
+  @Prop({reflect: true}) width
```
dgibson666: these should not be reflected and confused with deprecated HTML attributes. They should be translated to a CSS variable and assigned via setCSSProps below.

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,77 @@
...
+   
+    for  (let x=0; x < this.slideIndex.length; x++) {
+      const slide = this.slideIndex[x];
+      slide.hidden= true;
+    }
+    this.slideIndex[index].hidden = false;
+  }
+
+  componentDidRender(){
+    this.slideIndex = document.querySelector('.cbp-carousel-viewer').children;
```
dgibson666: * You definitely don't need to be redefining this every render, and it's very odd that it's done after the render.
* This should be defined once in componentWillRender because the slotted child components exist outside of the component lifecycle.
* You should almost never use document.querySelector. Always scope it to this.host.querySelector. All of your code breaks if there are multiple carousels on a page because it lumps all the nodes together.

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,77 @@
...
+    
+    let indicators = document.querySelectorAll('cbp-dot-indicator .dot-indicator');
```
dgibson666: you should probably have a "control" hoisted up at the component level with the items array.

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,77 @@
...
+    if (typeof this.sx == 'string') {
+      this.sx = JSON.parse(this.sx) || {};
+    }
+    setCSSProps(this.host, {
+      ...this.sx,
+    });
+  }
+
+  updateVisible(){
+    let index = document.querySelector('[slot="index-control"] [aria-selected="true"]').getAttribute('index'); //TODO: probably pretty weak selector/logic here
```
dgibson666: you should probably be hoisting the current item up to the component level too to track it. Maybe even as a @State. I feel like this logic overall can be made more modular. I would have a function like navigateCarousel(index) that can be called by the dots and the next/back buttons. The clicked index should be passed by the event (button value), not an invalid attribute on the button.

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,77 @@
...
+      slide.hidden= true;
+    }
+    this.slideIndex[index].hidden = false;
+  }
+
+  componentDidRender(){
+    this.slideIndex = document.querySelector('.cbp-carousel-viewer').children;
+    
+    let indicators = document.querySelectorAll('cbp-dot-indicator .dot-indicator');
+    indicators.forEach((dot) => {
```
dgibson666: I think all of these can be handled by an @Listen rather than setting listeners on each of them manually. And by doing it in componentDidRender, you are setting loads of duplicate listeners and it's going to cause issues, if not memory leaks.

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,77 @@
...
+      dot.addEventListener('click', () => this.updateVisible());
+    })
+    document.querySelector('#carousel-back').addEventListener('click', () => this.updateVisible())
+    document.querySelector('#carousel-forward').addEventListener('click', () => this.updateVisible())
+    
+    this.height != null ? this.host.style.height = this.height : null;
```
dgibson666: this should be handled by setCSSProps above, as noted.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+    return (
+      <Host
+      >
+        
+          <cbp-button
+            type='button'
+            fill='ghost'
+            color='secondary'
+            variant='square'
+            id='carousel-back'
```
dgibson666: You cannot set IDs in a component like this. There shouldn't be a need. Now if you had 2 on a page you have ID conflicts.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+    indicators[0].setAttribute('aria-selected', 'true'); //default set first index to active;
+  }
+
+  render() {
+    return (
+      <Host
+      >
+        
+          <cbp-button
+            type='button'
```
dgibson666: double-quote HTML attributes/props please. type=button is the default, so this can be omitted too.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+            fill='ghost'
+            color='secondary'
+            variant='square'
+            id='carousel-back'
+            onClick={() => {
+                document.querySelector('.cbp-carousel-viewer').setAttribute('aria-animation', 'carouselBackwards');
```
dgibson666: This should be delegated somehow - maybe to the navigation function on the parent. Again, you can't document.querySelector().

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+                this.setIndexActive(this.selectedIndex - 1)
```
dgibson666: give the button a value of the target index you're navigating to. I don't think you need the onclick at all.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+        <div class='indicators'
+          tabindex='0'
+          onKeyDown={({ key }) => {
+            this.keyboardNav(key);
+          }}
+        >
+          {this.generateIndicator()}
+        </div>
+        
+        <cbp-button
```
dgibson666, line 98: same comments as back button

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+
+  
+  @Element() host: HTMLElement; 
+  
+  private selectedIndex: number = 0; // index of the selected indicator
+  private focusIndex: number = 0; // index of the focused indicator, used for keyboard nav
+
+  /** Length of index dot-indicator is tracking */
+  @Prop({ reflect: true }) index: number;
```
dgibson666: Name this items. Index means something else and it's not the number of items or length.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+      item.setAttribute('tabindex', '-1');
+    })
+
+    this.selectedIndex = this.focusIndex = index;
+    let indicator = this.host.querySelector(`.dot-indicator[index="${index}"]`);
+    indicator.setAttribute('aria-selected', 'true');
+    indicator.setAttribute('tabindex', '0');
+  }
+
+  generateIndicator(){
```
dgibson666: This whole thing is way too static and not the web component/JSX way to do things.  I'll look for a better example - I don't have one off the top of my head. but basically you can create an array of DOM nodes and return a spread of them to the render function.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+      Tab: this.focusIndex=this.selectedIndex, // reset the focusIndex when tabbing out of the dot-indicator
+    }[key];
+    if (n !== undefined && key !== 'Tab') {
+      this.focusIndex = n;
+      let focusedIndicator = this.host.querySelectorAll(`.dot-indicator`)[this.focusIndex] as HTMLElement;
+      focusedIndicator.focus();
+    }
+  }
+
+  componentDidRender(){
```
dgibson666: Again, this is the wrong lifecycle hook to be doing this logic and it has all the same issues mentioned in the other components. Most of this should be done as part of the render/generation of dots.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+                document.querySelector('.cbp-carousel-viewer').setAttribute('aria-animation', 'carouselBackwards');
+                this.setIndexActive(this.selectedIndex - 1)
+              }
+            }
+          >
+            <cbp-icon name='angle-down' rotate={90}></cbp-icon>
+          </cbp-button>
+
+        <div class='indicators'
+          tabindex='0'
```
dgibson666: oh, this is the mysterious tab stop I could not figure out. It should not be tabbable.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+            onClick={() => {
+                document.querySelector('.cbp-carousel-viewer').setAttribute('aria-animation', 'carouselBackwards');
+                this.setIndexActive(this.selectedIndex - 1)
+              }
+            }
+          >
+            <cbp-icon name='angle-down' rotate={90}></cbp-icon>
+          </cbp-button>
+
+        <div class='indicators'
```
dgibson666: namespace your class names. Too generic = high chance of collisions with app CSS/code. I'm pretty sure we need some role like group to make this roving tabindex navigation make sense accessibly.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+    let indicator = this.host.querySelector(`.dot-indicator[index="${index}"]`);
+    indicator.setAttribute('aria-selected', 'true');
+    indicator.setAttribute('tabindex', '0');
+  }
+
+  generateIndicator(){
+    let html='';
+    for(let x=0; x < this.index; x++){
+      html+= `<button 
+                class='dot-indicator'
```
dgibson666: namespace your class name. Use double quotes for consistency.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+                index=${x}
+                tabindex="-1"
+              >
+                <div class='dot'>
```
dgibson666: why not make this a span? You shouldn't need a class on it because there is no user content in here - it's all generated, so the structure is fixed and predictable.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.scss
```diff
@@ -0,0 +1,73 @@
...
+  --cbp-dot-indicator-hover-color: var(--cbp-dot-indicator-hover-color-dark);
+  --cbp-dot-indicator-focus-color: var(--cbp-dot-indicator-focus-color-dark);
+  --cbp-dot-indicator-selected-color: var(--cbp-dot-indicator-selected-color-dark);
+}
+
+cbp-dot-indicator {
+  display: flex;
+  gap: var(--cbp-space-4x);
+
+  .indicators{
```
dgibson666: This is all generated content - I would rather see the structure used than a bunch of class names that are not needed. It would make a lot more sense knowing which selector was the button.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.scss
```diff
@@ -0,0 +1,73 @@
...
+      height: var(--cbp-space-9x);
+      width: var(--cbp-space-9x);
+      border: none;
+      background-color: transparent;
+      padding: none;
+      display: inline-flex;
+      align-items: center;
+      justify-content: center;
+
+      .dot{
```
dgibson666: do you even need the nested element? Can't you make the interactive element round and leverage its outline?

dgibson666 (commented):
Quick Testing Results:

When tabbing into the dot indicators (dots) from the back button, the first active item doesn't have any indication of focus. If an item other than the first is active, it takes 2 tabs to give it focus. I don't know where the focus is on the tab immediately after the back button.

"index" is not a valid attribute to place on a button and should be removed.

Animations are backward.
If I'm moving forward, the next item should slide in from the right.
If I'm moving backward, the next slide should slide in from the left.
They should also show the current item being slid out of view. I think this doesn't work because you are setting them as block, which stacks them, when they should be lined up horizontally.

Dot indicator probably needs a prop to set the current item as well in case the parent dictated that or updated it separately. This would override my comments about hoisting it to a state or private variable because it needs to be accessible to other components.

I haven't done accessibility testing yet.

See code comments.

You're also missing the API docs.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.scss
dgibson666: I take this one back - the button needs to be touch-target sized, so making a nested element the dot is fine.

packages/web-components/src/components/cbp-carousel/cbp-carousel.scss
bagrub: refactored animations

packages/web-components/src/components/cbp-carousel-item/cbp-carousel-item.scss
bagrub: update to inline-block

packages/web-components/src/components/cbp-carousel/cbp-carousel.scss
bagrub: moved to carousel-item.scss

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: updated to be set via setCSSProps

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: updated based on discussion to refactor the componentDidRender()

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: updated based on discussion

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: updated based on discussion

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: update to use an @listen for a dotIndicator emitted event

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: being set via setCSSProps

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: refactored to not use ID's

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: updated to use " instead of '

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: refactored animation to not use this

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: updated

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: updated as noted on back button comments

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: updated to be indicatorLength instead of index

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: updated as discussed

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: refactored as disucssed

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: removed

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: updated class names w/ name space

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: refactored to not use classnames for this generated code as noted below

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: removed class names as requested

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.scss
bagrub: updated structure to match the tags instead of using class names

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.scss
```diff
@@ -0,0 +1,72 @@
...
+[data-cbp-theme=dark] cbp-dot-indicator:not([context=dark-inverts]):not([context=light-always]) {
+  --cbp-dot-indicator-hover-color: var(--cbp-dot-indicator-hover-color-dark);
+  --cbp-dot-indicator-focus-color: var(--cbp-dot-indicator-focus-color-dark);
+  --cbp-dot-indicator-selected-color: var(--cbp-dot-indicator-selected-color-dark);
+}
+
+cbp-dot-indicator {
+  display: flex;
+  justify-content: center;
+  gap: var(--cbp-space-4x);
```
dgibson666: I don't think you want this gap. In the mocks, the buttons look uniformly spaced with the dots. Or do you need to add a gap to the dots to separate them and their touch targets?

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+})
+export class CbpDotIndicator {
+
+  @Element() host: HTMLElement; 
+  
+  private selectedIndex: number = 1; // index of the selected indicator
+  private focusIndex: number = 0; // index of the focused indicator, used for keyboard nav
+
+  /** the currently active dot */
+  @Prop ({reflect: true}) activeIndicator: number; //TODO: this may not need reflect: true
```
dgibson666: Both props are named too long. I'd do items: number and current: number. Neither need to be reflected.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+  /** Length of index dot-indicator is tracking */
+  @Prop({ reflect: true }) indicatorLength: number;
+
+  /** Custom event emitted when the Dot-indicator changes active indicator*/
+  @Event() handleIndexChange: EventEmitter;
```
dgibson666: your function name should be named with "handle" not your event. I'm not a fan of "index" in the name though. Your handler should be handleNavigation and your event should be more generic like "navigateCollection" - something we can reuse in a similar replacement like a filmstrip navigation.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+    this.selectedIndex = this.focusIndex = index;
+    
+    let indicator = this.host.querySelector('.dot-indicators-container').childNodes[index] as HTMLButtonElement;
+    indicator.setAttribute('aria-selected', 'true');
+    indicator.setAttribute('tabindex', '0');
+    this.activeIndicator = index;
+
+    this.handleIndexChange.emit({
+      host: this.host,
+      active: this.activeIndicator
```
dgibson666: Instead of active, I'd just use "item" or maybe even "index" here. Index would give more of an indication that it's zero-based. It's a navigation event, so referencing the item/index is pretty self-explanatory.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+    })
+  }
+
+  generateIndicator(){
+    let dotIndicator: HTMLButtonElement[] = [];
```
dgibson666: I'm being a little nitpicky on naming here, but I think the whole thing is the indicator. The individual things are dots, so name it "dots" and "newDot" etc. Shorter and clearer.

packages/web-components/src/components/cbp-carousel/cbp-carousel.stories.tsx
```diff
@@ -0,0 +1,70 @@
+export default {
+  title: 'Components/Carousel',
+  tags: ['beta'],
```
dgibson666: Please mark this as "new"

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,88 @@
...
+  @Listen('handleIndexChange')
+  handleIndexChange() {
+    this.updateVisible();
+  }
+
+  updateVisible(){
+      // let carouselContainer = this.host.querySelector('.cbp-carousel-container');
+      // carouselContainer.classList.contains('carousel-animation')?carouselContainer.classList.remove('carousel-animation'): null;
+
+      let activeCarouselItem = this.control.activeIndicator;
```
dgibson666: I thought we talked about the active item being moved up to a prop. I think makes sense and would allow an outside script to control it easier. I realize this may cause a re-render the way you have it implemented, but I think it makes more sense.

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,88 @@
...
+      });
+      // carouselContainer.classList.add('carousel-animation');
+  }
+
+  componentDidRender(){
+    this.updateVisible();
+  }
+
+  componentWillRender(){
+    this.control = this.host.querySelector('[slot="index-control"]');
```
dgibson666: this should be done once on component will load. There's no need to keep updating it every render as it should never change.

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,88 @@
...
+        <div
+          class="cbp-carousel-viewer"
+          role="group"
+          aria-description="slides"
+        >
+          <div class="cbp-carousel-container">
+            <slot></slot>
+          </div>
+        </div>
+        <slot name="index-control"></slot>
```
dgibson666: slot names need to be namespaced. cbp-carousel-controls.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.scss
```diff
@@ -0,0 +1,72 @@
+:root {
+  --cbp-dot-indicator-hover-color: var(--cbp-color-interactive-secondary-darker);
```
dgibson666: Your naming isn't following conventions. "color" comes before interactive states, etc.

dgibson666 (commented):
Functionally, it's looking good. See code comments. Also, please add API docs files and add JSDocs to document props, events, slots, and CSS vars.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+  componentDidRender(){
+  this.setIndexActive(this.activeIndicator);
+  }
+
+  render() {
+    return (
+      <Host
+      >
+        <cbp-button
```
dgibson666, line 81: this button has no accessible label.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+        <div class="dot-indicators-container"
+          onKeyDown={({ key }) => {
+            this.keyboardNav(key);
+          }}
+        >
+          {this.generateIndicator()}
+        </div>
+        
+        <cbp-button
```
dgibson666, line 98: this button has no accessible label.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+      host: this.host,
+      active: this.activeIndicator
+    })
+  }
+
+  generateIndicator(){
+    let dotIndicator: HTMLButtonElement[] = [];
+    for (let x = 0; x < this.indicatorLength; x++){
+      let newIndicator: HTMLButtonElement = 
+        <button
```
dgibson666: these buttons have no accessible label.

dgibson666 (commented):
added a few more notes - none of the buttons are accessible. Need accessibilityText for the components and aria-label for the standard buttons.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.scss
bagrub: removed

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: updates as requested

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: renamed as requested

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: renamed as requested

packages/web-components/src/components/cbp-carousel/cbp-carousel.stories.tsx
bagrub: updated

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: updated this to be prop

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: moved to componentWillLoad

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: updated name

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.scss
bagrub: updated variable names to follow convention

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: added accessible label

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: added accessible label

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: added accessible label

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+  tags: ['new'],
+  argTypes: {
+  },
+};
+
+
+const Template = ({slides}) => {
+  return `
+    <cbp-dot-indicator 
+        current=0
```
dgibson666: if this needs to be default in the story, it should be removed and defaulted in the component.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+        items=${slides.length}
```
dgibson666: this should just be a number control to sync with the prop. You don't need to create a fake array for it. It serves no purpose.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -36,34 +36,34 @@ export class CbpDotIndicator {
...
   generateIndicator(){
-    let dotIndicator: HTMLButtonElement[] = [];
-    for (let x = 0; x < this.indicatorLength; x++){
+    let dots: HTMLButtonElement[] = [];
+    for (let x = 0; x < this.items; x++){
       let newIndicator: HTMLButtonElement = 
         <button 
-          // aria-index={`${x}`} //TODO: aria-index not right, just here for testing
-          aria-selected = {x == this.activeIndicator ? "true" : "false"}
+          aria-label={(x+1) + " of " + this.items + " Dots"}
```
dgibson666: Use a template literal rather than string concatenation. And "Dots" isn't really helpful. I was thinking about if we should have a prop like "itemName" that gives a name to what this represents. Can default to "item" or "slide" but being customizable makes this more generic and reusable in other contexts.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -91,6 +91,7 @@ export class CbpDotIndicator {
           color="secondary"
           variant="square"
           value="carousel-back"
+          accessibilityText='carousel back'
```
dgibson666: definitely shouldn't say "carousel" in it since this is a generic component and decoupled from carousel.

If you are not doing anything with the value attribute, delete it.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -113,6 +114,7 @@ export class CbpDotIndicator {
           color="secondary"
           variant="square"
           value="carousel-forward"
+          accessibilityText='carousel forward'
```
dgibson666: definitely shouldn't say "carousel" in it since this is a generic component and decoupled from carousel.

If you are not doing anything with the value attribute, delete it.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -36,34 +36,34 @@ export class CbpDotIndicator {
     let indicator = this.host.querySelector('.dot-indicators-container').childNodes[index] as HTMLButtonElement;
```
dgibson666: This is the same querySelector as on line 29, so use the same selector.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -79,7 +79,7 @@ export class CbpDotIndicator {
   }
 
   componentDidRender(){
-  this.setIndexActive(this.activeIndicator);
+  this.setIndexActive(this.current);
```
dgibson666: I don't think this should be in componentDidRender. If you console.log your render function, you are likely causing a double render. This should happen before or during render.

dgibson666 (commented):
see code comments

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.stories.tsx
bagrub: removed & updated

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.stories.tsx
bagrub: updated and removed

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: updated to include prop as discussed

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: updated to use prop mentioned above for language consistency & removed value

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: same as above

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: removed as part of refactor on setIndexActive discussed

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: updated and removed and refactored generatetIndicator be based on prop as discussed

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,95 @@
...
+  styleUrl: 'cbp-carousel.scss'
+})
+
+export class CbpCarousel {
+
+  private control: HTMLCbpDotIndicatorElement;
+    
+  @Element() host: HTMLElement; 
+  
+  @Prop() height;
```
dgibson666: missing docs and typings. Didn't we also say they would default to 100%? Or do they do that in CSS already? Hard to test since you're hardcoding the values in the story with no controls.

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,95 @@
...
+  @Prop() width;
+
+  /** used to set the activeItem for the carousel*/
+  @Prop() activeItem
```
dgibson666: missing typings. With no type and vague description I have no idea what this prop accepts or represents. Without a watch on this, it doesn't appear to do anything. Also need control for it in the story to test.

packages/web-components/src/components/cbp-carousel/cbp-carousel.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const html = slides.map(({ content }) => {
+    return `<cbp-carousel-item>${content}</cbp-carousel-item>`;
+  });
+  return html.join('');
+}
+
+const Template = ({slides, sx}) => {
+  return `
+   <cbp-carousel
+    height="300px"
```
dgibson666: remove these and add controls. they should not be defaulted to a fixed size.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.scss
```diff
@@ -0,0 +1,71 @@
...
+  display: flex;
+  justify-content: center;
+
+  .dot-indicators-container{
+    display: inline-flex;
+
+    button{
+      height: var(--cbp-space-9x);
+      width: var(--cbp-space-9x);
+      border: none;
```
dgibson666: use 0 instead of none. it's shorter.

packages/web-components/src/components/cbp-carousel-item/cbp-carousel-item.tsx
```diff
@@ -0,0 +1,17 @@
+import { Component, Host, h } from '@stencil/core';
+
```
dgibson666, line 10: missing slot docs

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,95 @@
+import { Component, Element, Host, h, Listen, Prop} from '@stencil/core';
+import { setCSSProps } from '../../utils/utils';
+
+/**
+ * The Carousel is a control meant to take slotted Carousel-items to visually
+ * iterate though to display with a slotted control of a dot-indicator or other control
+ * 
+ * @slot - Anything placed into the default slot will be rendered in the cbp-carousel-container
```
dgibson666: This should really recommend to only slot `cbp-carousel-item`s in the default slot.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,111 @@
...
+})
+export class CbpDotIndicator {
+
+  @Element() host: HTMLElement; 
+  
+  private selectedIndex: number = 0; // index of the selected indicator
+  private focusIndex: number = 0; // index of the focused indicator, used for keyboard nav
+
+  /** the currently active dot */
+  @Prop () current= 0;
```
dgibson666: missing typing

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  /** unit of measure for what the dot indicator is measuring */
+  @Prop() unit = 'Item';
```
dgibson666: Call this "itemName", also missing typing.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  /** Length of index dot-indicator is tracking */
+  @Prop() items: number;
```
dgibson666, line 21: missing typing

dgibson666 (commented):
Still some issues. I think some are being glossed over based on your stories lack of controls. 

Also, please set your tab width to 2 spaces and reformat. Your CSS is correct but some docs and components are not consistent.

packages/web-components/src/components/cbp-carousel/cbp-carousel.scss
```diff
@@ -0,0 +1,32 @@
...
+  width: var(--cbp-carousel-width);
+
+  .cbp-carousel-viewer{
+    height: 100%;
+    width: 100%;
+    overflow: hidden;
+
+      .cbp-carousel-container{
+        display: flex;
+        animation: slide 2s forwards;
```
dgibson666: drop the animation/keyframes. Its this simple:

```CSS
        transform: translateX(var(--cbp-carousel-offset));
        transition: all 0.5s ease-out;
```

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: updated docs and typings. refactored as discussed

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: updated typings, story controls, & description

packages/web-components/src/components/cbp-carousel/cbp-carousel.stories.tsx
bagrub: removed and updated to hook into a control

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.scss
bagrub: update as requested

packages/web-components/src/components/cbp-carousel-item/cbp-carousel-item.tsx
bagrub: updated with slot documentation

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: update language as requested in documentation

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: update typings

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: update prop name & added typing

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: updated typing

packages/web-components/src/components/cbp-carousel/cbp-carousel.scss
bagrub: update to remove animation in favor of the above

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.stories.tsx
```diff
@@ -0,0 +1,17 @@
+export default {
+  title: 'Components/Dot Indicator',
+  tags: ['new'],
+  argTypes: {
```
dgibson666, line 4: Add controls for all props: current, items, itemName.

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,112 @@
...
+})
+
+export class CbpCarousel {
+
+  private control: HTMLCbpDotIndicatorElement;
+
+  @Element() host: HTMLElement;
+
+  //TODO: seperate these from the cbp-carousel-item instance
+  @Prop() height: string = '100%';
```
dgibson666, line 25: Still need docs with CSS units verbiage (reference flex, which uses it a lot)

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,112 @@
...
+  @Prop() width: string = '100%';
+
+  /** used to set the activeItem for the carousel*/
+  @Prop() activeItem
```
dgibson666: Change to activeIndex since it's zero-based and update story control. This also needs {mutable: true} if we are updating it internally within the component logic. Also default it to 0.

You still need a watch on this for reactivity - Storybook does a full reload when you change the control value, so it's tricking you into thinking it's working.

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,112 @@
...
+    this.updateVisible();
+  }
+
+  @Listen('navigateCollection')
+  navigateCollection() {
+    this.updateVisible();
+  }
+
+  handleResize(width) {
+    this.width = width;
```
dgibson666: This prop is not mutable and shouldn't be updated (per discussion). This should re-set the CSS offset var with the recalculated pixel value so it's positioned correctly after re-size.

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+    //TODO: need to find behavior for fullscreen and responsive to be triggered here
+  }
+
+  updateVisible() {
```
dgibson666: You can break this into 2 function:

* updateCurrent - handles wrap-around navigation before updating the prop
* scrollToItem - handles offset calculation based on calculated width and current item/index.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  render() {
+    return (
+      <Host
+      >
+        <cbp-button
+          fill="ghost"
+          color="secondary"
+          variant="square"
+          accessibilityText={this.itemName + " back"}
```
dgibson666: Do "Previous Item" (dynamic) - remember to use a template literal rather than string concatenation.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,111 @@
...
+          }}
+        >
+          {this.generateIndicator()}
+        </div>
+
+        <cbp-button
+          fill="ghost"
+          color="secondary"
+          variant="square"
+          accessibilityText={this.itemName + " forward"}
```
dgibson666: Do "Next Item" (dynamic) - remember to use a template literal rather than string concatenation.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
```diff
@@ -0,0 +1,111 @@
...
+      item: this.current
+    })
+  }
+
+  generateIndicator() {
+    let dots: HTMLButtonElement[] = [];
+    for (let x = 0; x < this.items; x++) {
+      let newIndicator: HTMLButtonElement =
+        <button
+          aria-label={(x + 1) + " of " + this.items + " " + this.itemName}
```
dgibson666: Remember to use a template literal rather than string concatenation.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+  },
+};
+
+
+const Template = () => {
+  return `
+    <cbp-dot-indicator
+        items=9
```
dgibson666: this should be a control, along with the others.

dgibson666 (commented):
See code comments, per discussion (I might have caught a couple more small things)

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -0,0 +1,112 @@
...
+      this.sx = JSON.parse(this.sx) || {};
+    }
+    setCSSProps(this.host, {
+      "--cbp-carousel-height": `${this.height}`,
+      "--cbp-carousel-width": `${this.width}`,
+      ...this.sx,
+    });
+
+    this.control = this.host.querySelector('[slot="cbp-carousel-controls"]');
+    this.updateVisible();
```
dgibson666: The resize observer should fire when it becomes visible, so this is probably not needed anymore.

dgibson666 (commented):
Added a couple more code comments I forgot.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.stories.tsx
bagrub: Added controls for current, items, & itemName

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: update doc using language found in the flex-item to match CSS units description as discussed

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: Update to ActiveIndex, set to mutable:true, defaulting to 0. Added a @watch on this as well and validated the activeIndex is within bounds of control

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: removed as part of updateVisibility() refactor below. now this is just calling the scrollToItem

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: refactored. updateCurrent is updating the props & scrollToItem is handling the calculcations for the css var offsets as discussed

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: Update language to match above

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: Update language to match above

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.tsx
bagrub: updated to use template literal

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.stories.tsx
bagrub: updated to be a control

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: removed

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+      items=${items}
+      item-name=${itemName}
+    ></cbp-dot-indicator>
+`;
+};
+
+export const dotIndicator = Template.bind({});
+dotIndicator.args = {
+  items: 9,
+  itemName: 'Slide'
```
dgibson666: remove itemName arg - it should be the component default with no arg passed.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.stories.tsx
```diff
@@ -18,14 +18,18 @@ export default {
...
-const Template = (current, items, itemName) => {
+const Template = ({ current, items, itemName }) => {
   return `
-    <cbp-dot-indicator>
-      ${current ? `current=${current}` : ''}
-      ${items ? `items=${items}` : ''}
-      ${itemName ? `itemName=${itemName}` : ''}
+    <cbp-dot-indicator
+      ${current ? `current=${current}` : ""}
+      items=${items}
```
dgibson666: both of these props should still have the ternary logic to not render if undefined.

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -43,55 +46,56 @@ export class CbpCarousel {
     });
 
     this.control = this.host.querySelector('[slot="cbp-carousel-controls"]');
-    this.updateVisible();
+  }
+
+  @Watch('activeIndex')
```
dgibson666: This watch isn't firing (or doing anything) and you shouldn't be looking at the controls anyway. You should have queried a collection of the slides as "private items: HTMLCbpCarouselItemElement = []"

This should be calling scrollToItem() because the current item has changed.

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -43,55 +46,56 @@ export class CbpCarousel {
...
-    let widthValue = 0;
-    if (this.width.includes('%')) {
-      widthValue = (this.host.offsetWidth * parseInt(this.width)) / 100;
-
-      console.log('sanity check for %: ', widthValue)
-    } else {
-      widthValue = parseInt(this.width)
-    }
+  scrollToItem() {
+    let carouselItems = this.host.querySelectorAll('cbp-carousel-item') as unknown as HTMLCbpCarouselItemElement[];
```
dgibson666: This should not be done every time navigation happens. It should be queried once on componentWillLoad and stored in a private variable as mentioned above.

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
```diff
@@ -43,55 +46,56 @@ export class CbpCarousel {
...
     this.control = this.host.querySelector('[slot="cbp-carousel-controls"]');
-    this.updateVisible();
+  }
+
+  @Watch('activeIndex')
+  watchActiveIndex(e) {
+    if (e < 0) {
+      this.activeIndex = 0;
+    } else if (e > this.control.items) {
+      this.activeIndex = this.control.items
```
dgibson666: I'm confused about this line. The number of items will always be higher than the index, so won't this break things? I think this watch also needs to sync the this.control?.current value

dgibson666 (commented):
see code comments

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.stories.tsx
bagrub: removed

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.stories.tsx
bagrub: reverted back to have ternary logic

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: updated based on paired programming

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: update based on paired programming

packages/web-components/src/components/cbp-carousel/cbp-carousel.tsx
bagrub: updated based on paired programming

dgibson666 (approved):
Looks good. Expect future iterations based on other variants and a11y updates.

PR #364, opened 2025-11-04
Feature/range slider

dgibson666:
* Heavily refactored `cbp-slider` to handle a range variant accepting 2 values.
* Added custom event emitter that was missing.
* Removed keypress listener from numeric inputs, as they were firing extra events.

I left some code comments in `cbp-form-field` that was testing, for future consideration. I will be revisiting forms and fields next sprint.


bagrub (approved):
looking good, approved

PR #366, opened 2025-11-05
bugfix/flex-grid-linear-spacing

dgibson666:
* Updated grid and flex components to apply a margin-bottom when linearized (based on vertical gap value by default).
* Fixed stories to allow a gap with 2 values.


bagrub (approved):
Approved

PR #368, opened 2025-11-14
Add disabled state to the chip component

bagrub:
Update chip component to have a disabled state

### DIFF
--- a/packages/web-components/src/components/cbp-chip/cbp-chip.scss
+++ b/packages/web-components/src/components/cbp-chip/cbp-chip.scss
@@ -20,7 +20,7 @@
  * @prop --cbp-chip-icon-color-outline-focus: var(--cbp-color-white);
  * @prop --cbp-chip-icon-color-outline-focus-dark: transparent;
  */
-:root { 
+:root {
   --cbp-chip-color-text: var(--cbp-color-text-lightest);
   --cbp-chip-color-text-dark: var(--cbp-color-text-darkest);
   --cbp-chip-color-text-hover: var(--cbp-color-text-lightest);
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/cbp-chip/cbp-chip.tsx
```diff
@@ -74,6 +77,7 @@ export class CbpChip {
   componentDidLoad() {
     if (!this.icon) this.icon = this.host.querySelector('cbp-icon');
     this.iconName = this.icon.name;
+    if (this.disabled) this.button.setAttribute('disabled', '');
```
dgibson666: we only do this for slotted controls. disabled should be in the render since the button is rendered.

packages/web-components/src/components/cbp-chip/cbp-chip.scss
```diff
@@ -121,7 +127,7 @@ cbp-chip {
     &:hover {
```
dgibson666: there's still a hover state on disabled buttons (there should not be any interactivity), so add :not([disabled]) to this selector and check that it doesn't affect anything else. Disabled buttons aren't focusable, so no need to add it to the focus states.

dgibson666 (commented):
couple code comments.

packages/web-components/src/components/cbp-chip/cbp-chip.tsx
bagrub: updated to have the disabled control in the render, removed from componentDidLoad

packages/web-components/src/components/cbp-chip/cbp-chip.scss
bagrub: Update to have :not(:disabled) to the hover as requested

dgibson666 (approved):

PR #372, opened 2025-11-20
Updates to archetype and user preference drawer

bagrub:
User Preference Drawer: 
*update drawer panel header to have brand color background and white text in both dark and light mode
*update drawer Panel header to have border-color to gray-cool
*update some text to match desgin spec
*update font size of headers inside panel content
*add top and bottom borders to theme toggle
*update label text for theme toggle

Archetype Page:
* update filter panel age slider to an range slider component
* update spacing for gap to account for linearized mode and tweak spacing


dgibson666 (commented):
Looks good so far. As discussed, just need:
* Updated User Prefs Drawer close button color for default state via `sx`.
* Fix broken responsive nav drawer in Passenger Archetype.

dgibson666 (approved):
looks good

PR #375, opened 2025-11-28
update typography story

bagrub:
update typography storybook to add story with all variants visible at once to allow for better testing of the component


dgibson666 (approved):
I pushed a small story update to shorten the variant text. Otherwise looks good and approved.

PR #376, opened 2025-12-03
Bugfix/hide story

dgibson666:
* Fix the story for the cbp-hide component, which breaks with spaces in control values
* Update docs


bagrub (approved):
Looks good to me

PR #379, opened 2025-12-09
Bugfix/form investigations

dgibson666:
* Fixed file input setting 'undefined' as name
* Some cleanup on form component, uncommented main console.log for testing
* Cleaned up stories, made form names different (reduces duplicate logs/events)


bagrub (approved):
Looks good to me

PR #380, opened 2025-12-09
Upgrade to Storybook 10

dgibson666:
Upgraded to Storybook 10.

It did some automigrations and warns about the HTML and Tag-Badges addons being incompatible, but it runs and they still work. I verified that they do not have new versions for Storybook 10 yet, likely because they still function correctly.

Run `npm install` after pulling down this branch.


bagrub (approved):
Looks good to me

PR #381, opened 2025-12-10
updates to address accessibility flags in storybook

bagrub:
*Action Bar => serious issues(2)  about color contrast (4.32) on the button color in dark mode (color-blue-30 over gray-cool-70)
*Card => flag card story: Image alt text required
*Footer => color contrast of 4.32 (foreground: (-{}cbp-color-interactive-primary-light background: ({-}-cbp-color-gray-cool-70))   
*Multicol => 
	*List structure(1): list element has direct children that are not allowed: cbp-multicol 
 	*List Item(10): list item does not have a <ul>, <ol> parent element.
*Segmented Button Group => Button Name (3): every button needs a label or accessible name [Icon example]
*Tag => cbp-tag(default/info) has contrast of 4.24 (color: ({}cbp-color-info-darker) bg: ({-}-cbp-color-info-light))
*Tooltip -> ARIA command name : needs a label or accessible name (has role='button' but no aria-label or etc)
*structured list selectable story: form label (6) => Every form field needs an associated label

### DIFF
--- a/packages/web-components/src/components/cbp-action-bar/cbp-action-bar.stories.tsx
+++ b/packages/web-components/src/components/cbp-action-bar/cbp-action-bar.stories.tsx
@@ -35,15 +35,17 @@ export default {
             <cbp-button 
             ${context && context != 'light-inverts' ? `context=${context}` : ''}   
             fill="ghost"
-            >
+            sx='{"--cbp-button-color-dark":"var(--cbp-color-blue-20"}'
+          >
               Action 1
-            </cbp-button>
-            <cbp-button 
+          </cbp-button>
+          <cbp-button 
               ${context && context != 'light-inverts' ? `context=${context}` : ''}     
               fill="ghost"
-            >
+              sx='{"--cbp-button-color-dark":"var(--cbp-color-blue-20"}'
+          >
               Action 2
-            </cbp-button>
+          </cbp-button>
         </cbp-action-bar>
       `;
   };
--- a/packages/web-components/src/components/cbp-action-bar/cbp-action-bar.stories.tsx
+++ b/packages/web-components/src/components/cbp-action-bar/cbp-action-bar.stories.tsx
@@ -35,14 +35,13 @@ export default {
             <cbp-button 
             ${context && context != 'light-inverts' ? `context=${context}` : ''}   
             fill="ghost"
-            sx='{"--cbp-button-color-dark":"var(--cbp-color-blue-20"}'
           >
               Action 1
           </cbp-button>
           <cbp-button 
               ${context && context != 'light-inverts' ? `context=${context}` : ''}     
               fill="ghost"
-              sx='{"--cbp-button-color-dark":"var(--cbp-color-blue-20"}'
+              color="secondary"
           >
               Action 2
           </cbp-button>
--- a/packages/web-components/src/components/cbp-footer/cbp-footer.stories.tsx
+++ b/packages/web-components/src/components/cbp-footer/cbp-footer.stories.tsx
@@ -43,8 +43,8 @@ const InternalTemplate = ({ footerNav }) => {
         
         <cbp-flex gap="var(--cbp-space-4x)" wrap="wrap">
           <span><i>Having an issue?</i></span>
-          <span><b>Email: </b><cbp-link href="mailto:somebody@example.com" context="dark-always">this-application-support@abc.def.gov</cbp-link></span>
-          <span><b>CBP Helpdesk: </b><cbp-link href="tel:555-555-5555" context="dark-always">(555) 555-5555</cbp-link></span>
+          <span><b>Email: </b><cbp-link href="mailto:somebody@example.com" context="dark-always" sx='{"--cbp-link-color-dark":"var(--cbp-color-blue-20)"}'>this-application-support@abc.def.gov</cbp-link></span>
+          <span><b>CBP Helpdesk: </b><cbp-link href="tel:555-555-5555" context="dark-always" sx='{"--cbp-link-color-dark":"var(--cbp-color-blue-20)"}'>(555) 555-5555</cbp-link></span>
         </cbp-flex>
       </section>
     </cbp-footer>
--- a/packages/web-components/src/components/cbp-footer/cbp-footer.stories.tsx
+++ b/packages/web-components/src/components/cbp-footer/cbp-footer.stories.tsx
@@ -34,7 +34,7 @@ const InternalTemplate = ({ footerNav }) => {
       </nav>
 
       <section>
-        <cbp-typography tag="h6" variant="heading-md" context="dark-always" sx='{"margin-bottom":"var(--cbp-space-2x)"}'>
+        <cbp-typography tag="span" variant="heading-md" context="dark-always" sx='{"margin-bottom":"var(--cbp-space-2x)"}'>
           <cbp-icon name='headset' size='1.25rem'> </cbp-icon>
           Application Support
         </cbp-typography>
--- a/packages/web-components/src/components/cbp-footer/cbp-footer.stories.tsx
+++ b/packages/web-components/src/components/cbp-footer/cbp-footer.stories.tsx
@@ -43,8 +43,8 @@ const InternalTemplate = ({ footerNav }) => {
         
         <cbp-flex gap="var(--cbp-space-4x)" wrap="wrap">
           <span><i>Having an issue?</i></span>
-          <span><b>Email: </b><cbp-link href="mailto:somebody@example.com" context="dark-always" sx='{"--cbp-link-color-dark":"var(--cbp-color-blue-20)"}'>this-application-support@abc.def.gov</cbp-link></span>
-          <span><b>CBP Helpdesk: </b><cbp-link href="tel:555-555-5555" context="dark-always" sx='{"--cbp-link-color-dark":"var(--cbp-color-blue-20)"}'>(555) 555-5555</cbp-link></span>
+          <span><b>Email: </b><cbp-link href="mailto:somebody@example.com" context="dark-always" >this-application-support@abc.def.gov</cbp-link></span>
+          <span><b>CBP Helpdesk: </b><cbp-link href="tel:555-555-5555" context="dark-always" >(555) 555-5555</cbp-link></span>
         </cbp-flex>
       </section>
     </cbp-footer>
--- a/packages/web-components/src/components/cbp-multicol/cbp-multicol.stories.tsx
+++ b/packages/web-components/src/components/cbp-multicol/cbp-multicol.stories.tsx
@@ -86,5 +86,5 @@ Multicol.args = {
   columns: 2,
   width: '8rem',
   gap: 'var(--cbp-space-4x)',
-  sx: {"padding":"var(--cbp-space-4x)"}
+  sx: {"padding-inline-start":"var(--cbp-space-4x)"}
 };
--- a/packages/web-components/src/components/cbp-multicol/cbp-multicol.stories.tsx
+++ b/packages/web-components/src/components/cbp-multicol/cbp-multicol.stories.tsx
@@ -66,8 +66,8 @@ function createChildren(children) {
 }
 
 const Template = ({ content, columns, width, gap, rule, nobreak, sx }) => {
+  //TODO: removed UL to deal with accessibility flag
   return `
-    <ul>
       <cbp-multicol
         ${columns ? `columns="${columns}"` : ''}
         ${width ? `width="${width}"` : ''}
@@ -78,7 +78,6 @@ const Template = ({ content, columns, width, gap, rule, nobreak, sx }) => {
       >
         ${createChildren(content)}
       </cbp-multicol>
-    </ul>
   `;
 };
 
--- a/packages/web-components/src/components/cbp-multicol/cbp-multicol.stories.tsx
+++ b/packages/web-components/src/components/cbp-multicol/cbp-multicol.stories.tsx
@@ -66,7 +66,6 @@ function createChildren(children) {
 }
 
 const Template = ({ content, columns, width, gap, rule, nobreak, sx }) => {
-  //TODO: removed UL to deal with accessibility flag
   return `
       <cbp-multicol
         ${columns ? `columns="${columns}"` : ''}
@@ -75,6 +74,7 @@ const Template = ({ content, columns, width, gap, rule, nobreak, sx }) => {
         ${rule ? `rule="${rule}"` : ''}
         ${nobreak ? `nobreak` : ''}
         ${sx ? `sx=${JSON.stringify(sx)}` : ''}
+        role="list"
       >
         ${createChildren(content)}
       </cbp-multicol>
@@ -85,5 +85,6 @@ export const Multicol = Template.bind({});
 Multicol.args = {
   columns: 2,
   width: '8rem',
-  gap: 'var(--cbp-space-4x)'
+  gap: 'var(--cbp-space-4x)',
+  sx: {"padding":"var(--cbp-space-4x)"}
 };
--- a/packages/web-components/src/components/cbp-multicol/cbp-multicol.tsx
+++ b/packages/web-components/src/components/cbp-multicol/cbp-multicol.tsx
@@ -51,7 +51,7 @@ export class CbpMulticol {
 
   render() {
     return (
-      <Host>
+      <Host role="list">
         <slot />
       </Host>
     );
--- a/packages/web-components/src/components/cbp-multicol/cbp-multicol.tsx
+++ b/packages/web-components/src/components/cbp-multicol/cbp-multicol.tsx
@@ -51,7 +51,7 @@ export class CbpMulticol {
 
   render() {
     return (
-      <Host role="list">
+      <Host>
         <slot />
       </Host>
     );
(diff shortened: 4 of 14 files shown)


packages/web-components/src/components/cbp-action-bar/cbp-action-bar.stories.tsx
```diff
@@ -35,15 +35,17 @@ export default {
             <cbp-button 
             ${context && context != 'light-inverts' ? `context=${context}` : ''}   
             fill="ghost"
-            >
+            sx='{"--cbp-button-color-dark":"var(--cbp-color-blue-20"}'
```
dgibson666: I'm not sure about using sx for an accessibility issue. I'm also wondering why we have 2 primary buttons here when that is usually not the guidance or case.

packages/web-components/src/components/cbp-multicol/cbp-multicol.tsx
```diff
@@ -51,7 +51,7 @@ export class CbpMulticol {
 
   render() {
     return (
-      <Host>
+      <Host role="list">
```
dgibson666: No, you can't assume all of these are lists. role is value on the custom element, so put it in the story instead.

packages/web-components/src/components/cbp-segmented-button-group/cbp-segmented-button-group.stories.tsx
```diff
@@ -36,6 +36,7 @@ function generateButtons(buttons) {
         ${pressed == true ? `pressed="${pressed}"` : ''}
         ${disabled == true ? 'disabled' : ''}
         ${variant ? `variant="${variant}"` : ''}
+        accessibility-text="segementedButtonExample"
```
dgibson666: this does not label the buttons and does not fix anything. In fact, it overrides the default labels of the ones with text with something inaccessible. Add real text labels to the buttons object in the second story to populate this prop.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -60,6 +60,7 @@ function generateSelectableItems(items, context){
         type="checkbox"
         name="checkbox"
         value=${index}
+        aria-label= ${content}
```
dgibson666: First, this isn't quoted, so if you look at the HTML, it's broken on spaces. Second, the content can be long, which is not ideal for an aria-label - verbosity is an issue. Let's try updating the story to use `<cbp-hide visually-hidden>` instead of a span with display: none. I think that should work.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -631,7 +632,7 @@ const StructuredListSelectableTemplate = ({ listItems, striped, showHeader, head
           ${context && context != 'light-inverts' ? `context=${context}` : ''}   
           ${sx ? `sx=${JSON.stringify(sx)}` : ''}
         >
-        ${showHeader ? `<div slot="cbp-structured-list-header" id="${headerId}"><cbp-checkbox><input type="checkbox" /><span style='display: none'>check all</span></cbp-checkbox> <span>5 Search Results</span></div>` : ''}
+        ${showHeader ? `<div slot="cbp-structured-list-header" id="${headerId}"><cbp-checkbox><input type="checkbox" aria-label="Select All"/><span style='display: none'>check all</span></cbp-checkbox> <span>5 Search Results</span></div>` : ''}
```
dgibson666: The aria-label can be removed if the cbp-hide update works.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.stories.tsx
```diff
@@ -35,7 +35,7 @@ const Template = ({ open, uid, alignment, title, content, tooltipControl, contex
             alignment=${alignment}
             ${context && context != 'light-inverts' ? `context=${context}` : ''}
             ${sx ? `sx=${JSON.stringify(sx)}` : ``}
-
+            aria-label=${title}
```
dgibson666: This isn't the right solution (and it's not quoted, so it's breaking output on spaces). The button is already described by the entire tooltip, including title, so this is verbose. The real issue is that the icon does not have accessibility text describing it. It could be applied to either the icon or the tooltip, but I think applying it to the icon makes it more clear what it's describing.

dgibson666 (commented):
Lots of code comments, because there's a lot of nuance to these accessibility issues.

packages/web-components/src/components/cbp-multicol/cbp-multicol.stories.tsx
```diff
@@ -66,8 +66,8 @@ function createChildren(children) {
 }
 
 const Template = ({ content, columns, width, gap, rule, nobreak, sx }) => {
+  //TODO: removed UL to deal with accessibility flag
   return `
-    <ul>
```
dgibson666, line 70: By removing the ul, you've also removed the default indent of the list. I'd add it back via sx. The browser default is `padding-inline-start: 40px` but I'd use `--cbp-space-4x`, which is what the list component defaults to.

packages/web-components/src/components/cbp-footer/cbp-footer.stories.tsx
```diff
@@ -43,8 +43,8 @@ const InternalTemplate = ({ footerNav }) => {
         
         <cbp-flex gap="var(--cbp-space-4x)" wrap="wrap">
           <span><i>Having an issue?</i></span>
-          <span><b>Email: </b><cbp-link href="mailto:somebody@example.com" context="dark-always">this-application-support@abc.def.gov</cbp-link></span>
-          <span><b>CBP Helpdesk: </b><cbp-link href="tel:555-555-5555" context="dark-always">(555) 555-5555</cbp-link></span>
+          <span><b>Email: </b><cbp-link href="mailto:somebody@example.com" context="dark-always" sx='{"--cbp-link-color-dark":"var(--cbp-color-blue-20)"}'>this-application-support@abc.def.gov</cbp-link></span>
```
dgibson666: Since these are accessibility issues, I would make them part of the component CSS rather than `sx`.

bagrub: added the cssvar overrides to component instead of the sx

packages/web-components/src/components/cbp-multicol/cbp-multicol.stories.tsx
bagrub: updated to have sx manage this padding

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.stories.tsx
bagrub: removed, added accessible label to the icon as discussed in sync

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: added cbp-hide, removed display:none from span

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: removed and added the cbp-hide to wrap the span, removed the display:none

packages/web-components/src/components/cbp-segmented-button-group/cbp-segmented-button-group.stories.tsx
bagrub: removed and added accessibility-text to icon as discussed

packages/web-components/src/components/cbp-multicol/cbp-multicol.tsx
bagrub: removed from component, put in story

packages/web-components/src/components/cbp-action-bar/cbp-action-bar.stories.tsx
bagrub: waiting to discuss with Jeremy as mentioned in sync

packages/web-components/src/components/cbp-multicol/cbp-multicol.stories.tsx
```diff
@@ -85,5 +85,6 @@ export const Multicol = Template.bind({});
 Multicol.args = {
   columns: 2,
   width: '8rem',
-  gap: 'var(--cbp-space-4x)'
+  gap: 'var(--cbp-space-4x)',
+  sx: {"padding":"var(--cbp-space-4x)"}
```
dgibson666: this should be padding-inline-start. Not padding on all sides.

packages/web-components/src/components/cbp-segmented-button-group/cbp-segmented-button-group.stories.tsx
```diff
@@ -102,21 +101,21 @@ export const SegmentedButtonGroupIcons = IconTemplate.bind({});
 SegmentedButtonGroupIcons.args = {
   buttons: [
     {
-      label: '<cbp-icon name="user"></cbp-icon>',
+      label: '<cbp-icon name="user" aria-label="User"></cbp-icon>',
```
dgibson666: This is not correct. You should use the accessibility-text prop and understand how it applies. The cbp-icon tag has no semantic meaning and will never read an aria-label on it.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -60,9 +60,10 @@ function generateSelectableItems(items, context){
         type="checkbox"
         name="checkbox"
         value=${index}
-        aria-label= ${content}
       />
-      <span style='display: none'>Checkbox ${index}</span>
+      <cbp-hide>
```
dgibson666: you're not using the visually-hidden attribute, so this still wouldn't work. Also, you should be able to remove the span.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -632,7 +633,7 @@ const StructuredListSelectableTemplate = ({ listItems, striped, showHeader, head
           ${context && context != 'light-inverts' ? `context=${context}` : ''}   
           ${sx ? `sx=${JSON.stringify(sx)}` : ''}
         >
-        ${showHeader ? `<div slot="cbp-structured-list-header" id="${headerId}"><cbp-checkbox><input type="checkbox" aria-label="Select All"/><span style='display: none'>check all</span></cbp-checkbox> <span>5 Search Results</span></div>` : ''}
+        ${showHeader ? `<div slot="cbp-structured-list-header" id="${headerId}"><cbp-checkbox><input type="checkbox"/><cbp-hide><span>check all</span></cbp-hide></cbp-checkbox> <span>5 Search Results</span></div>` : ''}
```
dgibson666: you're not using the visually-hidden attribute, so this still wouldn't work. Also, you should be able to remove the span.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.stories.tsx
```diff
@@ -55,7 +54,7 @@ Tooltip.args = {
     alignment: 'top-left',
     title: 'Test Tooltip Title',
     content: 'Stub text for tooltip.',
-    tooltipControl: '<cbp-icon name="user"></cbp-icon>',
+    tooltipControl: '<cbp-icon name="user" aria-label="User"></cbp-icon>',
```
dgibson666: You should use the accessibility-text prop and understand how it applies. The cbp-icon tag has no semantic meaning and will never read an aria-label on it.

dgibson666 (commented):
See code comments

packages/web-components/src/components/cbp-multicol/cbp-multicol.stories.tsx
bagrub: amended to be padding-inline-start

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: Updated to use visually-hidden on the hide component & removed the span tag

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: Updated to use visually-hidden on the hide component & removed the span tag

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.stories.tsx
bagrub: updated to use the accessibility-text prop instead of the aria-label

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -61,8 +61,9 @@ function generateSelectableItems(items, context){
         name="checkbox"
         value=${index}
       />
-      <cbp-hide>
-        <span>Checkbox ${index}</span>
+      <cbp-hide
+      visually-hide=true>
```
dgibson666: boolean attribute in HTML doesn't take a value. It's just there like "checked".

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -633,7 +634,7 @@ const StructuredListSelectableTemplate = ({ listItems, striped, showHeader, head
           ${context && context != 'light-inverts' ? `context=${context}` : ''}   
           ${sx ? `sx=${JSON.stringify(sx)}` : ''}
         >
-        ${showHeader ? `<div slot="cbp-structured-list-header" id="${headerId}"><cbp-checkbox><input type="checkbox"/><cbp-hide><span>check all</span></cbp-hide></cbp-checkbox> <span>5 Search Results</span></div>` : ''}
+        ${showHeader ? `<div slot="cbp-structured-list-header" id="${headerId}"><cbp-checkbox><input type="checkbox"/><cbp-hide visually-hide=true>check all</cbp-hide></cbp-checkbox> <span>5 Search Results</span></div>` : ''}
```
dgibson666: boolean attribute in HTML doesn't take a value. It's just there like "checked".

dgibson666 (commented):
See code comments about boolean attrs

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: updated

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: updated

packages/web-components/src/components/cbp-tag/cbp-tag.scss
```diff
@@ -6,11 +6,11 @@
  */
  :root{
   --cbp-tag-color: var(--cbp-color-info-lighter);
-  // --cbp-tag-color-dark: var(--cbp-color-info-darker);
-  --cbp-tag-color-dark: var(--cbp-color-blue-cool-80);
+  --cbp-tag-color-dark: var(--cbp-color-info-dark);
+  // --cbp-tag-color-dark: var(--cbp-color-blue-cool-80);
```
dgibson666: remove commented line

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
```diff
@@ -1015,7 +1015,7 @@ const InternalTemplate = ({ isLoggedIn, username, hashid, navItems, search, sear
       </nav>
 
       <section>
-        <cbp-typography tag="h6" variant="heading-md" context="dark-always" sx='{"margin-bottom":"var(--cbp-space-2x)"}'>Application Support</cbp-typography>
+        <cbp-typography tag="span" variant="heading-md" context="dark-always" sx='{"margin-bottom":"var(--cbp-space-2x)"}'>Application Support</cbp-typography>
```
dgibson666, line 1018: This needs to be updated in the search results archetype too.

dgibson666 (commented):
see code comments

packages/web-components/src/components/cbp-tag/cbp-tag.scss
bagrub: removed

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
bagrub: updated in the search results as well

dgibson666 (approved):

packages/web-components/src/components/cbp-action-bar/cbp-action-bar.stories.tsx
bagrub: discussed with Jeremy in requirements today & implemented discussed changes

packages/web-components/src/components/cbp-segmented-button-group/cbp-segmented-button-group.stories.tsx
bagrub: update to be the accessibility-text prop as per below conversation

PR #383, opened 2025-12-15
Move logic out of componentDidLoad and into componentDidRender

bagrub:
Update to Card to address techdebt about reactive changes in the componentDidLoad


dgibson666 (approved):
looks good

PR #386, opened 2025-12-23
Bugfix/hide operators

dgibson666:
* Updated cbp-hide docs about complex queries.
* Updates to `cbp-dropdown`:
  * Only emit the `valueChange` event when triggered by the user from the built-in counter pill control.
  * Add an optional `selected` (boolean) key to the `items` JSON, so that items may be easily selected by default when loading in a new set of dropdown items.
  * Added a debug flag and debug logs for testing integration issues.


bagrub (approved):
LGTM

PR #389, opened 2025-12-30
Bugfix/app header errors no search

dgibson666:
* Delete old dark mode toggle css in templates and archetypes
* Fix typos on segmented button group closing tag on templates and archetypes.
* Update search results archetype with new dark mode switcher
* Fix errors in app-header when search is not enabled.


bagrub (approved):
LGTM, approved

PR #390, opened 2025-12-31
Creation of Treeview

bagrub:
Create treeview component for the CBP design system


packages/web-components/src/components/cbp-treeview/cbp-treeview.stories.tsx
```diff
@@ -0,0 +1,123 @@
+export default {
+    title: 'Components/Treeview',
+    tags: ['new'],
+    argTypes: {
+        label : {
+            description: 'Specifies Text for the Treeview, is displayed with the treeview-controls',
+            control: 'text',
+        },
+        open: {
```
dgibson666: This isn't doing anything. You should be passing a JSON object of items. You should probably have more stories to show each of the variations we discussed yesterday.

packages/web-components/src/components/cbp-treeview/api-doc.mdx
```diff
@@ -0,0 +1,6 @@
+import { Meta, Markdown } from "@storybook/addon-docs/blocks";
+import Docs from './readme.md?raw';
```
dgibson666: This filename should be api-docs.mdx to be consistent. 
This needs to import the child docs too.

packages/web-components/src/components/cbp-treeview/cbp-treeview.specs.mdx
```diff
@@ -0,0 +1,46 @@
+import { Meta, Markdown } from "@storybook/addon-docs/blocks";
+
+<Meta title="Components/Treeview/Specifications" />
+
+# cbp-treeview
+
+## Purpose
+
+*Display information in a tiered or nested display with parents and child relationships with the option to either select or take action on items
```
dgibson666: need a space to show up as a bullet. Please review this file in Storybook for formatting, capitalization, and punctuation.

packages/web-components/src/components/cbp-treeview/cbp-treeview.tsx
```diff
@@ -0,0 +1,18 @@
...
+  tag: 'cbp-treeview',
+  styleUrl: 'cbp-treeview.scss',
+})
+
+export class CbpTreeview {
+
+  render() {
+    return (
+      <Host>
+
```
dgibson666: I'm not sure what the requirements are around a label, but a visible label is different from an aria-label. The aria-label (if not visible) should be an `accessibilityText` prop as in other components. A visible label  needs to be referenced via `aria-labelledby` on the host, pointing to the label text node by id. You would never use both and aria-labelledby would override aria-label if you used both, so the latter would be pointless.

packages/web-components/src/components/cbp-treeview/cbp-treeview.tsx
```diff
@@ -0,0 +1,18 @@
...
+@Component({
+  tag: 'cbp-treeview',
+  styleUrl: 'cbp-treeview.scss',
+})
+
+export class CbpTreeview {
+
+  render() {
+    return (
+      <Host>
```
dgibson666: probably need a role on the host tag (consult accessibility docs)

packages/web-components/src/components/cbp-treeview/cbp-treeview.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+            description: 'Supports adding inline styles as an object of key-value pairs comprised of CSS properties and values. Values should reference design tokens when possible.',
+            control: 'object',
+        },
+    },
+  };
+
+const Template = ({ label, open, context, sx }) => {
+    return ` 
+        <cbp-treeview>
+            <cbp-treeview-item
```
dgibson666: All of the items should be rendered by a self-referencing function call that iterates over the object passed in.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,50 @@
...
+            onClick={() => {this.toggleOpen()}}
+            targetProp="open"
+          >
+            <cbp-icon name="caret-down"></cbp-icon>
+          </cbp-button>
+          <slot name="treeview-checkbox-control"></slot>  
+          <span>{this.label}</span>
+          <slot name="treeview-button-control"></slot> 
+        </span>
+        <div class="cbp-treeview-content">
```
dgibson666: There isn't really "content" so I would call this "children" instead to be more clear.

A Button is a control, so the name is needlessly long and redundant. Drop "control".

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,50 @@
...
+            variant="square"
+            color="secondary"
+            fill="ghost"
+            class="cbp-treeview-toggle"
+            onClick={() => {this.toggleOpen()}}
+            targetProp="open"
+          >
+            <cbp-icon name="caret-down"></cbp-icon>
+          </cbp-button>
+          <slot name="treeview-checkbox-control"></slot>
```
dgibson666: these should still use cbp-treeview as the namespace (not just treeview). And don't forget to document slots.

A Checkbox is a control, so the name is needlessly long and redundant. Drop "control".

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,50 @@
...
+            fill="ghost"
+            class="cbp-treeview-toggle"
+            onClick={() => {this.toggleOpen()}}
+            targetProp="open"
+          >
+            <cbp-icon name="caret-down"></cbp-icon>
+          </cbp-button>
+          <slot name="treeview-checkbox-control"></slot>  
+          <span>{this.label}</span>
+          <slot name="treeview-button-control"></slot>
```
dgibson666: these should still use cbp-treeview as the namespace (not just treeview). And don't forget to document slots.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,50 @@
...
+  
+/**
+ * Label to be displayed in the control of the treeview.
+ */
+@Prop({ reflect: true}) label: string;
+
+/**
+ * determines if the component is in an open or expanded state
+ */
+@Prop({reflect: true}) open: boolean = false;
```
dgibson666: this needs to be mutable as well since you are changing it by internal logic.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,50 @@
+import { Component, Host, h, Prop } from '@stencil/core';
```
dgibson666: You should also be emitting events for when an item is expanded/collapsed. Selections should probably be emitted from the parent level component, so it can aggregate all of the checked items into one event.

packages/web-components/src/components/cbp-treeview/cbp-treeview.tsx
```diff
@@ -0,0 +1,18 @@
+import { Component, Host, h } from '@stencil/core';
```
dgibson666: Need an "accessibilityText" prop to provide an aria-label on the host tag.

Need Event and EventEmitter to emit the values of all checked items when one is changed (selected/deselected).

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,50 @@
...
+    return (
+      <Host>
+        <span class="cbp-treeview-control">
+          <cbp-button
+            variant="square"
+            color="secondary"
+            fill="ghost"
+            class="cbp-treeview-toggle"
+            onClick={() => {this.toggleOpen()}}
+            targetProp="open"
```
dgibson666: drop this - you are doing the work in this component, so the targetProp has no effect. (it requires a "controls" prop or something to make it work and is intended for loosely coupled components, not tightly coupled like this case)

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
```diff
@@ -0,0 +1,82 @@
...
+    & > .cbp-treeview-control{
+
+      .cbp-treeview-toggle cbp-icon{
+        transform: rotate(180deg)
+      }
+    }
+  }
+
+  cbp-treeview-item{
+    --cbp-treeview-item-indent: 11px; //Spec specified values for indent outside of avaliable CSSVars
```
dgibson666: the base indent (per child level) should be the root value. This should be deleted as a duplicate.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
```diff
@@ -0,0 +1,82 @@
...
+      top: (calc(var(--cbp-space-2x) * -1));
+      left: calc(var(--cbp-treeview-connection-width) * -1);
+      width: var(--cbp-treeview-item-indent);
+      margin-inline-start: calc(var(--cbp-treeview-item-indent) * -1);
+      height: calc(var(--cbp-treeview-indicator-spacing) + 1px);
+      border: solid var(--cbp-treeview-connection-color);
+      border-width: 0 0 var(--cbp-treeview-connection-width) var(--cbp-treeview-connection-width);
+    }
+
+    cbp-treeview-item{
```
dgibson666: there should be only 1 base indent value for each child (that compounds from parents who are also children). All of these are extraneous and not needed.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
```diff
@@ -0,0 +1,82 @@
...
+    min-height: var(--cbp-space-11x);
+    align-items: center;
+    gap: var(--cbp-space-3x);  
+  
+    cbp-button:not(.cbp-treeview-toggle){ //Techdebt: slotted buttons not part of MVP, needs to be centered with text of treeview-item
+      --cbp-button-height: var(--cbp-space-6x);
+    }
+  }
+
+  &:not([open]) .cbp-treeview-content{
```
dgibson666: This reverse logic is confusing. Just make it display: none by default and change it when open.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
```diff
@@ -0,0 +1,82 @@
...
+}
+
+[data-cbp-theme=light] cbp-treeview-item[context*=dark]:not([context=light-always]),
+[data-cbp-theme=dark] cbp-treeview-item:not([context=dark-inverts]):not([context=light-always]) {
+
+  --cbp-tooltip-color-bg: var(--cbp-tooltip-color-bg-dark);
+ }
+
+cbp-treeview-item {
+  display: block;
```
dgibson666, line 32: Wouldn't you make this whole thing flex? I thought it was necessary for the buttons at the end to line up properly. The rest should fall into place as long as you sent grow and shrink to 0 and the width: 100%.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
```diff
@@ -0,0 +1,82 @@
+/**
+ * @prop --cbp-
```
dgibson666: document so they show up in API docs

dgibson666 (commented):
Aside from the high level comments we discussed, see code comments.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
```diff
@@ -0,0 +1,72 @@
...
+  --cbp-treeview-indicator-spacing: 2.75rem;
+  --cbp-treeview-connection-width: 2px;
+  --cbp-treeview-connection-color: var(--cbp-color-gray-cool-20);
+  --cbp-treeview-connection-color-dark: var(--cbp-color-gray-cool-40);
+}
+
+[data-cbp-theme=light] cbp-treeview-item[context*=dark]:not([context=light-always]),
+[data-cbp-theme=dark] cbp-treeview-item:not([context=dark-inverts]):not([context=light-always]) {
+
+  --cbp-tooltip-color-bg: var(--cbp-tooltip-color-bg-dark);
```
dgibson666: delete. Is there anything that is affected by dark mode? The line colors?

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
```diff
@@ -0,0 +1,72 @@
...
+      }
+    }
+  }
+
+  cbp-treeview-item{
+    --cbp-treeview-item-indent: 11px; //Spec specified values for indent outside of avaliable CSSVars
+    border-left: var(--cbp-treeview-connection-width) solid var(--cbp-treeview-connection-color);
+    --cbp-treeview-item-indent: 2.25rem;
+
+    &::before{
```
dgibson666: This should probably be made more flexible by pointing to the center of the item. When text wraps, both the expand control and the checkbox stay centered.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * determines if treeviewItem is a parent to other treeviewItems
+ */
+@Prop({reflect: true}) parent: boolean; //TODO: remove to use name fore the parentNode ln 43
+
+/**
+ * children treeview items of a parent treeview item
+ */
+@Prop({mutable: true}) childrenItems;
```
dgibson666: this shouldn't be a prop

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,190 @@
...
+  styleUrl: 'cbp-treeview-item.scss',
+})
+
+export class CbpTreeviewItem {
+@Element() host: HTMLCbpTreeviewItemElement;
+
+/**
+ * determines if treeviewItem is a parent to other treeviewItems
+ */
+@Prop({reflect: true}) parent: boolean; //TODO: remove to use name fore the parentNode ln 43
```
dgibson666: this shouldn't be a prop

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * determines if the component is in an open or expanded state
+ */
+@Prop({reflect: true, mutable: true}) open: boolean = false;
+
+/**
+ * determines if the control for the treeview-item is slotted, if true the default checkbox will not be rendered
+ */
+@Prop() slottedControl: boolean;
```
dgibson666: not needed as a prop. Make a "selectable" prop instead to render checkboxes.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * used to determing if the treeviewItem is in a checked state
+ */
+@Prop({mutable: true}) treeviewItemChecked: boolean; //TODO: does this need to be updated from boolean to true/indeterimate/false ?
```
dgibson666: simplify to `selected` and another `indeterminate` prop. Both can be passed down to the rendered checkbox then.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+@Event() treeviewItemUpdate: EventEmitter;
+
+private children;
+
+private parentNode; //TODO: rename this to parent and refactor out the Prop
```
dgibson666: Change to `parent`

Need variables for immediate children and all children queried in componentWillLoad.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+@Listen('stateChanged')
+  handleCheck(e){
+    e.stopPropagation();
+
+    let checkbox = e.target;
+
+
+    this.treeviewItemChecked = checkbox.checked;
+    this.host.setAttribute('aria-selected', checkbox.checked)
```
dgibson666: I think I read that you should use aria-checked if multiple selections are possible and aria-selected if only 1 selection is allowed. This shouldn't be needed as setAttribute though - if it's going off of a prop, it should be taken care of in the render function.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,190 @@
...
+    
+  @Listen('treeviewItemUpdate')
+    handleTreeviewItemUpdate(e){ //TODO: can get some weird states for 'unchecking' children not updating parent/grandparents to correct state
+      if(this.host == e.detail.parent){
+
+      console.log('event: ', e)
+
+      let selectedChildren = 0;
+
+      for (let i = 0; i < this.children.length; i++){
```
dgibson666: this should be a simple Array.from(querySelectorAll()) and the count is the length.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,190 @@
...
+    let childNodes = this.content.children;
+    this.children = Array.from(childNodes).filter(element => {
+      return element.tagName === 'CBP-TREEVIEW-ITEM';
+    }) as HTMLCbpTreeviewItemElement[];
+  }
+}
+
+
+  componentWillLoad(){
+    let node = this.host.parentNode as HTMLElement;
```
dgibson666: this is an unnecessary step. just use this.host.parentNode.closest()

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,190 @@
...
+    this.parentNode = node.closest("cbp-treeview-item") //TODO: should return closest parent cbp-treeview item
+  }
+
+  componentDidRender(){
```
dgibson666: this should not be necessary and causes an extra re-render.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,190 @@
...
+    if(this.content){
+      this.childrenItems = this.content.children.length;
+    }
+  }
+
+  render() {
+
+    if(this.host.querySelector('cbp-treeview-item')){
```
dgibson666: not necessary

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,190 @@
...
+      this.parent = true;
+    }
+  
+    return (
+      <Host
+        role="treeitem"
+        aria-expanded={this.open ? 'true' : 'false'}
```
dgibson666: aria-expanded belongs on the control, not the treeview-item. As mentioned above, should have aria-checked on the host.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,190 @@
...
+      >
+        <span class="cbp-treeview-control">
```
dgibson666: Use the full component name as the namespace - this should be `cbp-treeview-item-control`

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,190 @@
...
+            fill="ghost"
+            class="cbp-treeview-toggle"
+            onClick={() => {this.toggleOpen()}}
+            targetProp="open"
+          >
+            <cbp-icon name="caret-down"></cbp-icon>
+          </cbp-button>
+          } 
+          
+          {!this.slottedControl &&
```
dgibson666: this should look at a `selectable` prop instead.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,190 @@
...
+            <cbp-checkbox ref={(el) => this.checkbox = el as HTMLCbpCheckboxElement}>
+              <input type="checkbox" name="checkbox" />
```
dgibson666: the label text should be implemented as the checkbox label when selectable and plain text when not selectable.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,190 @@
...
+            </cbp-checkbox>
+          }
+
+          <span>{this.childrenItems ? this.label + ' (' + this.childrenItems + ')' : this.label}</span>
```
dgibson666: use template literal format, not string concatenation. You don't need to store a variable with the count, as it can be derived from the array length.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+          {this.slottedControl &&
```
dgibson666: I would not put this slot behind a condition at all.

dgibson666 (commented):
See code comments.

packages/web-components/src/components/cbp-treeview/cbp-treeview.stories.tsx
```diff
@@ -0,0 +1,128 @@
+export default {
+    title: 'Components/Treeview',
+    tags: ['new'],
+    argTypes: {
```
dgibson666: need a control for `accessibilityText` and `selectable`

packages/web-components/src/components/cbp-treeview/cbp-treeview.stories.tsx
```diff
@@ -0,0 +1,128 @@
...
+        ${generateTreeviewitems(tree)}
+    </cbp-treeview>
+    `;
+};
+
+
+export const Treeview = Template.bind({});
+
+Treeview.args = {
+    icon: false,
```
dgibson666: remove this

packages/web-components/src/components/cbp-treeview/cbp-treeview.tsx
```diff
@@ -0,0 +1,34 @@
...
+  tag: 'cbp-treeview',
+  styleUrl: 'cbp-treeview.scss',
+})
+
+export class CbpTreeview {
+  
+  /**
+   * Label to be displayed in the control of the treeview.
+   */
+  @Prop({ reflect: true}) accessibilitylabel: string;
```
dgibson666: this should be `accessibilityText` for consistency and it should not be reflected. Get the docs/description from another component for consistency.

packages/web-components/src/components/cbp-treeview/cbp-treeview.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  render() {
+    return (
+      <Host
+        role="tree"
+        aria-label={this.accessibilitylabel}
+      >
+        
+        <slot name="label"></slot>
```
dgibson666: remove this slot.

packages/web-components/src/components/cbp-treeview/cbp-treeview.tsx
```diff
@@ -0,0 +1,34 @@
...
+        <cbp-typography
```
dgibson666: remove all of this this - the visible label is not part of this component.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
```diff
@@ -0,0 +1,112 @@
...
+  --cbp-tooltip-color-bg: var(--cbp-tooltip-color-bg-dark);
+ }
+
+cbp-treeview-item {
+  display: block;
+  padding-inline-start: var(--cbp-treeview-item-indent);
+  margin-inline-start: calc((var(--cbp-treeview-expand-width) - var(--cbp-treeview-connection-width)) / 2);
+
+  .cbp-treeview-control{
+    display: inline-flex;
```
dgibson666, line 40: all of your direct flex children (maybe except the checkbox?) need flex-shrink: 0 to prevent misalignment when you shrink it down and cause text wrapping.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
```diff
@@ -0,0 +1,112 @@
...
+      margin-inline-start: calc((var(--cbp-treeview-expand-width) / 2) * -1);// TODO: don't love, look for better answer
+    }
+  }
+}
+
+
+  cbp-treeview-item{
+    --cbp-treeview-item-indent: 11px; //Spec specified values for indent outside of avaliable CSSVars
+    border-left: var(--cbp-treeview-connection-width) solid var(--cbp-treeview-connection-color);
+    --cbp-treeview-item-indent: 1.5rem;
```
dgibson666: the base indent (per child level) should be the root value.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+  styleUrl: 'cbp-treeview-item.scss',
+})
+
+export class CbpTreeviewItem {
+@Element() host: HTMLCbpTreeviewItemElement;
+
+/**
+ * Label to be displayed in the control of the treeview item.
+ */
+@Prop({ reflect: true}) label: string;
```
dgibson666: should not be reflected.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+  handleCheck(e = undefined){
+    e?.stopPropagation();
+
+    let checkbox = e?.target || this.checkbox;
+
+    this.checked = checkbox.checked;
+
+    this.allChildren.forEach(a =>{
+      let childCheckbox = a.querySelector('cbp-checkbox');
+      checkbox.indeterminate = false;
```
dgibson666: This should be setting the treeview-item's indeterminate prop, not checkbox. Checkbox gets updates via render. This may be why indeterminate states are not being removed properly when checking/unchecking parents.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+@Listen('stateChanged')
+  handleCheck(e = undefined){
+    e?.stopPropagation();
+
+    let checkbox = e?.target || this.checkbox;
+
+    this.checked = checkbox.checked;
+
+    this.allChildren.forEach(a =>{
```
dgibson666: don't use "a" - use something meaningful like "item"

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+      let childCheckbox = a.querySelector('cbp-checkbox');
```
dgibson666: you don't need to deal with the checkboxes.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+      checkbox.indeterminate = false;
+
+      if (checkbox.checked){
```
dgibson666: none  of this logic is needed. item.checked = checkbox.checked (and indeterminate = false)

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+      }else{
+        a.checked = false;
+        childCheckbox.checked = false;
+      }
+    })
+
+    this.updateParent.emit({
+      host: this.host,
+      parent: this.parent,
+      checked: checkbox
```
dgibson666: shouldn't this be checkbox.checked?

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+  @Listen('updateParent')
+    handleUpdateParent(e){
+      if(this.host == e.detail.parent){
+
+      let selectedChildren = 0;
+      let allSelectedChildren = 0;
+      let indeterimateChild = false;
+
+      for (let i = 0; i < this.children.length; i++){
+        if(this.children[i].checked){
```
dgibson666: you don't need to loop and count twice. these should be simple query selectors and a comparison

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+    return (
+      <Host
+        role = "treeitem"
+        id = {this.uid}
+        aria-selected = {this.checkbox}
+      >
+        <span class="cbp-treeview-control">
+          {this.directChildren > 0 && 
+            <cbp-button
+              variant="square"
```
dgibson666: you don't need the variant if you're overriding the size.

You do need expanded = {`"${this.open}"`} because this is an expand control.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+              class="cbp-treeview-toggle"
+              onClick={() => {this.toggleOpen()}}
+            >
+              <cbp-icon name="caret-down"></cbp-icon>
+            </cbp-button>
+          }
+          {!this.selectable &&
+            <cbp-checkbox 
+              ref={(el) => this.checkbox = el as HTMLCbpCheckboxElement}
+              checked = {this.checked}>
```
dgibson666: indeterminate should be handled by the render as well.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+@Event() updateParent: EventEmitter;
+
+/** Specifies a unique `ID` for the dialog, used to wire up the controls and accessibility features. */
+@Prop() uid: string;
+
+private children = [];
+private allChildren = [];
+
+/**displays number of direct children slotted into parent */
+private directChildren;
```
dgibson666: You shouldn't need this. Just use this.children.length

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+            >
+              <cbp-icon name="caret-down"></cbp-icon>
+            </cbp-button>
+          }
+          {!this.selectable &&
+            <cbp-checkbox 
+              ref={(el) => this.checkbox = el as HTMLCbpCheckboxElement}
+              checked = {this.checked}>
+              <input type="checkbox"/>
+              {this.directChildren ? this.label + ' (' + this.directChildren + ')' : this.label}
```
dgibson666: this shouldn't be a ternary. always output this.label, then use a logical and for the count.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+            </cbp-checkbox>
+          }
+        </span>
+          <div 
+            class="cbp-treeview-content" 
+            role='group'
```
dgibson666: use double quotes

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+componentDidLoad(){
+  if(this.checked){
+    this.handleCheck()
+  }
+}
+
+  componentWillLoad(){ 
+    this.parent = this.host.parentElement.closest("cbp-treeview-item")
+    
+    if(this.host.lastElementChild != null){
```
dgibson666: remove - not needed.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+      this.directChildren = this.host.children.length
+    }
+
+    this.children = [...this.host.children]
```
dgibson666: This should be a querySelectorAll. You are making assumptions with this.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+    this.allChildren = this.host.querySelectorAll('cbp-treeview-item') as unknown as HTMLCbpTreeviewItemElement[];
```
dgibson666: why the "as unknown"? Shouldn't be needed. This should be Array.from() and your variable at the top should be typed (then you don't need it here).

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+  
+    return (
+      <Host
+        role = "treeitem"
+        id = {this.uid}
+        aria-selected = {this.checkbox}
+      >
+        <span class="cbp-treeview-control">
+          {this.directChildren > 0 && 
+            <cbp-button
```
dgibson666, line 152: this button has no accessible label. it should also be aria-labelledby the checkbox/text like the treeview-item

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
```diff
@@ -0,0 +1,112 @@
+/**
+ * @prop --cbp-treeview-item-indent: 0;
```
dgibson666, line 4: you have fewer variables documented than listed below.

dgibson666 (commented):
see code comments. some old ones are still there and still valid.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+              variant="square"
+              color="secondary"
+              fill="ghost"
+              class="cbp-treeview-toggle"
+              onClick={() => {this.toggleOpen()}}
+            >
+              <cbp-icon name="caret-down"></cbp-icon>
+            </cbp-button>
+          }
+          {!this.selectable &&
```
dgibson666: This is wrong - you're rendering a checkbox if it's NOT selectable. Add a control and set it to true in the story. But this seems weird at the individual item level. Maybe the parent treeview should have a selectable prop that can filter down for ease of implementation (by the developer)? Might require discussion.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,187 @@
...
+            <cbp-checkbox 
+              ref={(el) => this.checkbox = el as HTMLCbpCheckboxElement}
+              checked = {this.checked}>
+              <input type="checkbox"/>
```
dgibson666: need a name prop to optionally provide a name to the checkbox. Probably need a value too, as something needs to be passed for checked items...

dgibson666 (commented):
See additional code comments.

I think we might also need to discuss how this component conveys its values to an app. It probably needs a top-level EventEmitter that grabs all checked items and values.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+/** 
+ * Specifies a unique `ID` for the dialog, used to wire up the controls and accessibility features. 
+ */
+@Prop({mutable: true}) uid: string;
+
+/**
+ * Sets the value for the checkbox rendered in treeviewItem
+ */
+@Prop() checkboxValue: string;
```
dgibson666: This should just be `value`

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+@Event() updateParent: EventEmitter;
+@Event() updatedState: EventEmitter; //TODO: might need a different name
+
+
+private children = [] as HTMLCbpTreeviewItemElement[];
```
dgibson666: these don't need "as". AS is for casting a different type to something else. This should be:
```
private children = HTMLCbpTreeviewItemElement[] = [];
```

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,211 @@
...
+private allChildren = [] as HTMLCbpTreeviewItemElement[];
+
+private parent;
```
dgibson666: needs a type

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,211 @@
...
+      host: this.host,
+      parent: this.parent,
+      checked: checkbox.checked
+    })
+  }
+    
+  @Listen('updateParent')
+    handleUpdateParent(e){
+      
+      if(e.detail.parent == null){
```
dgibson666: I don't understand this logic at all. This item is listening for update parent. e.detail.parent should never be null. This should never run.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,211 @@
...
+        }
+        
+        this.updatedState.emit({
+          host: this.host,
+          selected: selected
+        })
+      }
+
+      if(this.host == e.detail.parent){
+
```
dgibson666: fix your indentation - format the whole doc.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,211 @@
...
+      for (let i = 0; i < this.children.length; i++){
+        if(this.children[i].checked){ 
+        selectedChildren= selectedChildren + 1;
+        }else if(this.children[i].indeterminate){
+          indeterimateChild = true;
+        }
+      }
+
+      for(let i = 0; i < this.allChildren.length; i++){
```
dgibson666: Here through 140 still needs to be refactored. You don't need to do anything with checkboxes. All prop updates happen on the treeview item and re-render the checkbox. 

This for look seems unnecessary. You should only need to compare the count of all children to selected children. If =, if 0, else indeterminate.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,211 @@
...
+      >
+        <span class="cbp-treeview-control">
+          {this.children.length > 0 && 
+            <cbp-button
+              color="secondary"
+              fill="ghost"
+              class="cbp-treeview-toggle"
+              onClick={() => {this.toggleOpen()}}
+              aria-labelledby = {`${this.uid}-checkbox`}
+            >
```
dgibson666, line 160: my previous comment is gone, but this is still missing an expanded prop

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,211 @@
...
+              <cbp-icon name="caret-down"></cbp-icon>
+            </cbp-button>
+          }
+            <cbp-checkbox 
+              ref={(el) => this.checkbox = el as HTMLCbpCheckboxElement}
+              checked = {this.checked}
+              indeterminate = {this.indeterminate}
+              name={this.name}
```
dgibson666, line 166: needs a value too. a name with no value doesn't buy us anything.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,211 @@
...
+                type="checkbox"
+                value={this.checkboxValue}
+              />
+                {this.label}
+                {this.children.length > 0 && `(${this.children.length})`}
+            </cbp-checkbox>
+          
+        </span>
+          <div 
+            class="cbp-treeview-content"
```
dgibson666: this should be `cbp-treeview-item-content`

packages/web-components/src/components/cbp-treeview/cbp-treeview.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * Label to be displayed in the control of the treeview.
+   */
+  @Prop({ reflect: true}) accessibilityText: string;
+
+  /**
+   * text to be rendered as UID for the component
+   */
+  @Prop() uid: string
```
dgibson666: UID doersn't make sense at the parent level. These need to be unique per item.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,211 @@
...
+    this.handleCheck()
+  }
+}
+
+  componentWillLoad(){ 
+    this.parent = this.host.parentElement.closest("cbp-treeview-item");
+    this.children = Array.from(this.host.querySelectorAll(':scope > cbp-treeview-item'));
+    this.allChildren = Array.from(this.host.querySelectorAll('cbp-treeview-item'));
+
+    this.name ??= this.host.closest('cbp-treeview').name
```
dgibson666: this and selectable should be pushed down from the top, not here.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,211 @@
...
+    this.uid ??= this.host.closest('cbp-treeview').uid
```
dgibson666: UID does not get inherited from the parent or it won't be unique.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+/**
+ * name to be passed to the rendered checkbox prop
+ */
+@Prop({mutable: true}) name: string;
+
+/** 
+ * Specifies a unique `ID` for the dialog, used to wire up the controls and accessibility features. 
+ */
+@Prop({mutable: true}) uid: string;
```
dgibson666: this needs to import CreateNameSpaceKey and generate an ID automatically if none is provided.

packages/web-components/src/components/cbp-treeview/cbp-treeview.tsx
```diff
@@ -0,0 +1,64 @@
...
+  @Listen('updatedState')
+  handleUpdatedState(e){
+    let combined = [...this.selectedChildren, ...e.detail.selected];
+    this.selectedChildren = combined.filter((item, index) => {
+      return combined.indexOf(item) === index;
+    });
+  
+    this.treeviewSubmit.emit({
+      host: this.host,
+      selected: this.selectedChildren,
```
dgibson666: this isn't really form data or useful. The checkboxes need values and this event should have a `values` key that is an array of objects that contain name/value pairs. If a name or a value is missing, it doesn't get added to the object.

Maybe for same-named keys, roll them into an array. I feel like I did that somewhere, but am not sure where offhand.

packages/web-components/src/components/cbp-treeview/cbp-treeview.tsx
```diff
@@ -0,0 +1,64 @@
...
+})
+
+export class CbpTreeview {
+  
+  @Element() host: HTMLCbpTreeviewElement;
+  
+  /**
+   * determines if the control renders with a checkbox as part of the treeview-item control
+   */
+  @Prop({ reflect: true }) selectable: boolean;
```
dgibson666: I don't think this needs to be reflected

packages/web-components/src/components/cbp-treeview/cbp-treeview.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * unique identifier to prefix the treeview-item control name
+   */
+  @Prop() name: string
+
+  /**
+   * Label to be displayed in the control of the treeview.
+   */
+  @Prop({ reflect: true}) accessibilityText: string;
```
dgibson666: doesn't need to be reflected

packages/web-components/src/components/cbp-treeview/cbp-treeview.tsx
```diff
@@ -0,0 +1,64 @@
...
+  
+  @Element() host: HTMLCbpTreeviewElement;
+  
+  /**
+   * determines if the control renders with a checkbox as part of the treeview-item control
+   */
+  @Prop({ reflect: true }) selectable: boolean;
+
+  /**
+   * unique identifier to prefix the treeview-item control name
```
dgibson666: not unique.

packages/web-components/src/components/cbp-treeview/cbp-treeview.stories.tsx
```diff
@@ -0,0 +1,141 @@
...
+        ${children ? generateTreeviewitems(children) : ''}
+        </cbp-treeview-item>`
+    });
+    return html.join('');
+}
+
+const Template = ({tree, name, accessibilityText, uid,  context, sx }) => {
+    return `
+    <cbp-treeview
+        accessibilityText=${accessibilityText}
```
dgibson666: this isn't working because you need dash-case in html.

packages/web-components/src/components/cbp-treeview/cbp-treeview.stories.tsx
```diff
@@ -0,0 +1,141 @@
...
+    `;
+};
+
+
+export const Treeview = Template.bind({});
+
+Treeview.args = {
+    tree: [
+        {
+            label: "Parent Level A 1",
```
dgibson666: need value keys in the objects

packages/web-components/src/components/cbp-treeview/cbp-treeview.specs.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+* labels will wrap incases where nessecary, intended UX guidance is to keep this wrap to 2 lines where possible
+
+### Accessibility
+
+* All tree nodes are contained in or owned by an element with role "tree"
+* Each element serving as a tree node has role "treeitem"
+* Each Parent element contains or owns an element with role "group"
+* Each element with the role treeitme that serves as a parent node has aria-expanded set to "false" when the node is in a closed state and "true" when it is in an open stat
+* If tree supports selection of more than one node, the element with role "tree" has aria-multiselectable set to "true"
```
dgibson666: I don't see aria-multiselectable in your component.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -0,0 +1,211 @@
...
+    this.uid ??= this.host.closest('cbp-treeview').uid
+  }
+
+  render() {
+  
+    return (
+      <Host
+        role = "treeitem"
+        id = {this.uid}
+        aria-selected = {this.checkbox}
```
dgibson666: this needs to be this.checked.

dgibson666 (commented):
see code comments

bagrub (commented):
LGTM

dgibson666 (approved):
I think this is in good shape for the first cut. TechDebt and enhancements noted in a Jira ticket.

PR #393, opened 2026-01-06
Docs/dark mode accessibility

dgibson666:
* Update dark mode docs
* Disable backgrounds/grid on all stories because cbp-app background covers it anyway


bagrub (approved):
LGTM

PR #398, opened 2026-01-26
Bugfix/docs private variables

dgibson666:
* Updated docs
* Added private to all host and state references for document generation
* Removed host where not used.


bagrub (approved):
lgtm

PR #400, opened 2026-01-29
Bugfix/upgrade storybook 10.2.1 badges plugin

dgibson666:
* Update storybook to latest 10.x
* Update tag-badges plugin to Storybook 10 compatible version
* Fix erroneous sx implementation in multiple components
* Remove scroll snap CSS conflicting with scrollIntoView()
* Create test story for tabs and scrolling issues
* Fix tab scrolling behaviors and side effects
  * Prevent vertical scrolling in most cases by adding `block: "nearest" to scrollIntoView behavior
  * Prevent running scrollIntoView() on component initialization because there's no way to stop vertical scrolling of the page if a tabset is out of view on load.


bagrub (approved):
LGTM

PR #402, opened 2026-02-09
Bugfix/form post2

dgibson666:
* Update slider docs
* Update changelog
* Update form component and stories to handle enhanced file upload (retry using formdata event)


bagrub (approved):
LGTM

PR #404, opened 2026-02-12
Bugfix/file input techdebt and docs

dgibson666:
* Updated treeview event value key and docs
* Updated slider docs
* Updated file input to handle duplicate file selections, not emit an event if the value didn't change (based on de-duping), updated docs.


bagrub (approved):
Looks good to me

PR #405, opened 2026-02-13
Bugfix/updates to user pref drawer

bagrub:
Update user preference drawer pattern to match updated specs. also fixed issue with passenger archetype filter panel's hide range

### DIFF
--- a/packages/web-components/src/components/cbp-section/cbp-section.scss
+++ b/packages/web-components/src/components/cbp-section/cbp-section.scss
@@ -1,3 +1,32 @@
+/**
+ * @prop --cbp-section-border-color: transparent;
+ * @prop --cbp-section-border-color-dark: transparent;
+ * @prop --cbp-section-border-width: var(--cbp-border-size-md);
+ * @prop --cbp-section-border-style: solid;
+ */
+:root {
+  --cbp-section-border-color: transparent;
+  --cbp-section-border-color-dark: transparent;
+  --cbp-section-border-width: var(--cbp-border-size-md);
+  --cbp-section-border-style: solid;
+}
+
+/* 
+ * Dark Mode - display dark design based on mode or context
+ */
+
+[data-cbp-theme=light] cbp-section[context*=dark]:not(#fakeId),
+[data-cbp-theme=dark] cbp-section:not([context=dark-inverts]):not([context=light-always]):not(#fakeId) {
+  --cbp-section-border-color: var(--cbp-section-border-color-dark); //Setting sx on story is overriding this
+}
+
+
+
+
 cbp-section:not([hidden]) {
   display: block;
+  border-width: var(--cbp-section-border-width);
+  border-style: var(--cbp-section-border-style);
+  border-color: var(--cbp-section-border-color);
+  
 }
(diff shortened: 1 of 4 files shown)


packages/web-components/src/stories/templates.stories.tsx
```diff
@@ -524,22 +532,38 @@ function renderUserPref(username) {
...
         
-        <cbp-segmented-button-group
-          id="darkmode"
+       <cbp-section
+          sx='{
+              "padding-block-start":"var(--cbp-space-4x)",
+              "padding-block-end":"var(--cbp-space-3x)",
+              "--cbp-section-color-border":"var(--cbp-color-gray-cool-20)",
+              "--cbp-section-color-border-dark":"var(--cbp-color-gray-cool-60)",
+              "--cbp-section-border-width":"var(--cbp-border-size-md) 0"
```
dgibson666: I thought Jeremy said this should be a 1px border? md is 2px?

packages/web-components/src/components/cbp-section/cbp-section.scss
```diff
@@ -1,3 +1,32 @@
+/**
+ * @prop --cbp-section-color-border: transparent;
+ * @prop --cbp-section-color-border-dark: transparent;
+ * @prop --cbp-section-border-width: var(--cbp-border-size-md);
```
dgibson666: this is showing up incorrect in the API docs. Should be 0

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
```diff
@@ -223,22 +223,38 @@ function renderUserPref(username) {
         </cbp-flex>
         <br />
 
-        <cbp-segmented-button-group
-          id="darkmode"
+        <cbp-section
+          sx='{
+              "padding-block-start":"var(--cbp-space-4x)",
+              "padding-block-end":"var(--cbp-space-3x)",
```
dgibson666, line 228: padding is supposed to be different between top and bottom?

It also might make more sense to apply margins (4x?) than using a `br` on top (no spacing on bottom if other content was added)

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
dgibson666: Shouldn't the "click here to Logout" be a link? Might as well address this since you're updating them all.

dgibson666 (commented):
See code comments.

packages/web-components/src/stories/templates.stories.tsx
bagrub: updated to the small border css var (it is 1px)

packages/web-components/src/components/cbp-section/cbp-section.scss
bagrub: updated to be 0

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
bagrub: updated to apply the 3x margins on the flex component wrapping the logout and remove the br tag. Padding start/end is different due to the spec being made before we decided to make this a section but it is matching what is in spec

packages/web-components/src/stories/Archetype_Passenger.stories.tsx
bagrub: updated to be link component instead of a typography one

dgibson666 (approved):
looks good

PR #409, opened 2026-02-18
add height and width prop to Dialog component, remove the cssvars for…

bagrub:
Add Height and width props to the dialog component. remove the cssvars for associated to those values


dgibson666 (approved):
I made some edits to the docs, but looks good. Approved.

PR #414, opened 2026-02-23
add opacity to resolve flickering issue with toaast is closed

bagrub:
Address issue with Toast flickering when open property is set to false


dgibson666 (approved):
looks good

PR #416, opened 2026-03-02
update to SX implementation in story files and dark mode declaration …

bagrub:
…overrides for CSSVars

Update to add ' wrapping the sx values in storybook files. Also reviewed dark mode CSSVar overrides and added !important as discussed

### DIFF
--- a/packages/web-components/src/components/cbp-nav-item/cbp-nav-item.scss
+++ b/packages/web-components/src/components/cbp-nav-item/cbp-nav-item.scss
@@ -72,21 +72,21 @@
  * for the dark version get written to the light within the scope of this component.
  */
 [data-cbp-theme=dark] cbp-nav-item:not(#fakeId) {
-  --cbp-nav-item-color: var(--cbp-nav-item-color-dark);
+  --cbp-nav-item-color: var(--cbp-nav-item-color-dark) !important;
   
-  --cbp-nav-item-color-hover: var(--cbp-nav-item-color-hover-dark);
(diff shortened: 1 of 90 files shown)


packages/web-components/src/components/cbp-nav-item/cbp-nav-item.scss
```diff
@@ -72,21 +72,21 @@
  * for the dark version get written to the light within the scope of this component.
  */
 [data-cbp-theme=dark] cbp-nav-item:not(#fakeId) {
```
dgibson666: remove the fakeId hack here and anywhere else (the important should do its job, but verify)

packages/web-components/src/components/cbp-toggle/cbp-toggle.scss
```diff
@@ -158,29 +158,29 @@
 [data-cbp-theme=dark] cbp-toggle:not([context=dark-inverts]):not([context=light-always]):not(#fakeId) {
   --cbp-toggle-text-color: var(--cbp-toggle-text-color-dark);
```
dgibson666: you missed this one and remove fakeId

dgibson666 (commented):
Couple code comments in addition to:

* Looks like you missed the Action bar CSS.
* You missed `sx` on the list stories, likely because you're using string concatenation instead of backticks and it didn't get caught by your search. (fix to backticks while you're at it).
* Subnav has an `sx` prop and control, but is not implemented in the story code.
* You missed `sx` on the toggle stories.
* You missed `sx` on the tooltip stories.

dgibson666 (approved):
looks good.

PR #419, opened 2026-03-06
bugfix/banner-remove-flex

dgibson666:
* Removed unnecessary flex in banner, which is causing side effects like making clicking to the right of the button still trigger the button.
* Formatted code


bagrub (approved):
Changes look good, approved

PR #420, opened 2026-03-06
Bugfix/pagination pages emit event

dgibson666:
* Emit `paginationChange` event when page size dropdown is changed
* Make labels visually hidden so they are read properly (this may be revisited to make a SASS mixin later and leverage it here)
* Update docs


bagrub (approved):
Changes look good, approving

PR #421, opened 2026-03-13
Address Techdebt issues in Treeview and Treeview Item components

bagrub:
Update to address techdebt issues outline for Treeview and Treeview Item:
*add Watch to the checked prop on Treeview Item
*refactor the --cbp-treeview-item-indent implementation
*Address connector 'gap' when text wraps
*Add Keyboard functionality (ArrowUp, ArrowDown, Home, End) for treeview navigation
*Create Story for tree with Action buttons
*Add aria-owns attribute to Treeview and Treeview Item parents

### DIFF
--- a/packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
+++ b/packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
@@ -11,7 +11,7 @@
 :root {
   --cbp-treeview-item-color: var(--cbp-color-text-darkest);
   --cbp-treeview-item-color-dark: var(--cbp-color-text-lightest);
-  --cbp-treeview-item-indent: 1.5rem;//0; //1.5rem nested; TechDebt: this should use usable as part of the API, but current implementation prevents it.
+  --cbp-treeview-item-indent: 1.5rem;
   --cbp-treeview-indicator-spacing: 1.5rem;
   --cbp-treeview-connection-width: 2px;
   --cbp-treeview-connection-color: var(--cbp-color-gray-cool-20);
@@ -119,7 +119,7 @@ cbp-treeview-item {
         margin-inline-start: calc(var(--cbp-treeview-item-indent) * -1);
         min-height: calc(var(--cbp-treeview-indicator-spacing) + var(--cbp-treeview-connection-width));
         border: solid var(--cbp-treeview-connection-color);
-        border-width: 0 0 var(--cbp-treeview-connection-width) 0;//var(--cbp-treeview-connection-width);
+        border-width: 0 0 var(--cbp-treeview-connection-width) 0;
       }
     }
   }
--- a/packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
+++ b/packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
@@ -11,7 +11,7 @@
 :root {
   --cbp-treeview-item-color: var(--cbp-color-text-darkest);
   --cbp-treeview-item-color-dark: var(--cbp-color-text-lightest);
-  --cbp-treeview-item-indent: 0; //1.5rem nested; TechDebt: this should use usable as part of the API, but current implementation prevents it.
+  --cbp-treeview-item-indent: 1.5rem;//0; //1.5rem nested; TechDebt: this should use usable as part of the API, but current implementation prevents it.
   --cbp-treeview-indicator-spacing: 1.5rem;
   --cbp-treeview-connection-width: 2px;
   --cbp-treeview-connection-color: var(--cbp-color-gray-cool-20);
@@ -31,7 +31,6 @@
 cbp-treeview-item {
   display: block;
   color: var(--cbp-treeview-item-color);
-  padding-inline-start: var(--cbp-treeview-item-indent);
   margin-inline-start: calc((var(--cbp-treeview-expand-width) - var(--cbp-treeview-connection-width)) / 2);
 
   .cbp-treeview-item-control {
@@ -105,10 +104,9 @@ cbp-treeview-item {
   // Nested items
   cbp-treeview-item {
     border-left: var(--cbp-treeview-connection-width) solid var(--cbp-treeview-connection-color);
-    --cbp-treeview-item-indent: 1.5rem; // TechDebt: this should be the default value, not reassigned here.
+    padding-inline-start: var(--cbp-treeview-item-indent);
 
     .cbp-treeview-item-control {
-      //--cbp-checkbox-margin: 0;
 
       &::before {
         content: '';
@@ -119,14 +117,14 @@ cbp-treeview-item {
         left: calc(var(--cbp-treeview-connection-width) * -1);
         width: var(--cbp-treeview-item-indent);
         margin-inline-start: calc(var(--cbp-treeview-item-indent) * -1);
-        height: calc(var(--cbp-treeview-indicator-spacing) + var(--cbp-treeview-connection-width));
+        min-height: calc(var(--cbp-treeview-indicator-spacing) + var(--cbp-treeview-connection-width));
         border: solid var(--cbp-treeview-connection-color);
-        border-width: 0 0 var(--cbp-treeview-connection-width) var(--cbp-treeview-connection-width);
+        border-width: 0 0 var(--cbp-treeview-connection-width) 0;//var(--cbp-treeview-connection-width);
       }
     }
   }
 
   &:last-child {
-    border-color: transparent;
+    border-image: linear-gradient(to bottom, var(--cbp-treeview-connection-color) 50%, transparent 50%) 1 100%;
   }
 }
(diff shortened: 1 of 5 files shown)


packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
```diff
@@ -11,7 +11,7 @@
 :root {
   --cbp-treeview-item-color: var(--cbp-color-text-darkest);
   --cbp-treeview-item-color-dark: var(--cbp-color-text-lightest);
-  --cbp-treeview-item-indent: 0; //1.5rem nested; TechDebt: this should use usable as part of the API, but current implementation prevents it.
+  --cbp-treeview-item-indent: 1.5rem;//0; //1.5rem nested; TechDebt: this should use usable as part of the API, but current implementation prevents it.
```
dgibson666: remove the comment if it's working (seems to be)

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
```diff
@@ -119,14 +117,14 @@ cbp-treeview-item {
         left: calc(var(--cbp-treeview-connection-width) * -1);
         width: var(--cbp-treeview-item-indent);
         margin-inline-start: calc(var(--cbp-treeview-item-indent) * -1);
-        height: calc(var(--cbp-treeview-indicator-spacing) + var(--cbp-treeview-connection-width));
+        min-height: calc(var(--cbp-treeview-indicator-spacing) + var(--cbp-treeview-connection-width));
         border: solid var(--cbp-treeview-connection-color);
-        border-width: 0 0 var(--cbp-treeview-connection-width) var(--cbp-treeview-connection-width);
+        border-width: 0 0 var(--cbp-treeview-connection-width) 0;//var(--cbp-treeview-connection-width);
```
dgibson666: remove comment. the width does not seem to change. (I recall adding `flex-shrink: 0` somewhere to stop it in the previous PR).

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -124,6 +120,12 @@ export class CbpTreeviewItem {
     }
   }
 
+
+  @Watch('checked')
```
dgibson666: This isn't doing what it needs to do, which is act like an actual click. Emit events, etc. The select all down the tree and parent status up the tree needs to be triggered.

The checkbox is rendered by the component, so you don't need to set its checked state explicitly here - it will be updated in the re-render automatically.

packages/web-components/src/components/cbp-treeview/cbp-treeview.specs.mdx
```diff
@@ -45,9 +46,10 @@ import { Meta, Markdown } from "@storybook/addon-docs/blocks";
   * Has a `role="tree"`.
   * If it is selectable it also has `aria-multiselectable="true"`.
   * Should always contain an aria-label (via `accessibilityText` property) if it does not reference an external heading via `aria-labelledby`.
+  * Has aria-owned populated by root level treeview-items id's
```
dgibson666: Put attributes (aria-owns and id) in backticks like the rest of them.

packages/web-components/src/components/cbp-treeview/cbp-treeview.tsx
```diff
@@ -83,6 +87,30 @@ export class CbpTreeview {
     }
   }
 
+  handleKeyPress(e){   
+    this.treeviewItems = Array.from(this.host.querySelectorAll('cbp-treeview > cbp-treeview-item, cbp-treeview-item[open] > .cbp-treeview-item-children > cbp-treeview-item'));
```
dgibson666, line 91: interesting way to do it - seems to work.

packages/web-components/src/components/cbp-treeview/cbp-treeview.tsx
```diff
@@ -83,6 +87,30 @@ export class CbpTreeview {
...
+
+    const { key } = e;
+    const navKeys = ['ArrowDown', 'ArrowUp', 'Home', 'End']; // all keys that will do the default open action
```
dgibson666: this comment isn't right because it's copy and paste - you can remove it.

packages/web-components/src/components/cbp-treeview/cbp-treeview.tsx
```diff
@@ -101,12 +129,22 @@ export class CbpTreeview {
     });
   }
 
+  componentDidLoad(){
+    this.treeviewItems = Array.from(this.host.querySelectorAll('cbp-treeview > cbp-treeview-item, cbp-treeview-item[open] > .cbp-treeview-item-children > cbp-treeview-item'));
+  }
+
   render() {
+    setTimeout(() => { //set aria-owns based on populated treeview
```
dgibson666: This is problematic - it shouldn't be done every render. Can probably move it to componentDidLoad without the setTimeout? (test)

dgibson666 (commented):
See code comments. Biggest issue is the watch, which isn't doing what it needs to.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
bagrub: Removed

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.scss
bagrub: removed, confirmed the width is changing visually with css var update. Note changes to width cssvar don't affect the indent, that is a separate cssvar which might have been what you where seeing, let me know if you want to discuss/refactor

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
bagrub: Updated as discussed in sync to rework the handleCheck() to leverage watch and new method doCheck(). Added a setTimeOut on doCheck to match one already in updateTreeviewParent() to avoid recursive calls between parent & child. Also set the checked & indeterminate prop to default to false to deal with some strangeness with first click event that presented itself during refactor

packages/web-components/src/components/cbp-treeview/cbp-treeview.specs.mdx
bagrub: Added backticks as requested

packages/web-components/src/components/cbp-treeview/cbp-treeview.tsx
bagrub: removed comment

packages/web-components/src/components/cbp-treeview/cbp-treeview.tsx
bagrub: Moved to componentDidLoad and removed the setTimeout.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
```diff
@@ -120,11 +127,11 @@ export class CbpTreeviewItem {
...
           parent: this.parent,
         })
-      }, 50);
+      }, this.eventTimer);
     }
   }
 
-  toggleOpen() {
+  private toggleOpen() {
     this.open === false ? this.open = true : this.open = false;
```
dgibson666: I didn't notice this before, but this can be written as `this.open = !this.open;`

packages/web-components/src/components/cbp-treeview/cbp-treeview.specs.mdx
```diff
@@ -45,9 +46,10 @@ import { Meta, Markdown } from "@storybook/addon-docs/blocks";
   * Has a `role="tree"`.
   * If it is selectable it also has `aria-multiselectable="true"`.
   * Should always contain an aria-label (via `accessibilityText` property) if it does not reference an external heading via `aria-labelledby`.
+  * Has `aria-owned` populated by root level treeview-items id's
 * Each child `cbp-treeview-item` element:
   * Has a `role="treeitem"`.
-  * With nested items renders the children inside of container with a `role="group"`.
+  * With nested items renders the children inside of container with a `role="group"` and populates an `aria-owned` on the parent's container with the children's id.
```
dgibson666: "on the host" is more clear than "parent's container"

dgibson666 (commented):
Couple small code comments. Looks good overall, but let's drop the @watch - it is causing too much re-evaluation and I'm seeing some broken/buggy behavior in a couple cases.

packages/web-components/src/components/cbp-treeview-item/cbp-treeview-item.tsx
bagrub: updated to the above logic

packages/web-components/src/components/cbp-treeview/cbp-treeview.specs.mdx
bagrub: updated language as requested

dgibson666 (approved):
looks good

PR #425, opened 2026-03-17
Creation of circle variant for Button

bagrub:
Creation of the 'circle' variant of cbp-button. Intended for icon only similar to the square button variant


dgibson666 (approved):
looks good

PR #426, opened 2026-03-19
Bugfix/table responsiveness

dgibson666:
* Got the resize-observer's debounce prop working.
* Updated breadcrumbs to use the debounce prop rather than import
* Implemented table responsiveness (linearization and horizontal scroll)
* Implemented controls that scroll by column.
* Added live region slot and basic hookup, but not full tested. (this will be revisited later because it's hard to test)

* Add story controls for showCaption (default hidden) and live region text, moved action bar outside of `cbp-table` component because it is adversely affected by responsiveness.

* Update table docs with new changes.

### DIFF
--- a/packages/web-components/src/components/cbp-table/cbp-table.tsx
+++ b/packages/web-components/src/components/cbp-table/cbp-table.tsx
@@ -1,11 +1,13 @@
 import { Component, Prop, State, Element, Event, EventEmitter, Method, Host, h } from '@stencil/core';
-import { setCSSProps } from '../../utils/utils';
+import { setCSSProps, createNamespaceKey } from '../../utils/utils';
 
 /**
  * The Table component is a wrapper component encapsulating design system styles for semantic HTML 
  * tables as well as applying progressive enhancements to the contained table.
  * 
  * @slot - The semantic table HTML is placed within the default slot.
+ * @slot cbp-table-live-toolbar - Any sort of filters or table controls may be slotted within this named slot.
+ * @slot cbp-table-live-region - For complex tables with many controls and/or pagination, the state of the data may be quantified and described accessibly via an `aria-live` region, which is hidden from view. E.g., Filtered by the term "test", ordered by Column 1 ascending, displaying records 100-200 of 1234.
  */
 @Component({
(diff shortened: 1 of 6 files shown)


packages/web-components/src/components/cbp-table/cbp-table.tsx
```diff
@@ -58,8 +68,29 @@ export class CbpTable {
     });
   }
 
-  makeSortable(){
-    this.columnHeadings = Array.from(this.host.querySelectorAll('thead th'));
+  // TechDebt: how to make this reactive?
+  private addHeaderDataAttrs() {
+    //const columnHeadings: HTMLElement[] = Array.from(this.host.querySelectorAll('thead th'));
```
bagrub: your declaring this in componentWillLoad() to set, can remove comment for declaration here & in makeSortable() on ln 93

also not seeing the tableCell being used anywhere else, could be removed as well

packages/web-components/src/components/cbp-table/cbp-table.tsx
```diff
@@ -1,11 +1,13 @@
 import { Component, Prop, State, Element, Event, EventEmitter, Method, Host, h } from '@stencil/core';
-import { setCSSProps } from '../../utils/utils';
+import { setCSSProps, createNamespaceKey } from '../../utils/utils';
 
 /**
  * The Table component is a wrapper component encapsulating design system styles for semantic HTML 
  * tables as well as applying progressive enhancements to the contained table.
  * 
  * @slot - The semantic table HTML is placed within the default slot.
+ * @slot cbp-table-live-toolbar - Any sort of filters or table controls may be slotted within this named slot.
```
bagrub: Not seeing the implementation of this toolbar slot mentioned

dgibson666: copy-paste mangled it.

packages/web-components/src/components/cbp-table/cbp-table.tsx
dgibson666: cleaned up those comments

bagrub (approved):
LGTM

PR #428, opened 2026-03-23
update table alignment options and menu options for stories

bagrub:
Update to the table stories to:
* right align Actions column (header and cells)
* update menu options to use view, edit, delete


dgibson666 (approved):
looks good

PR #433, opened 2026-03-24
Bugfix/html panel prerendered code

dgibson666:
* Remove getAbsolutePath use in main.ts
* Update preview.ts with renderPreHydrated decorator to show the pre-rendered web component code.
* Fixed broken story output for drawer stories.

### DIFF
--- a/packages/web-components/.storybook/preview.ts
+++ b/packages/web-components/.storybook/preview.ts
@@ -4,6 +4,8 @@ import type { DecoratorFunction } from 'storybook/internal/csf';
 import { isChromatic } from './chromatic';
 import {defineCustomElements} from '../dist/loader';
 
+import { addons } from 'storybook/preview-api'
+
 // Used for docs
 import { themes } from 'storybook/theming';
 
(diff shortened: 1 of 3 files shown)


packages/web-components/.storybook/preview.ts
```diff
@@ -56,8 +58,42 @@ const withAnimationControl: DecoratorFunction<WebComponentsRenderer> = (storyFn)
   return storyFn();
 };
 
+
+// Get the Pre-rendered Stencil web component code for the HTML Panel
+/* TODO:
```
bagrub, line 63: This looks more like a techdebt than a todo. also can we remove the console.log(context) on ln 70 or is that needed for debug?

packages/web-components/.storybook/preview.ts
```diff
@@ -116,17 +152,17 @@ const preview: Preview = {
...
-        tabWidth: 2,
-        useTabs: false,
-        htmlWhitespaceSensitivity: 'css', // ignore?
-        quoteProps: "as-needed",
-        proseWrap: "always",
-        bracketSameLine: false,
-        singleAttributePerLine: true,
-        printWidth: 80,
-      },
+      //prettier: {
```
bagrub, line 155: Why are we keeping these as commented instead of removing?

bagrub (commented):
some small comments to resolve, but looking good overall

packages/web-components/.storybook/preview.ts
dgibson666: I want to reference our previous settings and/or do comparative testing.

packages/web-components/.storybook/preview.ts
dgibson666: it's not really techdebt. just possible improvements. TODO is accurate.

bagrub (approved):
LGTM

PR #434, opened 2026-03-24
bugfix/html-panel-prerendered-code

dgibson666:
* Updated changelog 
* Updated the HTML panel to reference the proper root element

This means that most stories do not show the `cbp-app` component, but the templates, which specify a higher-level root, do include the `cbp-app` as desired.


bagrub (approved):
LGTM

PR #435, opened 2026-03-25
Update loader styles for firefox, additional update to storynames

bagrub:
*Update loader styling to address issues in firefox browser
*Update story name for page templates, Passenger, and search
*Update drawer CSSVars doc to include --cbp-drawer-close-button-color

### DIFF
--- a/packages/web-components/src/components/cbp-loader/cbp-loader.scss
+++ b/packages/web-components/src/components/cbp-loader/cbp-loader.scss
@@ -110,6 +110,10 @@ cbp-loader {
     align-self: flex-start;
   }
 
+  &:not([determinate]):not([success]):not([error]) progress{  
+    visibility: hidden;
+  } 
+
   // Linear variant
   &[variant=linear] {
     --cbp-icon-size: var(--cbp-space-3x);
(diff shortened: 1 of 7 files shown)


packages/web-components/src/components/cbp-loader/cbp-loader.scss
```diff
@@ -119,26 +123,40 @@ cbp-loader {
...
         background-color: var(--cbp-loader-track-color);
-      }
 
-      &::-webkit-progress-value {
-        background-color: var(--cbp-loader-color);
+        &::-webkit-progress-bar {
+          background-color: var(--cbp-loader-track-color);
+        }
+
+        &::-webkit-progress-value,
```
dgibson666: I think you need to separate these selectors. If one selector in the selector list is invalid, the whole rule is ignored by the browser.

dgibson666 (commented):
Just 1 code comment. I went ahead and fixed the page template structure since I wasn't sure what was "Correct" to comment without doing it.

packages/web-components/src/components/cbp-loader/cbp-loader.scss
bagrub: Separated

dgibson666 (approved):

PR #437, opened 2026-03-25
Add additional color options for structure list item

bagrub:
*Update structured list item color prop to include: warning, success, and info
*Update striped css to not override background if color prop has been set


dgibson666 (approved):
looks good

PR #440, opened 2026-03-30
Chore/story cleanup 2026

dgibson666:
* Cleaned up stories with improper booleans and unquoted values.
* Formatted code.
* Adjusted indenting in many cases to make the HTML panel look better.
* Reorganized component categories as discussed.


bagrub (approved):
LGTM

PR #441, opened 2026-03-31
Resolve techdebt in pagination, toast, and grid

bagrub:
Resolve techdebt:
*Toast: techdebt already resolved, removing techdebt tag and TODO that was resolved
*Grid: techdebt already resolved as part of recent reorg chore
*Pagination: added mixin to global/style/index.scss as discussed in sync

### DIFF
--- a/packages/web-components/src/global/styles/index.scss
+++ b/packages/web-components/src/global/styles/index.scss
@@ -4,7 +4,7 @@
  */
 @forward 'breakpoints';
 
-@mixin hiddenText {
+@mixin visuallyHidden {
     border: 0;
     clip: rect(1px, 1px, 1px, 1px);
     -webkit-clip-path: inset(50%);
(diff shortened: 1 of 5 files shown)


packages/web-components/src/global/styles/index.scss
```diff
@@ -2,4 +2,18 @@
  * Caution: "global styles" get injected into every component and can cause file size bloat.
  * These should only include SASS variables and mixins that are not written out to CSS directly
  */
-@forward 'breakpoints';
\ No newline at end of file
+@forward 'breakpoints';
+
+@mixin hiddenText {
```
dgibson666: Can you please name this "visuallyHidden" - that's what it does. It's visually hidden, but still available to screen readers.

dgibson666 (commented):
Just ask that you rename the mixin to accurately describe what it is. Also, check the cbp-app core.scss (and others) to see if there's a visually-hidden utility class that can also use the mixin.

packages/web-components/src/global/styles/index.scss
bagrub: updated name to above & added to core.scss as requested

dgibson666 (approved):
looks good

PR #442, opened 2026-04-02
Chore/complex table testing

bagrub:
*Added Multi Header Table story to testing sub section.
  *Added styles for colspan on th in the thead to get appropriate look
*Added useRowHeader control to table stories
  *Enabled the sticky styling for the th in the tbody
  *reworked the table cell background color declarations to support scrolling with a sticky row header

### DIFF
--- a/packages/web-components/src/components/cbp-table/cbp-table.scss
+++ b/packages/web-components/src/components/cbp-table/cbp-table.scss
@@ -231,8 +231,10 @@ cbp-table {
   }
 
   // Cell hover effect
+  //".cbp-table-wrapper table tbody tr" selectors to up specificity so hover override the striped styling
   &[hover=cell] {
-    tbody {
+    .cbp-table-wrapper table tbody tr {
+      th:hover,
       td:hover{
           background-color: var(--cbp-table-row-color-bg-hover);
         
--- a/packages/web-components/src/components/cbp-table/cbp-table.scss
+++ b/packages/web-components/src/components/cbp-table/cbp-table.scss
@@ -156,13 +156,6 @@ cbp-table {
       left: 0;
       inset-inline-start: 0;
     }
-    
-    tbody th{
-      &[scope="row"],
-      &[colspan]:not(:last-child){
-        border-right: solid var(--cbp-table-row-border-size) var(--cbp-table-row-color-border);
-      }
-    }
   }
   
   // Responsive linearized table
@@ -423,21 +416,23 @@ cbp-table {
     th {
       font-weight: var(--cbp-font-weight-medium);
 
-      // &[scope="row"],
-      // &[colspan]:not(:last-child){
-      //   border-right: solid var(--cbp-table-row-border-size) var(--cbp-table-row-color-border);
-      // }
-     
       &[colspan]{
         text-align: center;
+        
+        &:not(:last-child){
+        border-right: solid var(--cbp-table-row-border-size) var(--cbp-table-row-color-border);
+      }
       }
     }
-
 
     td {
       text-wrap: pretty;
     }
   }
+  
+  &:not(.cbp-table-linearize ) table th[scope="row"]{
+    border-right: solid var(--cbp-table-row-border-size) var(--cbp-table-row-color-border);
+  } 
 }
 
 // Shows a gradient mask on the table for either direction that has overflowed content
--- a/packages/web-components/src/components/cbp-table/cbp-table.scss
+++ b/packages/web-components/src/components/cbp-table/cbp-table.scss
@@ -389,7 +389,8 @@ cbp-table {
       border-bottom: solid var(--cbp-table-row-border-size) var(--cbp-table-row-color-border);
     }
 
-    th {
+    th,
+    td[scope="row"] {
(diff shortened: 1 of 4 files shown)


packages/web-components/src/components/cbp-table/cbp-table.stories.tsx
```diff
@@ -85,31 +88,32 @@ function generateTableRows(data, selectable, context) {
             </td>
     `;
 
-    let cells = row.map( ({td, danger, highlight, alignmentRight})  => {
+    let cells = row.map( ({td, danger, highlight, alignmentRight}, j)  => {
```
dgibson666, line 91: can we compress all these if into 1 output? It seems like you're just outputting a class, which can be done inline as a ternary.

packages/web-components/src/stories/tests/multiheaderTable-test.stories.tsx
```diff
@@ -0,0 +1,288 @@
...
+            <td>
+              <cbp-checkbox ${context && context != 'light-inverts' ? `context="${context}"` : ''}>
+                <input type="checkbox" name="rowid" value="${i+1}">
+                <span style="display:none">Select row ${i+1}</span>
+              </cbp-checkbox>
+            </td>
+    `;
+
+    let cells = row.map( ({td, danger, highlight, alignmentRight}, j)  => {
+      if(danger){
```
dgibson666: Mirror changes from the main story here (compress ifs into 1 output)

packages/web-components/src/components/cbp-table/cbp-table.scss
```diff
@@ -390,18 +411,23 @@ cbp-table {
...
-      outline-width: 0;
-      outline-style: solid;
-      outline-color: var(--cbp-table-header-color-outline);
-      outline-offset: calc(var(--cbp-space-1x) * -1);
+
+      &[scope="row"]{
+        border-right: solid var(--cbp-table-row-border-size) var(--cbp-table-row-color-border);
+      }
+     
+      &[colspan]{
```
dgibson666, line 421: combine these 2 selectors by using a :not(:last-of-type) on this one. Can't you use :last-child instead since th should be the only valid child?

packages/web-components/src/components/cbp-table/cbp-table.scss
```diff
@@ -370,8 +389,10 @@ cbp-table {
...
     tbody {
       tr {
-        background-color: var(--cbp-table-row-color-bg);
-
+        td, th{
+          background-color: var(--cbp-table-row-color-bg);
+        }
+        
         // Selected state should override all others (striping, danger, highlight)
         &:has(input[type=checkbox]:checked) td {
```
dgibson666: row headers are not showing selected state.

packages/web-components/src/components/cbp-table/cbp-table.scss
```diff
@@ -390,18 +411,23 @@ cbp-table {
...
-      background-color: var(--cbp-table-header-color-bg);
       font-weight: var(--cbp-font-weight-medium);
-      text-wrap: balance;
-      outline-width: 0;
-      outline-style: solid;
-      outline-color: var(--cbp-table-header-color-outline);
-      outline-offset: calc(var(--cbp-space-1x) * -1);
+
+      &[scope="row"]{
+        border-right: solid var(--cbp-table-row-border-size) var(--cbp-table-row-color-border);
```
dgibson666: We don't want this when the table linearizes, however. You might have to rewrite your selector to start at the top and filter out cbp-table.cbp-table-linearize.

dgibson666 (commented):
See code comments. Couple things not working and a few other minor comments.

packages/web-components/src/components/cbp-table/cbp-table.stories.tsx
bagrub: compressed if statements for class into 1 line ternary

packages/web-components/src/stories/tests/multiheaderTable-test.stories.tsx
bagrub: compressed if statements for class into 1 line ternary to match main story

packages/web-components/src/components/cbp-table/cbp-table.scss
bagrub: combined selectors for the border-right & use :last-child for the colspan instead of :last-of-type

packages/web-components/src/components/cbp-table/cbp-table.scss
bagrub: updated to include the th in this selector so row headers get selected state

packages/web-components/src/components/cbp-table/cbp-table.scss
bagrub: Updated to include the border styling in the .cbp-table-scroll starting on ln 137 so that this is not included when table linearizes. tried refactoring as mentioned but was getting alot of regressions, seemed simpler to only include when scrolling

dgibson666 (approved):
I made some small story tweaks, but looks good otherwise.

PR #443, opened 2026-04-03
Bugfix/button sizing

dgibson666:
* Update link and button download prop to allow a boolean or string, optionally specifying the name of the downloaded file.
* npm install and tsconfig tweaks to try to get rid of vscode errors
* Fix VSCode errors and mark test stories to be ignored by Chromatic
* Update Tag CSS to work better with custom cases (`align-items:stretch` rather than center)
* Update Button CSS to work more intuitively and better for custom cases.
  * Removed default minimum height. This was actually causing the Accordion Item to render larger than intended.
  * Kept the CSS vars for min width/height (default to auto) because they are being used by other components.
  * Setting `width` and `height` properties on the component now sets/overrides the corresponding CSS variables, `--cbp-button-width` and `--cbp-button-height` respectively. This is more intuitive and works better with the API. This fixes the Tab overflow issue currently on the live Storybook.
  * Removed display: inline-block on the host tag, which made it less versatile in layout.
  * Also applying sx and height and width to the host now instead of button tag. This makes variable overrides work as expected, while it may break setting button properties directly that aren't represented by the CSS API.

I went through several iterations. Removing the height and width from the host tag completely broke subnav (because as an inline element width 100% of the child doesn't expand it - it only makes it fill the container) so I had to keep it.


bagrub (approved):
LGTM

PR #445, opened 2026-04-09
Test/template with responsive tables

dgibson666:
* Tweak tsconfig settings - set back to "node" and other updates
* Created test story for table responsiveness in a template and grid.
* Add table debugging and fix responsive calculations
* Updated column highlighting to work with multi-row headers (only the last row)


bagrub (approved):
LGTM

PR #446, opened 2026-04-13
update to various stories

bagrub:
update to address the below:
*Update structured list warning dark variables to not fail color contrast for accessibility
*Update accordion story to render h3 tags for danger example
*Consolidate chip stories into 1 and add storybook control for Icon


dgibson666 (approved):
I pushed some story tweaks and also fixed the chip not rendering the name prop as a button attribute. Everything else looks good overall.

PR #448, opened 2026-04-14
update structured list

bagrub:
Update structured list:
*Update status colors to update cssVars instead of setting css properties
*Update striping prop and styling to support even, odd, & none
*Refactor stories to reuse template instead of setting each story to have their own
*Remove the selected property from the Structured list

### DIFF
--- a/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
+++ b/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
@@ -41,7 +41,7 @@ cbp-structured-list-item {
     --cbp-structured-list-color-bg: var(--cbp-structured-list-item-color-bg-danger);
   }
   
-  &[color=warning]:not(:has(input[type=checkbox]:checked)) {
+  &[color=warning] {
     --cbp-structured-list-color: var(--cbp-structured-list-item-color-warning);
     --cbp-structured-list-color-bg: var(--cbp-structured-list-item-color-bg-warning);
   }
--- a/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
+++ b/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
@@ -34,23 +34,22 @@
 
 cbp-structured-list-item {
   display: block;
-  background-color: var(--cbp-structured-list-color-bg);
   color: var(--cbp-structured-list-color);
 
   &[color=danger] {
-    --cbp-structured-list-color-bg: var(--cbp-structured-list-item-color-bg-danger);
+    --cbp-structured-list-item-color-bg: var(--cbp-structured-list-item-color-bg-danger);
   }
   
   &[color=warning] {
     --cbp-structured-list-color: var(--cbp-structured-list-item-color-warning);
-    --cbp-structured-list-color-bg: var(--cbp-structured-list-item-color-bg-warning);
+    --cbp-structured-list-item-color-bg: var(--cbp-structured-list-item-color-bg-warning);
   }
 
   &[color=success] {
-    --cbp-structured-list-color-bg: var(--cbp-structured-list-item-color-bg-success);
+    --cbp-structured-list-item-color-bg: var(--cbp-structured-list-item-color-bg-success);
   }
 
   &[color=info] {
-    --cbp-structured-list-color-bg: var(--cbp-structured-list-item-color-bg-info);
+    --cbp-structured-list-item-color-bg: var(--cbp-structured-list-item-color-bg-info);
   }
 }
--- a/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
+++ b/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
@@ -33,15 +33,13 @@
 }
 
 cbp-structured-list-item {
-  display: block;
-  color: var(--cbp-structured-list-color);
 
   &[color=danger] {
     --cbp-structured-list-item-color-bg: var(--cbp-structured-list-item-color-bg-danger);
   }
   
   &[color=warning] {
-    --cbp-structured-list-color: var(--cbp-structured-list-item-color-warning);
+    --cbp-structured-list-item-color: var(--cbp-structured-list-item-color-warning);
     --cbp-structured-list-item-color-bg: var(--cbp-structured-list-item-color-bg-warning);
   }
 
--- a/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
+++ b/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
@@ -41,7 +41,7 @@ cbp-structured-list-item {
     --cbp-structured-list-color-bg: var(--cbp-structured-list-item-color-bg-danger);
   }
   
-  &[color=warning] {
+  &[color=warning]:not(:has(input[type=checkbox]:checked)) {
     --cbp-structured-list-color: var(--cbp-structured-list-item-color-warning);
     --cbp-structured-list-color-bg: var(--cbp-structured-list-item-color-bg-warning);
   }
--- a/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
+++ b/packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
@@ -8,8 +8,6 @@
  * @prop --cbp-structured-list-item-color-bg-success-dark: var(--cbp-color-success-darker);
  * @prop --cbp-structured-list-item-color-bg-info: var(--cbp-color-info-lighter);
  * @prop --cbp-structured-list-item-color-bg-info-dark: var(--cbp-color-info-darker);
- * @prop --cbp-structured-list-item-color-bg-selected: var(--cbp-color-interactive-selected-light);
- * @prop --cbp-structured-list-item-color-bg-selected-dark: var(--cbp-color-interactive-selected-dark);
 */
 :root {
     --cbp-structured-list-item-color-bg-danger: var(--cbp-color-danger-lighter);
@@ -22,8 +20,6 @@
     --cbp-structured-list-item-color-bg-success-dark: var(--cbp-color-success-darker);
     --cbp-structured-list-item-color-bg-info: var(--cbp-color-info-lighter);
     --cbp-structured-list-item-color-bg-info-dark: var(--cbp-color-info-darker);
-    --cbp-structured-list-item-color-bg-selected: var(--cbp-color-interactive-selected-light);
-    --cbp-structured-list-item-color-bg-selected-dark: var(--cbp-color-interactive-selected-dark);
 }
 
 [data-cbp-theme=light] cbp-structured-list[context*=dark]:not([context=light-always]),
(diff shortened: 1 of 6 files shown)


dgibson666 (commented):
Main things that need fixing still:

* Selected prop should go away and style selected like tables.
* Colors are an interface on structured list items, so create an .scss file for items to hold these.
  * The dark mode selector in the new file will still reference `cbp-structured-list`, not the item because we don't have context on items.
* In your CSS vars, abbreviate "background" as "bg" to be consistent with tables and elsewhere. Striped should be color-bg-striped, not color-striped-background.

packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
```diff
@@ -0,0 +1,75 @@
...
+    --cbp-structured-list-item-color-bg-info: var(--cbp-structured-list-item-color-bg-info-dark) !important;
+    --cbp-structured-list-item-color-bg-selected: var(--cbp-structured-list-item-color-bg-selected-dark) !important;
+}
+
+cbp-structured-list-item {
+    display: block;
+    background-color: var(--cbp-structured-list-color-bg);
+    color: var(--cbp-structured-list-color);
+
+    &[color=danger]:not(:has(input[type=checkbox]:checked)) {
```
dgibson666: Do you need these nots anymore? You shouldn't because the selected selector has higher specificity. Simplify the selector to just the color.

packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
```diff
@@ -0,0 +1,75 @@
...
+      --cbp-structured-list-color-bg: var(--cbp-structured-list-item-color-bg-success);
+    }
+
+    &[color=info]:not(:has(input[type=checkbox]:checked)) {
+      --cbp-structured-list-color-bg: var(--cbp-structured-list-item-color-bg-info);
+    }
+  
+    &:has(input[type=checkbox]:checked){
+
+      background-color: var(--cbp-structured-list-item-color-bg-selected);
```
dgibson666: shouldn't this be the variable, not the direct CSS property?

packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
```diff
@@ -0,0 +1,75 @@
...
+      --cbp-structured-list-border-color: var(--cbp-color-gray-cool-30);
+      --cbp-structured-list-border-color-dark: var(--cbp-color-gray-cool-50);
+
+        &:hover{
+        --cbp-structured-list-item-color-bg-selected: var(--cbp-color-gray-cool-10);
```
dgibson666: Then these should be setting the color/color-bg variables, not the selected variables. Do we need a nested hover selector here at all? This looks wrong.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
```diff
@@ -133,7 +102,7 @@ cbp-structured-list {
 
     &:hover {
       color: var(--cbp-structured-list-color) !important;
```
dgibson666: Can these override the variables too? That would be my first choice if it works.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -35,31 +39,39 @@ export default {
...
-  return html.join('');
-}
-
-function generateItems(items) {
-  const html = items.map(({ content, color, selected }) => {
+function generateLIs(items, selectable, context) {
+  const html = items.map(({ content }, index) => {
     return `
-      <cbp-structured-list-item ${color != 'default' ? `color="${color}"` : ''} ${selected ? `selected` : ''}>
+        <li style="display:flex; align-items:center;">
```
dgibson666: Why is an inline style being hard coded on each `li`? If this is needed, it should go in the CSS, but I feel like this is too heavy handed and should not be here.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
```diff
@@ -145,54 +114,14 @@ cbp-structured-list {
   
   /** Selectable */
 
-  &[selectable] [slot=cbp-structured-list-header] {
+  &:has(input[type=checkbox]) [slot=cbp-structured-list-header] {
     gap: var(--cbp-space-4x);
   }
 
-  &[selectable] cbp-structured-list-item{
+  &:has(input[type=checkbox]) cbp-structured-list-item{
```
dgibson666: I'm confused why the component is at the end of this selector. It's also missing the `li`, which only seems to be working because you've hardcoded inline styles on it.

packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
```diff
@@ -0,0 +1,75 @@
...
+
+    &[color=success]:not(:has(input[type=checkbox]:checked)) {
+      --cbp-structured-list-color-bg: var(--cbp-structured-list-item-color-bg-success);
+    }
+
+    &[color=info]:not(:has(input[type=checkbox]:checked)) {
+      --cbp-structured-list-color-bg: var(--cbp-structured-list-item-color-bg-info);
+    }
+  
+    &:has(input[type=checkbox]:checked){
```
dgibson666: this selector also need to target `li`s, which it isn't doing here. It might have to go back up to the parent component CSS.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.tsx
```diff
@@ -27,11 +27,11 @@ export class CbpStructuredList {
...
 
-  /** Specifies whether the list is "striped,"" with even rows shaded. */
-  @Prop({ reflect: true }) striped: boolean;
+  /** Specifies whether the list is striped, designating whether the colored rows are the odd or even rows (CBP DS standard is even when used). */
+  @Prop({ reflect: true }) striped: 'odd' | 'even';
 
-  /** Specifies whether the list items are selectable (via checkbox - provided by the application). */
-  @Prop({ reflect: true }) selectable: boolean;
+  // /** Specifies whether the list items are selectable (via checkbox - provided by the application). */
+  // @Prop({ reflect: true }) selectable: boolean;
```
dgibson666: Just delete it.

dgibson666 (commented):
I have some more comments. Also, I don't see any modifications to `cbp-structured-list-item`, which needs its `selected` prop removed.

packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
bagrub: removed nots on these selectors

packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
bagrub: has been refactored to structured-list.scss and converted to target a CSSvar

packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
bagrub: Refactored to structured-list.scss and added component level variables for this selected-hover state. We do need the hover here to manage the selected-hover as it differs from default hover styles per visual spec

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
bagrub: updated to target cssvars

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: removed inline styles and added to as part of stylesheet as per below. initially there to support the selectable state with li, which was not initially designed for but became possible when refactoring templates

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
bagrub: refactored selector and included the selectable li here as well

packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
bagrub: moved this selector back to the structured-list.scss to also support the selectable li

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.tsx
bagrub: removed

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
```diff
@@ -118,10 +135,24 @@ cbp-structured-list {
...
+
+  cbp-structured-list-item:has(input[type=checkbox]:checked),
+  li:has(input[type=checkbox]:checked){
+    --cbp-structured-list-color-bg: var(--cbp-structured-list-item-color-bg-selected);
+
+    --cbp-structured-list-border-color: var(--cbp-color-gray-cool-30);
+    --cbp-structured-list-border-color-dark: var(--cbp-color-gray-cool-50);
+
+      &:hover{
+        --cbp-structured-list-color-bg: var(--cbp-structured-list-item-color-bg-selected-hover);
```
dgibson666: aren't all of the hover colors the same? Why is this one a separate variable?

dgibson666 (commented):
1 minor code comment.

Also, I would delete the second and last story since 1) you've made them all selectable (story 2 still has selected key in its object, which isn't valid anymore) and 2) you've made them able to toggle the tags between `li` and `cbp-structures-list-item`.

Be sure to remove `selectable` from all of the item objects.

Also, the selectable items get way too large. You need to bring in the same overrides used in table to alter the checkboxes:

```CSS
    --cbp-checkbox-min-height: 0;
    --cbp-checkbox-margin: 0;
```

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
bagrub: deprecated this Cssvar as per above and discussion

dgibson666 (commented):
Still need to remove the `selected` prop from structured list item. Also remove mentions in the specs. 

I also noticed that if you change a color to warning and then select that row, the text color isn't readable (needs to change).

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
```diff
@@ -100,9 +100,11 @@ cbp-structured-list {
     }
   }
 
-  &[striped=even] [role=list] > *:nth-child(even):not([color]):not([selected]),
-  &[striped=odd] [role=list] > *:nth-child(odd):not([color]):not([selected]) {
-    background-color: var(--cbp-structured-list-color-bg-striped);
+  &[striped=even],
+  &[striped=odd]{
+    [role=list] > *:nth-child(even):not([color]):not(:has(input[type=checkbox]:checked)){
```
dgibson666: you can't nest this part with "even" in it.... doesn't work for odd now. The even and odd parts must be paired in the first part of the selector. You can nest `&:not...`

dgibson666 (commented):
See 1 code comment. You also still have `color: 'default',` in many of the stories, which causes  striping to not work.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.scss
bagrub: refactored to as per above

packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
```diff
@@ -34,23 +34,22 @@
 
 cbp-structured-list-item {
   display: block;
```
dgibson666: this is a duplicate from the parent. delete here.

packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
```diff
@@ -34,23 +34,22 @@
...
-  background-color: var(--cbp-structured-list-color-bg);
   color: var(--cbp-structured-list-color);
```
dgibson666: Shouldn't this also go in the parent CSS file with the background? I would also name it --cbp-structured-list-item-color since it's overridden at the item level.

dgibson666 (commented):
see code comments

dgibson666 (approved):
look good!

packages/web-components/src/components/cbp-structured-list-item/cbp-structured-list-item.scss
bagrub: updated to be --cbp-structured-list-item-color and moved to parent file

PR #449, opened 2026-04-20
updates to card stories and reorganizing action button functionality

bagrub:
* confirm story prop for stretch is populating example correctly
*move color control to all cards stories
*consolidate story templates to reuse general template where possible and add action button templates as story control to applicable stories

### DIFF
--- a/packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
+++ b/packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
@@ -2,6 +2,12 @@ export default {
   title: 'Content/Card',
   tags: ['beta'],
   argTypes: {
+    variant: {
+      name: "Variant",
+      description: "set Variant of the card",
+      control: "select",
+      options: ["generic", "banner", "decision", "flag"]
+    },
     title: {
       name: 'Title (slotted)',
       description: 'Set the title in the banner area of the card',
@@ -12,6 +18,27 @@ export default {
       description: 'Set the body text of the card',
       control: 'text',
     },
+
+    actionsLayout:{
+      name: "Actions Layout",
+      description: "Choose actions layout of the card component",
+      control: "select",
+      options: ["single", "double", "triple"],
+    },
+    actionsFill: {
+      name: "Actions Fill",
+      description: "Choose the fill of the actions in the card component",
+      control: "radio",
+      options: ["solid", "outline", "ghost"],
+      if: {arg: "actionsLayout"}
+    },
+    actionsConfig: {
+      name: "Decision Card Actions",
+      description: "Configure card button labels and colors. Available button colors: `primary`, `secondary`, `tertiary` and `danger`",
+      control: "object",
+      if: {arg: "actionsLayout"}
+    },
+
     stretch: {
       control: 'boolean',
     },
@@ -83,13 +110,13 @@ const renderActions = (layout, fill, color, context, withIcon, headingId,{ btn1,
 const GeneralTemplate = ({ variant, color, title, headingId, bodyText, actionsLayout, actionsFill, actionsConfig, withIcon, stretch, context, sx }) => {
   return ` 
     <cbp-card
-      ${variant ? `variant=${variant}` : ``}
+      ${variant != 'generic'? `variant=${variant}` : ``}
       ${color ? `color="${color}"` : ''}
       ${stretch ? `stretch` : ``}
       ${context && context != 'light-inverts' ? `context="${context}"` : ''}      
       ${sx ? `sx='${JSON.stringify(sx)}'` : ''}
     >
-      <cbp-typography tag="h4" slot="cbp-card-title">
+      <cbp-typography tag="h4" slot="cbp-card-title" id=${headingId}>
         ${withIcon ? `<cbp-icon name="triangle-exclamation" size="1.25rem"></cbp-icon>` : ''}
         ${title}
       </cbp-typography>
@@ -111,7 +138,7 @@ const FlagTemplate = ({ title, color, bodyText, withIcon, context, sx }) => {
       <div slot="cbp-card-flag">
         <img src="https://api.dicebear.com/9.x/personas/svg" alt="Flag Card random image"/>
       </div>
-      <cbp-typography tag="h4" slot="cbp-card-title">
+      <cbp-typography tag="h4" slot="cbp-card-title" >
         ${withIcon ? `<cbp-icon name="triangle-exclamation" size="1.25rem"></cbp-icon>` : ''}
         ${title}
       </cbp-typography>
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
```diff
@@ -211,16 +185,53 @@ const InteractiveTemplate = ({ title, color, disabled, bodyText, withIcon, inter
...
+  },
+};
+GeneralCard.argTypes = {
+actionsLayout:{
+      name: "Actions Layout",
+      description: "Choose actions layout of the card component",
+      control: "select",
+      options: ["single", "double", "triple"],
+    },
+    actionsFill: {
```
dgibson666: these 2 should only be shown conditionally on the Actions Layout selection. It's confusing seeing controls with selections that don't even apply to the story.

dgibson666 (commented):
See note on controls. 

Also, the header ID control is not generating an id on the header. It's generating an `aria-describedby` on the buttons, referencing a nonexistent ID that's supposed to be on the header.

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
bagrub: updated the actionFill & actionConfig to only be visible when actionsLayout is set. I also disabled actionsLayout on the flag, interactive, & interactive radio list stories

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
```diff
@@ -308,60 +337,9 @@ InteractiveCard.argTypes = {
...
-      color: "secondary",
-    },
-    btn3: {
-      label: "Action 3",
-      tag: "button",
-      color: "tertiary",
-    },
-  },
+  actionLayout: {
+    control: false
```
dgibson666, line 300: this is showing a separate control, not disabling the existing one. Perhaps because of differing names.

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
```diff
@@ -2,6 +2,12 @@ export default {
   title: 'Content/Card',
   tags: ['beta'],
   argTypes: {
+    variant: {
+      name: "Variant",
+      description: "set Variant of the card",
+      control: "select",
+      options: ["generic", "banner", "decision", "flag"]
```
dgibson666: Per discussion, decision should not be a variant here or on the component as there is no component logic looking for it. "Generic" should also not be an option. This should only list valid variants.

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
```diff
@@ -74,53 +107,21 @@ const renderActions = (layout, fill, color, context, withIcon, headingId,{ btn1,
   }
 };
 
-const GeneralTemplate = ({ color, title, bodyText, withIcon, context, sx }) => {
+const GeneralTemplate = ({ variant, color, title, headingId, bodyText, actionsLayout, actionsFill, actionsConfig, withIcon, stretch, context, sx }) => {
   return ` 
     <cbp-card
+      ${variant != 'generic'? `variant=${variant}` : ``}
       ${color ? `color="${color}"` : ''}
```
dgibson666: You should not be outputting `color="default"`

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
```diff
@@ -450,7 +428,7 @@ InteractiveRadioList.args = {
   title: "Banner Card Title",
   bodyText: "Here is an example of some supplementary text for this purely informational card",
   interactive: "clickable",
-  variant: "default",
+  variant: "generic",
```
dgibson666: This shouldn't be needed at all. It should be unselected as the default.

dgibson666 (commented):
See code comments

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
bagrub: removed, also suppressed the variant control for this story

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
bagrub: updated to not output default in ternary

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
bagrub: removed the 'generic' from the options

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
bagrub: updated to match the names

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
```diff
@@ -281,36 +260,17 @@ BannerCard.args = {
...
   title: "Card Title",
   bodyText: "Here is an example of some body text for this purely informational card",
 };
 
 FlagCard.argTypes ={
-  actionLayout: {
+  actionsLayout: {
+    control: false
+  },
+  variant:{
```
dgibson666, line 273: Why is variant disabled for the flag story but not for the banner story? I would show the control and have it defaulted to the proper variant.

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
```diff
@@ -240,58 +233,53 @@ DecisionCard.args = {
...
+    control: false
+  }
+}
 
 export const InteractiveCard = InteractiveTemplate.bind({});
 InteractiveCard.args = {
   title: "Banner Card Title",
   bodyText: "Here is an example of some supplementary text for this purely informational card",
   interactive: "clickable",
-  variant: "default",
```
dgibson666, line 294: The radiolist story disables the variant control, but this one doesn't. It probably should because banner cards don't look correct when clickable. Or make them both enable the variant control... We might want to discuss either way for future iterations.

dgibson666 (commented):
just a couple things are still inconsistent.

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
bagrub: discussed on sync, these stories are currently using different templates hence the inconstancy

packages/web-components/src/components/cbp-card/cbp-card.stories.tsx
bagrub: removed the variant control from the interactive story

dgibson666 (approved):
looks good

PR #450, opened 2026-04-22
Bugfix/combobox refactor use input

dgibson666:
* Refactor dropdown comboboxes to use text input control - resulting in quite a few downstream changes.
* Filtering triggered on input event now instead of keydown, allowing for arbitrary character deletion/addition and pasting.
* Update tsconfig to ignore structNullChecks to be consistent with previous coding standards.
* Updated the icon alignment in links to make it more centered.
* Updated listbox to remove Typescript errors.
* Updated docs.


bagrub (approved):
Looking good

PR #451, opened 2026-04-23
Bugfix/pagination max pages

dgibson666:
* Implement maxPages property on pagination.
* Fix dropdown showing "no items" option due to searchString being inadvertently set when it shouldn't.
* Fix TS error in notice story.


bagrub (approved):
LGTM

PR #452, opened 2026-04-23
Bugfix/Toast spacing issues

bagrub:
*Update Toast to have display: none when the open prop is not enabled.
*Add expected functionality to the dismiss button in Toast stories


dgibson666 (approved):
looks good. I made some small tweaks to the story file. feel free to re-test.

PR #457, opened 2026-05-15
update interactive color tokens

bagrub:
*update Primary Dark token to cbp-color-blue-70
*update Primary Darker token to cbp-color-blue-80
*update Secondary Dark token to cbp-color-gray-warm-70
*update Secondary Darker token to cbp-color-gray-warm-80


dgibson666 (approved):
looks good

PR #458, opened 2026-05-19
Update Typography tokens

bagrub:
*update typography tokens to use new font size and line height tokens
*update typography component to use new tokens
*update typography component to have new prop 'size' to consume new tokens


packages/design-tokens/src/tokens/typography.json
```diff
@@ -14,51 +14,132 @@
     "size": {
       "masthead": {
```
dgibson666: I thought we were deprecating the masthead variants

packages/web-components/src/components/cbp-typography/cbp-typography.stories.tsx
```diff
@@ -53,114 +55,136 @@ Typography.argTypes = {
...
+  for(var x=20; x > 0; x--){
+  html +=`
+      <cbp-typography
+        size=${x}
+        ${tag ? `tag="${tag}"` : ''}
+        ${divider != 'none' ? `divider="${divider}"` : ''}
+        ${context && context != 'light-inverts' ? `context="${context}"` : ''}
+        ${sx ? `sx='${JSON.stringify(sx)}'` : ''}
+      >
+        ${text} --> size=${x}
```
dgibson666: get rid of this ascii arrow and use parentheses again. Should also include the variants first, then sizes. Delete all the commented out stuff below.

packages/web-components/src/components/cbp-typography/cbp-typography.stories.tsx
```diff
@@ -33,11 +33,13 @@ export default {
   },
 };
 
-const Template = ({ text, tag, variant, divider, context, sx }) => {
+const Template = ({ text, tag, variant, size, divider, context, sx }) => {
+  console.log('size: ', size)
```
dgibson666: remove console.log()

dgibson666 (commented):
See code comments. I think we need to discuss in standup. I thought we wanted additional controls for font-weight on the component and I'm not sure if the line-height should be coupled with size in the CSS. I feel like they should be separate props.

packages/design-tokens/src/tokens/typography.json
bagrub: deprecated and removed masthead tokens, props, and css

packages/web-components/src/components/cbp-typography/cbp-typography.stories.tsx
bagrub: replaced as requested & removed commented out code. Also added the variants back to the story, seperating the size and variants with an HR

packages/web-components/src/components/cbp-typography/cbp-typography.stories.tsx
bagrub: removed

packages/web-components/src/components/cbp-typography/cbp-typography.scss
```diff
@@ -56,14 +54,208 @@ cbp-typography {
     //text-wrap: balance; // not at this time
   }
 
+  &[size="1"]{
```
dgibson666: You don't need the quotes at all on their selectors. You're inconsistent with them too - using single quotes down below. Just remove them all.

Also, these (size, weight, and line height)  do not appear to be working at all, regardless of whether a variant is set. Shouldn't they all use the same `> *` selector as the variant? The style should go on the child tag, not the host.

packages/web-components/src/components/cbp-typography/cbp-typography.scss
```diff
@@ -101,68 +293,55 @@ cbp-typography {
...
-    line-height: var(--cbp-space-11x);
-  }
-
-  &[variant=masthead-2] > * {
-    font-size: var(--cbp-font-size-masthead-2);
-    font-weight: var(--cbp-font-weight-thin);
-    line-height: var(--cbp-space-10x);
-  }
-
   &[variant=heading-xxl] > *, h1 {
```
dgibson666: all of the variant declarations should be before the new props to leverage the cascade. The weight, size, and line-height should override the variant values.

packages/web-components/src/components/cbp-typography/cbp-typography.stories.tsx
```diff
@@ -51,116 +54,65 @@ Typography.argTypes = {
...
-    <cbp-typography
-      ${tag ? `tag="${tag}"` : ''}
-      variant = "body-text"
-      ${divider != 'none' ? `divider="${divider}"` : ''}
-      ${context && context != 'light-inverts' ? `context="${context}"` : ''}
-      ${sx ? `sx='${JSON.stringify(sx)}'` : ''}
-    >
-      ${text} (body-text)
-    </cbp-typography>
+  for(var x=20; x > 0; x--){
```
dgibson666: Let's make these go from smallest to largest so the jarringly large 20 is at the bottom.

packages/web-components/src/components/cbp-typography/cbp-typography.tsx
```diff
@@ -19,8 +19,17 @@ export class CbpTypography {
...
   
   /** Specifies the visual style of the text regardless of the semantic tag. */
-  @Prop({ reflect: true }) variant: 'masthead-1' | 'masthead-2' | 'heading-xxl' | 'heading-xl' | 'heading-lg' | 'heading-md' | 'heading-sm' | 'heading-xs' | 'body-text' | 'subhead' | 'heading-dialog';
+  @Prop({ reflect: true }) variant: 'heading-xxl' | 'heading-xl' | 'heading-lg' | 'heading-md' | 'heading-sm' | 'heading-xs' | 'body-text' | 'subhead' | 'heading-dialog';
 
+  /** Specifies the visual size of a the text which will set the font-size */
+  @Prop({ reflect: true }) size: '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9' | '10' | '11' | '12' | '13' | '14' | '15' | '16' | '17' | '18' | '19' | '20';
+
+  /** Specifies the line-height of a the text */
+  @Prop({ reflect: true }) lineheight: '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9' | '10' | '11' | '12' | '13' | '14' | '15' | '16' | '17' | '18';
```
dgibson666: this should be lineHeight (and the attribute is line-height)

packages/web-components/src/components/cbp-typography/cbp-typography.tsx
```diff
@@ -19,8 +19,17 @@ export class CbpTypography {
...
+
+  /** Specifies the font-weight of a the text */
+  @Prop({ reflect: true }) fontweight: 'thin' | 'light' | 'regular' | 'medium' | 'bold' | 'black' ;
```
dgibson666: this should be fontWeight and the attribute is font-weight.

packages/web-components/src/components/cbp-typography/cbp-typography.scss
```diff
@@ -56,14 +54,208 @@ cbp-typography {
...
+    font-size: var(--cbp-font-size-19);
+    line-height: var(--cbp-line-height-17);
+  }
+
+  &[size="20"]{
+    font-size: var(--cbp-font-size-20);
+    line-height: var(--cbp-line-height-18);
+  }
+
+  &[lineheight='1']{
```
dgibson666: I'm wondering if we can leverage attr() to reduce  the duplication here. (and maybe above some too)

It's worth a try to verify: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/attr

dgibson666 (commented):
See code comments. The new props don't seem to be working at all for me.

packages/web-components/src/components/cbp-typography/cbp-typography.scss
bagrub: the numeric values are breaking build without the quotes, standardized them to all be double quotes and removed from the font-weight ones. 

Also added the > * on all the new selectors

packages/web-components/src/components/cbp-typography/cbp-typography.scss
bagrub: moved new prop declarations to after the variant declarations. seems to have solved overriding issue mentioned in above comment as well

packages/web-components/src/components/cbp-typography/cbp-typography.stories.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-typography/cbp-typography.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-typography/cbp-typography.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-typography/cbp-typography.scss
bagrub: Was not able to get working initially, but like the idea to reduce duplication. Can we sync and discuss?

packages/web-components/src/components/cbp-typography/cbp-typography.stories.tsx
```diff
@@ -79,7 +79,7 @@ Typography.argTypes = {
 // TechDebt: This can be more efficiently achieved by looping over an array of all variants; refactor when we add the additional variants/tokens.
 const AllStyles = ({ text, tag, divider,  context, sx }) => {
   
-  var variants = [ 'heading-xxl', 'heading-xl', 'heading-lg', 'heading-md', 'heading-sm', 'heading-xs', 'body-text', 'subhead'];
+  var variants = [ 'subhead', 'body-text', 'heading-xs', 'heading-sm', 'heading-md', 'heading-lg', 'heading-xl', 'heading-xxl',];
```
dgibson666: I didn't say to reverse this one. This one shows hierarchy by going largest to smallest. The other sizes are just arbitrarily huge to start. So revert this list order.

dgibson666 (commented):
Everything seems to be working now. Just revert the variants list order and it's good I think.

packages/web-components/src/components/cbp-typography/cbp-typography.stories.tsx
bagrub: reverted, thought we wanted both to be descending in original request

dgibson666 (approved):

PR #459, opened 2026-05-28
Update icons to font awesome 7

bagrub:
*Update icon svg mappings to font awesome 7
*added missing icons to story control


dgibson666 (commented):
This looks good, but there are a number of non-included SVGs embedded into stories and some embedded in CSS (because you can't use a component for a background). These should all be updated to the proper FA7 icons and viewboxes as well. 

I think the trash can icon used in table and menu stories should probably be added to the component (and those stories updated to use it). Let's review if any others should.

Search the code base for `<svg`.

I've included this PR in the changelog in my current docs branch.

dgibson666 (commented):
There are chevrons in the dropdown CSS that were not updated still. The one used by pagination is different - I think it would be good to comment the icon name when used directly in CSS for future reference.

dgibson666 (approved):
Looks good. Some of the icons are noticeably smaller. I tried adjusting the viewbox, but didn't see a discernable change.

PR #460, opened 2026-06-02
Deprecate universal header style and replace with flex component

bagrub:
Address techdebt in cbp-universal-header.scss
*update stories, templates, and archetypes to use cbp-flex and cbp-flex-item instead of ul and li
*updating note for deprecating the styles w/ date of change


dgibson666 (approved):
looks good. Technically I don't think we need the flex-item tags at all, but it's all slotted and helps prevent unintended behavior from the flex context applied to the buttons or other children.

PR #462, opened 2026-06-08
Feature/multiselect dropdown pocs

dgibson666:
* Fix sizing issue on multi-select (CSS).
* Add additional dropdown stories and multi-select with chips code (from POC).
* Fix issues with updating selected items from value when changed (for chips use case).


bagrub (approved):
LGTM

PR #466, opened 2026-06-16
Feature/toast updates

bagrub:
*update toast minimum width to 20rem
*update toast container to add css var for width and control width of toasts inside
*update toast stories to use inverted context to style slotted buttons
*removed css vars overrides for cbp-buttons inside of toast
*update tabs to have border between tabs and panel
*update icon mapping to use font awesome 7 svg set not the full svg set


dgibson666 (commented):
Everything looking good, but if you set a duration on the multiple toasts story, it doesn't work. Probably not hooked up right in the story.

dgibson666 (approved):
looks good

PR #467, opened 2026-06-17
update for drawer story a11y

bagrub:
*remove duplicate aria-labeledby in the panel story
*remove aria-labeledby from panel in drawer story
*add accessibility text value defaults to drawer stories

### DIFF
--- a/packages/web-components/src/components/cbp-drawer/cbp-drawer.stories.tsx
+++ b/packages/web-components/src/components/cbp-drawer/cbp-drawer.stories.tsx
@@ -66,7 +66,6 @@ const Template = ({ position, withIcon, open, persistAt, uid, accessibilityText,
           slot="cbp-panel-header"
           tag="h3"
           variant="heading-lg"
-          id="panelheader"
         >
           ${withIcon ? `<cbp-icon name='triangle-exclamation'></cbp-icon>` : ''}
           Drawer Header
@@ -120,7 +119,6 @@ const UserPreferencesTemplate = ({ position, open, persistAt, uid, accessibility
           slot="cbp-panel-header"
           tag="h3"
           variant="heading-lg"
-          id="userprefs-panel-header"
         >
         
           ${withIcon ? `<cbp-icon name='user'></cbp-icon>` : ''}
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/cbp-drawer/cbp-drawer.stories.tsx
dgibson666: let's delete the ID as well here since it's hardcoded and not exposed through a control. It's not referenced at all now, so really no need for it. This one in particular is generic and easy to get collisions from copy-paste.

dgibson666: let's delete the ID as well here since it's hardcoded and not exposed through a control. It's not referenced at all now, so really no need for it.

packages/web-components/src/components/cbp-panel/cbp-panel.stories.tsx
```diff
@@ -50,7 +50,6 @@ export default {
 const PanelTemplate = ({ role, headingLevel, header, headerId, content, ariaLabel, showIcon, context, sx }) => {
   return `
     <cbp-panel
-      ${headerId ? `aria-labelledby="${headerId}"` : ''}
```
dgibson666, line 53: Let's not just remove this. If the panel has a role, then it needs to be labelled. So add `role != 'none'` to this condition and keep it.

dgibson666 (commented):
see code comments.

packages/web-components/src/components/cbp-drawer/cbp-drawer.stories.tsx
bagrub: removed id

bagrub: removed id

packages/web-components/src/components/cbp-panel/cbp-panel.stories.tsx
bagrub: ln 53 is a dup, there is another one on ln 55. update the ternary on ln 55 to have the role != 'none' in its condition

dgibson666 (approved):

PR #468, opened 2026-06-24
expand and volume slider fixes

dgibson666:
* Fix slotted label on expand component. (improperly getting hidden attribute applied due to implementation)
* Added slotted label control to story.
* Fix volume slider icons to be the same size after icon updates.


bagrub (approved):
LGTM

PR #475, opened 2026-07-09
Feature/floating UI

bagrub:
*Add a functional component to interface with the float ui library
*update cbp-menu component to use the above functional component


package.json
```diff
@@ -48,6 +48,7 @@
     "./packages/react-components"
   ],
   "dependencies": {
+    "@floating-ui/dom": "^1.7.6",
```
dgibson666: this should be in the web components package next to Stencil.

packages/web-components/src/utils/floating-placement.tsx
```diff
@@ -0,0 +1,33 @@
+import {computePosition, flip, offset, shift} from '../../../../node_modules/@floating-ui/dom';
```
dgibson666: you shouldn't need the full path to the node modules directory.

This also isn't a functional component because you're not importing it from stencil:

`import { FunctionalComponent, h } from '@stencil/core';`

And it doesn't have a JSX tag used in the menu render. (I think having a dash in the file name is problematic because only web components should use a dash, so make this camelCase.)

Probably more to discuss about the implementation and if it needs to be a functional component or not.

dgibson666 (commented):
Initial comments. I will test functionality tomorrow.

dgibson666 (approved):
looks good

package.json
bagrub: updated as requested

packages/web-components/src/utils/floating-placement.tsx
bagrub: discussed in sync, holding off on converting to functional component currently. Will iterate on this to meet needs of dropdown & tooltip and evaluate if needed as part of that work

PR #476, opened 2026-07-13
Update dot inidcator and structured list stories

bagrub:
*update dot indicator css vars to have light/dark variable mappings for border color
*update structured list stories to address text size issues on the Media Calendar story

### DIFF
--- a/packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.scss
+++ b/packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.scss
@@ -11,8 +11,8 @@
  */
 
 :root {
-  --cbp-dot-indicator-border-color: var(--cbp-color-interactive-secondary-darker);
-  --cbp-dot-indicator-border-color-dark: var(--cbp-color-interactive-secondary-lighter);
+  --cbp-dot-indicator-color-border: var(--cbp-color-interactive-secondary-darker);
+  --cbp-dot-indicator-color-border-dark: var(--cbp-color-interactive-secondary-lighter);
 
   --cbp-dot-indicator-color-hover: var(--cbp-color-interactive-secondary-darker);
   --cbp-dot-indicator-color-hover-dark: var(--cbp-color-interactive-secondary-lighter);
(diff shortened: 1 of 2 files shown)


packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
```diff
@@ -259,6 +259,7 @@ StructuredListMediaCalendar.args = {
                       "width":"100%",
                       "height":"4.25rem",
                       "text-align":"center",
+                      "align-content":"center",
```
dgibson666: this is part of the flex API so use the prop instead of doing it in sx.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
dgibson666: The fact that text is being sized here means that it should be using the typography component, not just doing it on the flex component.

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.scss
```diff
@@ -11,6 +11,9 @@
  */
 
 :root {
+  --cbp-dot-indicator-border-color: var(--cbp-color-interactive-secondary-darker);
```
dgibson666: You're not following naming conventions. It should be color-border.

dgibson666 (commented):
functionally looks good, but see code comments.

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: discussed in sync, keeping the sx align-content. Possible later ticket to update this story to use something other than the cbp-flex

packages/web-components/src/components/cbp-structured-list/cbp-structured-list.stories.tsx
bagrub: removed this sx & updated the text to use the the cbp-typography component

packages/web-components/src/components/cbp-dot-indicator/cbp-dot-indicator.scss
bagrub: updated css var to be color-border

dgibson666 (approved):
I pushed some small changes for formatting and cleanup. Looks good.

PR #477, opened 2026-07-15
Bugfix/icon alignment

dgibson666:
* Simplify icon CSS after testing.
* Remove nested flex context for icons in links and buttons.
* Update icons embedding in CSS with proper FA7 versions in dropdown and form-field CSS.
* Update panel story with different icon and proper margin.
* Make all card, panel, and dialog headings use flex for proper icon alignment, removing margins from story where needed.
* Update icons embedding in CSS with proper FA7 versions in dropdown and form-field CSS.
* Fix dropdown and other custom form fields when font-size is not default by using relative token in the calculation.
* Defensively prevent flex context from affecting the shape of the badge.

I am still evaluating whether to add `vertical-align: text-top` to the default CSS of `cbp-icon`. This seems to make alignment in the flow of text with default icon/body text size work best.

### DIFF
--- a/packages/web-components/CHANGELOG.md
+++ b/packages/web-components/CHANGELOG.md
@@ -4,6 +4,13 @@ This CHANGELOG.md tracks the updates to the web components package of the CBP de
 
 The React components are wrappers generated from this package and will share the same changes. Projects using React 19 may use the native web components without React wrappers.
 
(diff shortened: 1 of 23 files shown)


bagrub (approved):
LGTM

PR #479, opened 2026-07-23
Feature/update tooltip to use float UI util

bagrub:
*Update cbp-tooltip component to consume the floatPlacement util component
*update floatPlacement util to enable arrow
*update cbp-tooltip alignment prop to better match values in floatPlacement


bagrub:
As mentioned in sync, working to update:
*tooltip.tsx to use a ref not queryselector when passing args to floatPlacement
*update floatPosition to use an attribute instead of a class to style arrow off of

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -21,9 +22,8 @@ export class CbpTooltip {
   /** used to set styles for the definition link for text controls*/
   @Prop({ reflect: true }) variant: 'definition';
 
-
   /** sets where the tooltip will be displayed and where the caret will be placed */
-  @Prop({ reflect: true}) alignment: "top-left" | "top-center" | "top-left" | "right-top" | "right-center" | "right-bottom" | "bottom-left" | "bottom-center" | "bottom-right" | "left-top" | "left-center" | "left-bottom" = "top-center";
+  @Prop({ reflect: true}) alignment: "top-start" | "top" | "top-end" | "right-start" | "right" | "right-end" | "bottom-start" | "bottom" | "bottom-end" | "left-start" | "left" | "left-end" = "top-start";
```
dgibson666: Let's rename this to "position" while we're at it - consistent with menu. The default was top (center), so keep it - don't change that functionality. I think it is the most common anyway.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -77,6 +77,26 @@ export class CbpTooltip {
     } 
   }
 
+componentDidRender(){
+    if(this.open){
+   
+    const control = this.host.querySelector('[role="button"] > *:first-child') as HTMLElement
+    const floatingEl = this.host.querySelector('[role="tooltip"]') as HTMLElement
```
dgibson666: these should both be refs as well. If it's rendered by the component (and not slotted), it's a ref.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -77,6 +77,26 @@ export class CbpTooltip {
...
+
+    const floatUiprops: floatUIProps= {
+        placement: this.alignment,
+        offset: {
+          mainAxis: 16,
+        },
+        flip: true,
+        shift: true,
+        arrow: this.host.querySelector('#arrow') as HTMLElement
```
dgibson666: use ref, as discussed.

dgibson666 (commented):
See code comments

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: updated prop name to be "position" instead of alignment

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: updated to use ref for arrow as discussed

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: updated to us refs instead of querySelectors

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
```diff
@@ -289,25 +154,36 @@ div[role=tooltip] {
       line-height: var(--cbp-line-height-xs);
     }
   }
-}
 
+  #arrow{
```
dgibson666: this cannot be an ID. That would mean you can only ever have 1 tooltip per page. It should be a namespaced class like `cbp-tooltip-arrow`

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -14,16 +15,18 @@ import { setCSSProps, createNamespaceKey, getInvertedContext } from '../../utils
...
   /** When set, specifies that the tooltip is open */
   @Prop({ reflect: true }) open: boolean = false;
 
   /** used to set styles for the definition link for text controls*/
   @Prop({ reflect: true }) variant: 'definition';
 
-
   /** sets where the tooltip will be displayed and where the caret will be placed */
-  @Prop({ reflect: true}) alignment: "top-left" | "top-center" | "top-left" | "right-top" | "right-center" | "right-bottom" | "bottom-left" | "bottom-center" | "bottom-right" | "left-top" | "left-center" | "left-bottom" = "top-center";
+  @Prop({ reflect: true}) position: "top-start" | "top" | "top-end" | "right-start" | "right" | "right-end" | "bottom-start" | "bottom" | "bottom-end" | "left-start" | "left" | "left-end" = "top-start";
```
dgibson666: You did not update the default per my previous comment - should be top (center) as it was.

dgibson666 (commented):
see code comments

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
bagrub: updated from id to class 'cbp-tooltip-arrow'

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: updated default to 'top' as requested

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
dgibson666, line 137: this can be removed - the variables are undefined and it's no longer needed.

dgibson666, line 140: this should be transparent. the browser is flagging it as invalid.

dgibson666 (commented):
couple more CSS comments

dgibson666 (approved):
Approving as-is. Can continue with some CSS cleanup and sizing specific tweaks afterward.

PR #480, opened 2026-07-27
Css Clean up

bagrub:
Clean up tooltip css:
*remove properties from css that are overridden by floatPosition util
*remove max-width property for replacement of max-inline-size
*add cssvars for height, width, and max-inline-size so end users can customize from defaults

### DIFF
--- a/packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
+++ b/packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
@@ -22,7 +22,7 @@
  :root{
   --cbp-tooltip-height: auto;
   --cbp-tooltip-width: max-content;
-  --cbp-tooltip-inline-size: 20rem;
+  --cbp-tooltip-max-width: max(20rem, 33vw);
 
   --cbp-tooltip-color: var(--cbp-color-text-lightest);
   --cbp-tooltip-color-dark: var(--cbp-color-text-darkest);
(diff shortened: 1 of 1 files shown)


packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
```diff
@@ -126,18 +129,14 @@ cbp-tooltip {
...
   position: absolute;
-  top: var(--cbp-tooltip-placement-top);
-  left: var(--cbp-tooltip-placement-left);
-  bottom: var(--cbp-tooltip-placement-bottom);
-  height: var(--tooltip-height, auto);
+  height: var(--cbp-tooltip-height);
   padding: var(--cbp-space-1x);
   min-height: var(--cbp-space-11x);
-  width:  max-content;
-  max-width: 33vw;
```
dgibson666, line 136: There should still be a max-width so that it is never larger than the viewport. Maybe 90%?

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.scss
```diff
@@ -20,7 +20,10 @@
  * @prop   --cbp-tooltip-control-color-bg-focus-dark: var(--cbp-color-interactive-focus-light);
  */
  :root{
- 
+  --cbp-tooltip-height: auto;
+  --cbp-tooltip-width: max-content;
```
dgibson666, line 24: This is not a good default because it will stretch the content out on 1 line to the maximum length possible. This needs something as a good default that works at all sizes or a clamp function, which may take some trial and error to get right.

dgibson666 (commented):
Couple code comments. I think we should also make the height and width component props for easier overrides. Outside of that, there are other updates to the component, but they can be iterated on in another PR.

dgibson666 (approved):
looks good enough. can iterate on sizing with additional changes.

PR #481, opened 2026-07-28
Feature/menu remove close button

bagrub:
Update menu to do the below: 
*remove the offset so the menu does not overlay the control
*remove the close button on the menu render


dgibson666 (commented):
This looks good overall, but Jeremy said there should be a 3x gap (offset) between the control and the menu. This should be made into a CSS variable so it can be customized.

dgibson666 (approved):

PR #482, opened 2026-07-28
Bugfix/menu gap

dgibson666:
* Update the offset to use the proper relative value computed in pixels.
* Increase z-index by 1.


bagrub (approved):
LGTM

PR #483, opened 2026-07-30
update tooltip to display on hover and dismiss on clickaway

bagrub:
*update tooltip to display/dismiss on hover seperate from click and focus display
*update click and focus to dismiss on click away
*add props for height and width  to map to CSSvars

### DIFF
--- a/packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
+++ b/packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
@@ -14,12 +14,14 @@ import { floatUI, floatUIProps } from '../../utils/floatingPlacement';
 })
 export class CbpTooltip {
 
-  @State() hoverActivated:boolean = false;
   @Element() private host: HTMLElement;
   private arrow: HTMLElement;
   private control: HTMLElement;
   private floatingEl: HTMLElement;
-  
+
+  @State() hoverActivated:boolean = false;
+  private timeoutId = undefined;
+
   /** When set, specifies that the tooltip is open */
   @Prop({ reflect: true }) open: boolean = false;
 
@@ -56,14 +58,20 @@ export class CbpTooltip {
   }
 
   hoverTooltip(x){
+
+    if(typeof this.timeoutId === "number"){
+        clearTimeout(this.timeoutId)
+        this.timeoutId = undefined;
+      }
+
     if(x && !this.open){
       this.hoverActivated = true;
       this.open= true;
-    }else if(this.hoverActivated && !x){
-      // setTimeout(() => { //TODO: not persisting on child hover like anticipated
+    }else if(this.hoverActivated && !x){      
+      this.timeoutId = setTimeout(() => {
         this.hoverActivated = false;
         this.open=false;
-      // }, 1000);
+      }, 1000);
     }
   }
 
@@ -134,9 +142,8 @@ componentDidRender(){
         aria-describedby={`${this.fieldId}`}
         role="button"
         tabindex="0"  
-        onmouseenter={() => this.hoverTooltip(true)}
         onmouseover={() => this.hoverTooltip(true)}
-        onmouseleave={() => setTimeout(() => {this.hoverTooltip(false)}, 1000)}
+        onmouseleave={() => this.hoverTooltip(false)}
         onfocus={() => this.handleFocus()}
         onClick={() => this.handleClick()}
         ref={el => (this.control = el)}
--- a/packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
+++ b/packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
@@ -14,7 +14,7 @@ import { floatUI, floatUIProps } from '../../utils/floatingPlacement';
 })
 export class CbpTooltip {
 
-  @State() hoverActivated = false;
+  @State() hoverActivated:boolean = false;
   @Element() private host: HTMLElement;
   private arrow: HTMLElement;
   private control: HTMLElement;
@@ -59,10 +59,22 @@ export class CbpTooltip {
     if(x && !this.open){
       this.hoverActivated = true;
       this.open= true;
-    }else if(this.hoverActivated && !x){//TODO: add a && this.open? getting weird cases when tooltip opened by other means
-      this.hoverActivated = false;
-      this.open=false;
+    }else if(this.hoverActivated && !x){
+      // setTimeout(() => { //TODO: not persisting on child hover like anticipated
+        this.hoverActivated = false;
+        this.open=false;
+      // }, 1000);
     }
+  }
+
+  handleClick(){
+    this.hoverActivated = false;
+    this.host.focus();
+  }
+
+  handleFocus(){
+    !this.open ? this.open = true : '';
+    clickAwayListener(this.host, _ => {this.open = false})    
   }
 
   dismissTooltip(){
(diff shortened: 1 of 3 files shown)


packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -14,6 +14,7 @@ import { floatUI, floatUIProps } from '../../utils/floatingPlacement';
 })
 export class CbpTooltip {
 
+  @State() hoverActivated = false;
```
dgibson666: Needs typing `:boolean`

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -103,8 +122,16 @@ componentDidRender(){
         aria-describedby={`${this.fieldId}`}
         role="button"
         tabindex="0"  
-        onfocus={() => this.open=true}
-        onClick={() => this.host.focus()}
+        onmouseenter={() => this.hoverTooltip(true)}
+        onmouseleave={() => this.hoverTooltip(false)}
+        onfocus={() => {
```
dgibson666: I would create a function to handle these if they are more than a single line. handleFocus and handleClick.

handleClick should also explicitly open the tooltip because there's currently an issue where if it's already focused and you click the control after the tooltip was closed (already has focus), it doesn't open again.

dgibson666 (commented):
Couple code comments. Additionally:

* Should be able to hover the tooltip itself to remain open. I think this is not working because of the gap. A short timeout on mouseleave before closing it might solve this.
* Shouldn't render the close button when activated by hover.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: update typing as requested

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: Broke out click & focus functionality into separate functions. and confirmed that focus & click issue resolved via ternary in function (issue persisted with suggested fix above)

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -56,14 +58,20 @@ export class CbpTooltip {
   }
 
   hoverTooltip(x){
```
dgibson666: please rename "x" to something that makes sense like "active" or "hovered".

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -14,12 +14,14 @@ import { floatUI, floatUIProps } from '../../utils/floatingPlacement';
...
 
-  @State() hoverActivated:boolean = false;
   @Element() private host: HTMLElement;
   private arrow: HTMLElement;
   private control: HTMLElement;
   private floatingEl: HTMLElement;
-  
+
+  @State() hoverActivated:boolean = false;
+  private timeoutId = undefined;
```
dgibson666, line 23: can we call this something like "hoverExitTimeout" or something that conveys its use?

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -134,9 +142,8 @@ componentDidRender(){
         aria-describedby={`${this.fieldId}`}
         role="button"
         tabindex="0"  
-        onmouseenter={() => this.hoverTooltip(true)}
         onmouseover={() => this.hoverTooltip(true)}
-        onmouseleave={() => setTimeout(() => {this.hoverTooltip(false)}, 1000)}
+        onmouseleave={() => this.hoverTooltip(false)}
```
dgibson666: shouldn't this be onmouseout to correspond with mouseover?

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -18,10 +18,19 @@ export class CbpTooltip {
...
   private floatingEl: HTMLElement;
-  
+
+  @State() hoverActivated:boolean = false;
+  private timeoutId = undefined;
+
   /** When set, specifies that the tooltip is open */
   @Prop({ reflect: true }) open: boolean = false;
 
+  /** Specifies the dialog height in CSS units (preferably relative units such as rem). */
```
dgibson666: replace "dialog" with "tooltip".  I would also say "Optionally" at the start.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -18,10 +18,19 @@ export class CbpTooltip {
...
+  @Prop() height: string;
+ 
+  /** Specifies the dialog width in CSS units (preferably relative units such as rem). */
```
dgibson666: replace "dialog" with "tooltip". I would also say "Optionally" at the start and add "(defaults to "auto")" at the end.

dgibson666 (commented):
seems to be working well. See code comments mostly about naming and docs.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: update to be hoverExitTimeout instead of hoverActivated

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: update to be onmouseout instead of onmouseleave

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: updated as requested

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: updated x to be hovered

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -42,10 +51,40 @@ export class CbpTooltip {
...
+      this.open= true;
+    }else if(this.hoverExitTimeout && !hovered){      
+      this.timeoutId = setTimeout(() => {
+        this.hoverExitTimeout = false;
+        this.open=false;
+      }, 250);
+    }
+  }
+
+  handleClick(){
```
dgibson666: This still isn't working as expected. You need to explicitly set open = true because if it's already focused, it doesn't activate.

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -103,13 +142,20 @@ componentDidRender(){
         aria-describedby={`${this.fieldId}`}
         role="button"
         tabindex="0"  
-        onfocus={() => this.open=true}
-        onClick={() => this.host.focus()}
+        onmouseover={() => this.hoverTooltip(true)}
+        // onmouseleave={() => this.hoverTooltip(false)}
```
dgibson666: remove

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
```diff
@@ -121,7 +167,7 @@ componentDidRender(){
             context={this.invertContext()}
             variant="square"
             accessibilityText="Close Tooltip"
-            onClick={() => this.dismissTooltip()}
+            onClick={() => this.dismissTooltip()} //TODO: recursively focusing to open tooltip, so close button not working
```
dgibson666: what is this comment about? It's working correctly for me, minus the part about clicking it again.

dgibson666 (commented):
see code comments

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: removed

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: comment to self while in process, removed

packages/web-components/src/components/cbp-tooltip/cbp-tooltip.tsx
bagrub: updated to have the this.open = true and filtered the events as discussed to make sure the 'close' button is still functional

dgibson666 (approved):
ship it!

PR #484, opened 2026-07-31
Bugfix/cbp form field techdebt

dgibson666:
* Create form field group stories.
* Add some stories/POCs for form field groups as dividers.
* Fix gaps when flex wraps in stories.
* Use "Insert" verbiage in stories for form labels and descriptions.
* Fix issues with checkbox and radio stories (causing major issues with parent event handling)
* Update cbp-form-field to omit valueChange events from components that already emit their own.
* Refactor slider to only emit its valueChange event on native change (not input)
* Only fields with errors within a group should be highlighted, not all fields within that group.
* Refactor typography to leverage setCSSProps over hard coding some tokens in CSS.
* Update typography to work within a form field group/legend tag.

Much of the techDebt fixed revolved around multiple valueChange events firing. These appear to be fixed by ensuring that `cbp-form-field` does not emit events when a child component is doing it already.

Most of the event updates can be tested/validated by inspecting the `cbp-form-field` and executing the following in the console:

```JavaScript
$0.addEventListener('valueChange', (e) => {console.log(e)});
```


bagrub (approved):
LGTM

bagrub (approved):
LGTM

PR #485, opened 2026-08-04
Make records, page, and pageSize props reactive and reduce duplicate …

dgibson666:
* Make records, page, and pageSize props reactive.
* Reduce duplicate paginationChange events.
* Improve performance by reducing the number of re-renders when a change occurs.

To test reactivity, update those props (as attributes) in the DOM.

To test events, set up a listener on the host tag:

```JavaScript
$0.addEventListener('paginationChange', (e) => {console.log(e)});
```

Updating the props in the DOM should not fire events (this is how a framework would update them, so an event isn't needed). Events are only emitted for user interactions.


bagrub (approved):
LGTM

